PR #45, opened 2013-09-14
Remove incompatibility with 'mysql'

mscdex:
Without this, the callback signatures differ when querying.


mscdex:
So I just ran into this same exact problem again after I upgraded from mysql2 0.12 to 0.15. Can we get this re-merged again?

PR #109, opened 2014-06-10
SqlString - improved speed

Mattijah:
Benchmarks:

**Format regular insert query**
Before: 362,899 ops/sec ±5.60% (86 runs sampled)
After: 611,534 ops/sec ±4.20% (86 runs sampled)

**Format bulk insert**
Before: 59,775 ops/sec ±6.43% (79 runs sampled)
After: 114,782 ops/sec ±5.58% (91 runs sampled)

**SqlString.arrayToList**
Before: 748,544 ops/sec ±5.95% (77 runs sampled)
After: 3,443,988 ops/sec ±5.53% (83 runs sampled)

I think that whole `SqlString` should be highly optimized. Functions are called hundreds to millions of times. It will make a big difference if you are working with a lot of data. (Many inserts etc...)

### DIFF
--- a/lib/sql_string.js
+++ b/lib/sql_string.js
@@ -20,13 +20,13 @@ SqlString.escape = function(val, stringifyObjects, timeZone) {
   if (val instanceof Date) {
     val = SqlString.dateToString(val, timeZone || "Z");
   }
+  
+   if (Array.isArray(val)) {
+    return '(' + SqlString.arrayToList(val, timeZone) + ')';
+  }
 
   if (Buffer.isBuffer(val)) {
     return SqlString.bufferToString(val);
-  }
-
-  if (Array.isArray(val)) {
-    return SqlString.arrayToList(val, timeZone);
   }
 
   if (typeof val === 'object') {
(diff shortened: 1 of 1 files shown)


sidorares:
to do for me: rebase this and also look at https://github.com/felixge/node-mysql/commit/f39856c8b1edeb275a0daaccafe6b62f35e971a8 diff

jtwebman:
Also @Mattijah I looked over this and you are not handling the nested array's in your optimization. That might be why is is faster.

Mattijah:
@jtwebman good you mentioned that as it seems like the current solution is not work correctly.

**Instead of:**
```
INSERT INTO table VALUES ('a' 'b', 'c'), ('d', 'e', 'f');
```

**the output is:**
```
INSERT INTO table VALUES (('a' 'b', 'c'), ('d', 'e', 'f'));
```

I don't know how could I miss that, anyway... I am not sure if there is any point for me to spent time fixing this issue as this PR is here seven months not bothering many people.

sidorares:
Hi @Mattijah - sorry it was low on my priority because 1) its perf only with no test 2) it's a change of sqlstring, and because most of the code from sqlstring and pool is taken from node-mysql but not shared properly ( as npm module ) I'm hesitant to do any changes there

Overall if there is performance benefit which is visible on a top scale ( e.i it's actually a hot path that contributes visible amount of time when measured connection-query x n-disconnect ) I'm really happy to see any experiments and PRs!

Mattijah:
Hi @sidorares - Yeah, I understand. I just noticed now that the performance of `SqlString` was already improved here https://github.com/felixge/node-mysql/pull/1331. Have you seen it?

sidorares:
yes, I'm keen to pull your change ( or that commit ). It looks very similar to what you've done, right?

sidorares:
actually @Mattijah  if you can find time to rebase your changes and/or incorporate felixge/node-mysql#1331 and test with this library that would be really great

sidorares:
@Mattijah I'm really sorry this never landed. We are using now sqlstring package shared between node-mysql and node-mysql2 - see https://github.com/mysqljs/sqlstring

PR #125, opened 2014-10-02
WIP: allow named parameters in queries

sidorares:
Currently only for server-side processed parameters. Named to unnamed happens on the client if `namedParameters` flag is set to true:

```
   db.execute("select foo from bar where id > :id", { id: 123 }, callback);
   // is the same as
   db.execute("select foo from bar where id > ?", [123], callback)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,5 @@ language: node_js
 before_script:
   - mysql -e 'create database if not exists test;'
 node_js:
-  - "0.8"
   - "0.10"
   - "0.11"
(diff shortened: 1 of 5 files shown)


mnpenner:
Did you redact this feature? It'd be really nice to have :smile:

mnpenner:
@sidorares Apparently I do need help. I ran a search on the codebase for `namedParameters` and didn't find any references.

Oh... I see, it's been renamed `namedPlaceholders` hasn't it? I'll try this out, thanks :smile:

sidorares:
yes, it was renamed ( readme is referring to a correct name )

Actually it might be a good idea to throw error if unknown config key is passed to save time debugging typos

mnpenner:
Probably wouldn't hurt :smile: 

`namedPlaceholders` works as expected. Thanks!!

PR #142, opened 2015-01-08
implement load local infile (text protocol only) #64

sidorares:


README.md
```diff
@@ -105,6 +104,20 @@ connection.query(options, function(err, results) {
...
+
+```
+// local file
+connection.query('LOAD DATA LOCAL INFILE "/tmp/data.csv" INTO TABLE test FIELDS TERMINATED BY ? (id, title)', onInserted1);
+// local stream - does not exist on fs. Any name can be used, if matched to mapping stream parameteris used, otherwise local file.
+
+var sql = 'LOAD DATA LOCAL INFILE "mystream" INTO TABLE test FIELDS TERMINATED BY ? (id, title)';
+connection.query({
+  sql: sql,
+  infileMapping: { mysqtream: request }
```
dougwilson: typo

Issue #113, opened 2014-07-29
Unhandled exception in Packet.readLengthCodedNumber


matthew-dean-hp:
This issue was first reported for node-mysql (see http://github.com/felixge/node-mysql/issues/867 for the exact setup) but is also present in this library. Basically when connected to two nodes of a galera cluster if I repeatedly call a stored procedure that returns multiple result sets an unhandled exception sometimes occurs.

The debug output for a test case is as follows:

~~~
Add command: Connection
 ==> Connection#handshakeInit(0,,102)
 <== Connection#handshakeInit(1,,73)
 ==> Connection#handshakeResult(2,maybeOK,7)
Add command: query
Add command: Connection
 <== query#start(0,,62)
 ==> Connection#handshakeInit(0,,102)
 <== Connection#handshakeInit(1,,73)
 ==> Connection#handshakeResult(2,maybeOK,7)
Add command: query
Add command: Connection
 <== query#start(0,,62)
 ==> Connection#handshakeInit(0,,102)
 <== Connection#handshakeInit(1,,73)
 ==> Connection#handshakeResult(2,maybeOK,7)
Add command: query
Add command: Connection
 <== query#start(0,,62)
 ==> query#resultsetHeader(1,,1)
 ==> query#readField(2,,52)
 ==> query#readField(3,,56)
 ==> query#fieldsEOF(4,EOF,5)
 ==> query#row(5,,39)
 ==> query#row(6,EOF,5)
 ==> query#resultsetHeader(7,maybeOK,7)
 ==> Connection#handshakeInit(0,,102)
 <== Connection#handshakeInit(1,,73)
 ==> Connection#handshakeResult(2,maybeOK,7)
Add command: query
 <== query#start(0,,62)
Add command: query
 <== query#start(0,,62)
Add command: Connection
 ==> query#resultsetHeader(1,,1)
 ==> query#readField(2,,52)
 ==> query#readField(3,,56)
 ==> query#fieldsEOF(4,EOF,5)
 ==> query#row(5,,39)
 ==> query#row(6,EOF,5)
 ==> query#resultsetHeader(7,maybeOK,7)
 ==> Connection#handshakeInit(0,,102)
 <== Connection#handshakeInit(1,,73)
 ==> query#resultsetHeader(1,maybeOK,7)
 ==> Connection#handshakeResult(2,maybeOK,7)
Add command: query
 <== query#start(0,,62)
Add command: query
 <== query#start(0,,62)
Add command: query
[...]

PR #155, opened 2015-02-23
fix header2 and header3 states lengtgh calculation

sidorares:
fixed a bug where parser incorrectly handles network frame crossing first 2 or 3 bytes of mysql packet

see #113


sidorares:
merged as da1c4c2254ce963fc856e89056193fe0450f1e96  51405918826479fcc9532bbf2ea017eb74bbc7b9 c2d4e01ef008dd39b026757ceaa197de2620d47a

PR #172, opened 2015-05-25
Support bigNumberStrings config

woollybogger:
Adds support for the `bigNumberStrings` config option which fixes an incompatibility issue with node-mysql.

Previously, numbers <=2<sup>53</sup> were being returned as numbers for columns of type `BIGINT` even when `bigNumberStrings` was set to `true`. This PR aligns the node-mysql2 API with that of node-mysql so that `BIGINT` column values are always returned as strings when the `supportBigNumbers` and `bigNumberStrings` config options are `true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ In addition to password `createConnection()`, `createPool()` and `changeUser()` 
 
 ## Known incompatibilities with node-mysql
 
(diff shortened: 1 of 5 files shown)


sidorares:
thanks!

PR #175, opened 2015-05-25
Fixed #174

### LINKED ISSUE #174: Query.stream exception
Hi,

I am getting this exception:
.../mysql2/lib/commands/query.js:221
    if (!stream.push(row)) self._connection.pause();
                                           ^
TypeError: Cannot read property 'pause' of undefined

If I put a breakpoint there, I can see that indeed _connection is null.
The fix was to add:
```js
this._connection = connection;
```
in Query start method (Query.js line 32)

asafyish:


sidorares:
no, `null` in constructor is just for performance ( so that structure of the object is not modified outside of constructor )

PR #180, opened 2015-06-21
Add .escapeId() to Connection and Pool

woollybogger:
This fixes an incompatibility with **node-mysql** since this is in the documentation [here](https://github.com/felixge/node-mysql#escaping-query-values) and [here](https://github.com/felixge/node-mysql#escaping-query-identifiers).


sidorares:
Thanks!

PR #192, opened 2015-08-25
Fix node-mysql compatibility

mscdex:
This fixes issues when this module is used in conjunction with modules that check `arguments.length` of the callbacks, like the `async` module's `waterfall()`.

Ref #45 and #46

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -53,11 +53,11 @@ Query.prototype.done = function() {
     }
     if (fields) {
       process.nextTick(function() {
-        self.onResult(null, rows, fields, self._resultIndex + 1);
+        self.onResult(null, rows, fields);
       });
     } else {
       process.nextTick(function() {
-        self.onResult(null, rows, void(0), self._resultIndex + 1);
+        self.onResult(null, rows);
       });
     }
   }
@@ -75,8 +75,8 @@ Query.prototype.doneInsert = function(rs) {
   }
   this._rows.push(rs);
   this._fields.push(void(0));
-  this.emit('result', rs, this._resultIndex);
-  this.emit('fields', void(0), this._resultIndex);
+  this.emit('result', rs);
+  this.emit('fields', void(0));
   if (rs.serverStatus & ServerStatus.SERVER_MORE_RESULTS_EXISTS) {
     this._resultIndex++;
     return this.resultsetHeader;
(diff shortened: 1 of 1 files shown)


sidorares:
I'll need to look ad failing "multiple statements" tests. I did big refactoring of common query/execute paths into base query, this is probably where incompatibility you are fixing was introduced again

mscdex:
This is still a problem with the latest version, any chance to get it merged or otherwise remedied?

sidorares:
I'd love to have this merged, but some tests are failing ( test\integration\connection\test-multiple-results.js at least ) do you have capacity to look at them? Otherwise I'll try myself over weekend

mscdex:
Well, my guess is that it's because the tests are expecting the argument I'm removing for backwards compatibility. Other code will probably need to be refactored to pass the same information along.

Osterjour:
I assume we're running into the same issue as we're using mysql2 with async and sometimes our code just stops running (maybe due to a not called callback). So a merge (and npm release afterwards) would be highly appreciated. As @mscdex mentioned the tests seem to expect arguments which the pull requests removes. I didn't dig deeper into this but for somebody who's familiar with the code and tests I suppose this shouldn't be that hard and would make mysql2 actually usable again for many people...

sidorares:
Hi @Osterjour 
I'm keen to fix this and it's actually very easy to do so, but I keep thinking about "what's the proper way to handle multi result queries" as I really don't like the fact that shape of response might be completely different if you have one or more than one result

Osterjour:
Hi @sidorares,
in the actual pull request I don't see a different way of handling different queries (but to be honest I didn't really dig deeper into it and just checked the diff). So maybe you could please put some light on the real issue for you here so we can discuss this and (hopefully) find a suitable solution?

sidorares:
It removes information that helps to distinguish for example "array of two rows" vs "array of two resultsets" ( number of results in response, that third parameter that causing problems )

sidorares:
I'm leaning towards removing it as almost no one is using multiple statements and there is still a way to track them correctly - listen 'fields' event and increment counter in your code

sidorares:
moved to #266

sidorares:
@Osterjour @mscdex - this is published to npm as v1.0.0-rc.1

mscdex:
Thanks @sidorares I appreciate this very much :-)

sidorares:
sorry it took so long :)

actually I probably need to add more tests that will guard against unwanted parameters in callback. Recent change only fixed tests that relied on that parameters

Osterjour:
@sidorares Thanks a lot!

Issue #39, opened 2013-08-28
custom typeCast handlers in generated parser


danielsantiago:
The typeCast function in ConnectionConfig is not been called. I need this to automatically cast some VARBINARY and BINARY columns to hexadecimal. Another use case is that I have an application that query more than 2,000 calendar events and converting DATETIME fields to Date objects is too much for me, with typeCast I was omitting this.

PS: Good jobs with the SSL implementation, thanks.

sushantdhiman:
Implemented in https://github.com/sidorares/node-mysql2/pull/351

PR #199, opened 2015-09-16
Add decimal numbers option

jtwebman:
I know you say it is faster to have decimals as strings but we really need to have an override to make them numbers. I went ahead and added the option. Would you be game to merge this into master or another branch? It was a simple add and I also included some tests for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,13 @@ In addition to password `createConnection()`, `createPool()` and `changeUser()` 
 In contrast to node-mysql, `zeroFill` flag is ignored in type conversion.
 You need to check corresponding field zeroFill flag and convert to string manually if this is of importance to you.
 
-DECIMAL and NEWDECIMAL types always returned as string
+DECIMAL and NEWDECIMAL types always returned as string unless you pass this config option:
+```javascript
+{
+  decimalNumbers: true
+}
+```
+**_Warning this option could lose precision on the number as Javascript Number is a Float!_**
 
 ## Examples
 
--- a/lib/compile_binary_parser.js
+++ b/lib/compile_binary_parser.js
@@ -121,6 +121,9 @@ function readCodeFor(field, config) {
     return "packet.readTimeString()";
   case Types.DECIMAL:
   case Types.NEWDECIMAL:
+    if (config.decimalNumbers) {
+      return "packet.parseLengthCodedFloat();";
+    }
     return "packet.readLengthCodedString();";
   case Types.GEOMETRY:
     return "packet.parseGeometryValue();";
--- a/lib/compile_text_parser.js
+++ b/lib/compile_text_parser.js
@@ -83,6 +83,9 @@ function readCodeFor(type, charset, config) {
     return "null; packet.skip(1);";
   case Types.DECIMAL:
   case Types.NEWDECIMAL:
+    if (config.decimalNumbers) {
+      return "packet.parseLengthCodedFloat();";
+    }
     return "packet.readLengthCodedString(); //" + type + ' ' + charset;
   case Types.DATE:
     if (config.dateStrings)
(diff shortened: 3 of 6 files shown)


sidorares:
I don't mind adding this option ( and I was planning to bump at least minor version due to ::name support in named placeholders in next release )
What's the main reason for you? It was that way because on wire level it's a string and it's meant to be used in precise arithmetics ( accounting etc ). Might be even better to have option like  `newDecimalBignumber`

jtwebman:
I can pull in the accounting module and add the function if you like that better? I understand it is dangerous in some cases and we using the accounting module for inserts and when doing money math.

jtwebman:
Actually never mind that is really only for money types. I can work on the newDecimalBignumber function if you think that is better but the float one works and since it needs to be a float for the number to work in JavaScript it seems to make sense.

sidorares:
> that is really only for money types
and conversion to JS number loses some precision. Is that ok for you?

jtwebman:
It is because we take care to not put to big of numbers in so we don't lose precision. I'l update the readme to make sure people understand you can lose precision with that option.

jtwebman:
Hmm maybe we should add a packet.parseDecimalStringCodedFloat() function that doesn't add padding on to string as it isn't needed for decimals. Would you agree?

sidorares:
not sure I understand problem with padding - can you give an example?

jtwebman:
When we return a decimal as a string for say a 19,4 decimal and the value is only 9.25 we return '9.2500' as the string. Is there a reason we do this?

sidorares:
I think its coming from the database that way. Try to connect with `{debug: true}` and you'll see raw binary packets ( also there are flags coming from schema that may affect this - see `column_length` and `decimals` fields http://dev.mysql.com/doc/internals/en/com-query-response.html#packet-Protocol::ColumnDefinition )

jtwebman:
Ok cool, then I'll leave it.

jtwebman:
@sidorares Ok I think this is ready to go. Let me know if there is anything else you want me to add.

sidorares:
:+1:  I'll try to push this with other changes as a 0.16.0 release tonight

sidorares:
one more thing @jtwebman - could you add both `query()` and `execute()` to both versions of test? You added change to binary protocol compiler ( used to parse `execute()` result ) but test is only using text protocol ( `connection2.query(...)` )

dougwilson:
Hi @jtwebman , the reason decimal is a string by default is not because of precision loss, but because floats are in base-2, you will loose money when doing float arithmetic. Simple example:
```
# Let's add up a cart of one $10.60 item and one $1.20 item
$ node -pe 'parseFloat("10.60") + parseFloat("1.20")'
11.799999999999999
# OK...
```

jtwebman:
Hi @dougwilson, like I said above we take care to not store 11.799999999 in the database so that is why I made an option to allow decimals to be stored in numbers. The default is still strings.

jtwebman:
FYI, we are running this in production now off my fork and branch.

jtwebman:
One thing that might be nice is if this library allowed the end user to override the conversion of any type themselves. As I could see someone want to have it output a Decimal from this library (http://mikemcl.github.io/decimal.js/) as well as maybe having dates output as moments instead.

It wouldn't be hard to add a hook to allow someone to override how a type in the system is parsed. (Even though if they do a bad job it could be much slower).

sidorares:
@jtwebman yes, we started this discussion quite a while ago at #39

jtwebman:
@sidorares since this was a few months old I went ahead and rebased off master. I also squashed the commits to make it easier for you to pull in. Let me know if you need anything else.

sidorares:
@jtwebman thanks, I'm right in the middle of reviewing what's good for next release, hope to include your PR

jtwebman:
@sidorares Thanks

sidorares:
sorry @jtwebman looks like needs rebasing again after my changes

jtwebman:
@sidorares Ok will do now

jtwebman:
Ok @sidorares all tests passing with rebase.

sidorares:
thanks!

PR #225, opened 2015-11-25
Update pool.js

hugnosis:
[performance]
double-ended-queue: 1240 ms
pattern/queue: 255 ms

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -3,7 +3,7 @@ var mysql        = require('../index.js');
 var EventEmitter   = require('events').EventEmitter;
 var Util           = require('util');
 var PoolConnection = require('./pool_connection.js');
-var Queue          = require('double-ended-queue');
+var Queue          = require('pattern/queue');
 
 module.exports = Pool;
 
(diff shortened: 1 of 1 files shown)


sidorares:
could you show benchmarks used?

Also interesting to compare performance for command queue - https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L42-L46

sidorares:
Also just noticed that pool.js is using splice on queues ( originally Array was used as container ) - https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L42-L46

need to check if `Pool._removeConnection` is used at all

sidorares:
results I have using [benchmark](https://github.com/petkaantonov/deque/blob/master/benchmark/thousand.js) from double-ended-queue:

```
double-ended-queue x 19,347,920 ops/sec ±8.16% (84 runs sampled)
pattern-queue x 2,646,092 ops/sec ±2.13% (85 runs sampled)
node-deque x 10,957,111 ops/sec ±15.80% (87 runs sampled)
built-in array x 2,410,360 ops/sec ±2.23% (92 runs sampled)
```

hugnosis:
Sorry, there was a bug in my tests.
Thanks to your response.

sidorares:
no problems, thanks for interest!

PR #227, opened 2015-11-26
Minor improvments to SqlString

sapics:
This pr is just minor improvments to SqlString.

### DIFF
--- a/lib/sql_string.js
+++ b/lib/sql_string.js
@@ -82,12 +82,10 @@ SqlString.format = function(sql, values, stringifyObjects, timeZone) {
 SqlString.dateToString = function(date, timeZone) {
   var dt = new Date(date);
 
-  if (timeZone != 'local') {
-    var tz = convertTimezone(timeZone);
-
-    dt.setTime(dt.getTime() + (dt.getTimezoneOffset() * 60000));
-    if (tz !== false) {
-      dt.setTime(dt.getTime() + (tz * 60000));
+  if (timeZone !== 'local') {
+    var timezoneOffset = convertTimezone(timeZone) + dt.getTimezoneOffset();
+    if (timezoneOffset) {
+      dt.setTime(dt.getTime() + timezoneOffset * 60000);
     }
   }
 
@@ -130,11 +128,11 @@ function zeroPad(number, length) {
 }
 
 function convertTimezone(tz) {
-  if (tz == "Z") return 0;
+  if (tz === "Z") return 0;
 
   var m = tz.match(/([\+\-\s])(\d\d):?(\d\d)?/);
   if (m) {
-    return (m[1] == '-' ? -1 : 1) * (parseInt(m[2], 10) + ((m[3] ? parseInt(m[3], 10) : 0) / 60)) * 60;
+    return (m[1] === '-' ? -1 : 1) * (parseInt(m[2], 10) * 60 + (m[3] ? parseInt(m[3], 10) : 0));
   }
   return false;
 }


lib/sql_string.js
```diff
@@ -130,11 +122,11 @@ function zeroPad(number, length) {
...
-  if (tz == "Z") return 0;
+  if (tz === "Z") return 0;
 
   var m = tz.match(/([\+\-\s])(\d\d):?(\d\d)?/);
   if (m) {
-    return (m[1] == '-' ? -1 : 1) * (parseInt(m[2], 10) + ((m[3] ? parseInt(m[3], 10) : 0) / 60)) * 60;
+    return (m[1] === '-' ? -1 : 1) * (parseInt(m[2], 10) * 60 + (m[3] ? parseInt(m[3], 10) : 0));
   }
-  return false;
+  return 0;
```
gajus: This does not look right to me. Whats the reason for returning `0` in case `tz` is "Z" or `tz.match` does not match anything? I am guessing you have done it to hide `consistent-return` ESLint error. Though, that warning is not saying that you should try to make the returning types match – it is warning about a potentially dangerous pattern.

In this case, I think that this function should throw an error in all cases that it is now returning `0`. This further implies that the calling context must be aware of these limitations and must not call the function in situations where it would cause error.

lib/sql_string.js
```diff
@@ -82,24 +82,16 @@ SqlString.format = function(sql, values, stringifyObjects, timeZone) {
...
-  var year   = dt.getFullYear();
-  var month  = zeroPad(dt.getMonth() + 1, 2);
-  var day    = zeroPad(dt.getDate(), 2);
-  var hour   = zeroPad(dt.getHours(), 2);
-  var minute = zeroPad(dt.getMinutes(), 2);
-  var second = zeroPad(dt.getSeconds(), 2);
-  var millisecond = zeroPad(dt.getMilliseconds(), 3);
-
-  return year + '-' + month + '-' + day + ' ' + hour + ':' + minute + ':' + second + '.' + millisecond;
+  return dt.getFullYear() + '-' + zeroPad(dt.getMonth() + 1, 2) + '-' + zeroPad(dt.getDate(), 2) + ' ' +
```
gajus: I don't agree with this format. It is a lot more cryptic than the original format.

lib/sql_string.js
```diff
@@ -82,24 +82,16 @@ SqlString.format = function(sql, values, stringifyObjects, timeZone) {
...
-  if (timeZone != 'local') {
-    var tz = convertTimezone(timeZone);
-
-    dt.setTime(dt.getTime() + (dt.getTimezoneOffset() * 60000));
-    if (tz !== false) {
-      dt.setTime(dt.getTime() + (tz * 60000));
+  if (timeZone !== 'local') {
+    var timezoneOffset = convertTimezone(timeZone) + dt.getTimezoneOffset();
+    if (timezoneOffset) {
+      dt.setTime(dt.getTime() + (timezoneOffset * 60000));
```
gajus: Brackets around the multiplication are redundant.

lib/sql_string.js
sapics: I am agree with throwing an error is useful for decreasing mistakes of timezone format.
Sorry, I don't know about the reason for returning 0 in case tz is "Z".

On the other hand, node-mysql2 need to have a compatibility with node-mysql.
From the node-mysql source code, "return false" would be better for consistency.
https://github.com/felixge/node-mysql/blob/master/lib/protocol/SqlString.js#L177

Thus, in this PR, I would keep original "return false".
I guess that it would be better to "throwing an error", after node-mysql change.

lib/sql_string.js
sapics: Thanks! I am agree with you.

lib/sql_string.js
sapics: Thanks! I will fix it.

lib/sql_string.js
sidorares: I can't remember if there is anything specific to this module here in sql_string ( except from file name convention change ) We started discussion with @dougwilson about factoring out some common code own packages and moving some shared code to https://github.com/mysqljs org. SqlString might be good first candidate.

sapics: "Sharing common code" is sounds great! :+1:

sidorares:
@sapics @gajus I created https://github.com/mysqljs/sqlstring

Ideally we should cherry-pick sql-string related commits from node-mysql ( implementation + tests ) so that git comments are preserved

Also need to contact https://www.npmjs.com/package/sqlstring author, it serves the same goal, maybe use this npm name he is happy ? ( ping @node-modules @fengmk2 )

fengmk2:
yep, very happy to see this. I will transfer sqlstring name to it.

fengmk2:
@sidorares Please ping me when https://github.com/mysqljs/sqlstring get ready.

dougwilson:
@sidorares starting with the SqlString stuff would be a great first module :) !

@sidorares @fengmk2 what if instead of starting from scratch, if @fengmk2 wants to hand the `sqlstring` module name to the `mysqljs` organization, we simply move the https://github.com/node-modules/sqlstring repository into the `mysqljs` organization and simply move forward from that? Otherwise, we can always re-extract everything as a module and then publish it as 2.0.0 after that is complete?

sidorares:
@dougwilson yeah, I was going to suggest to move https://github.com/node-modules/sqlstring repo under mysqljs ( and preserve issues ) but would be also good to extract original git commits ( which https://github.com/node-modules/sqlstring does not have ). Though either way is a good start :)

fengmk2:
@sidorares remove https://github.com/mysqljs/sqlstring and I transfer https://github.com/node-modules/sqlstring to mysqljs's org.

sidorares:
@fengmk2 done

fengmk2:
> You don’t have admin rights to mysqljs
 
@sidorares Can I join the mysqljs org?

dougwilson:
@fengmk2 I just invited you :)

fengmk2:
done and add you both to npm owners.

```
npm owner ls sqlstring
dougwilson <doug@somethingdoug.com>
fengmk2 <fengmk2@gmail.com>
sidorares <sidorares@yandex.com>
```

gajus:
Regarding NPM package names, instead of spending time trying to get existing names, why not create NPM scope @mysqljs? http://blog.npmjs.org/post/116936804365/solving-npms-hard-problem-naming-packages

dougwilson:
@sidorares the `sqlstring` module has been published as 2.0.1 and it's a copy of what is currently in the `mysql` module (or what was until it was just removed and replaced as a reference to `sqlstring`).

sidorares:
Awesome, @dougwilson ! I'll try to use it and see if tests pass

sidorares:
master is now using https://github.com/mysqljs/sqlstring, tests look ok

PR #228, opened 2015-11-26
Add spliceConnection function

### LINKED ISSUE #226: Error in pool.js
Thank you for your great work!
It's really fast!

Currently I am using node-mysql2 with pool feature.
I had an error `TypeError: this._allConnections.splice is not a function` at https://github.com/sidorares/node-mysql2/blob/master/lib/pool.js#L161
I guess that double-ended-queue api dose not have `splice` function, from https://github.com/petkaantonov/deque.

sapics:
This pr is just one of the method to fix #226.
If this library has similarity with node-mysql, `connection._pool = null;` at https://github.com/felixge/node-mysql/blob/master/lib/Pool.js#L261 might be necessary.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -154,21 +154,12 @@ Pool.prototype.execute = function (sql, values, cb) {
 };
 
 Pool.prototype._removeConnection = function(connection) {
-  var i;
 
-  for (i = 0; i < this._allConnections.length; i++) {
-    if (this._allConnections[i] === connection) {
-      this._allConnections.splice(i, 1);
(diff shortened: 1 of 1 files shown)


sidorares:
would be great to have some tests covering removeConnection. You'll need to set up local server similar to https://github.com/sidorares/node-mysql2/blob/master/test/integration/connection/test-disconnects.js and connect using pool, then terminate some connections and check that they are removed from pool

sidorares:
I'll add more tests as part of #230 , merging this

PR #246, opened 2016-01-07
Use rowsAsArray instead rowsAsHash for statementKey and keyFromFields Evaluation

### LINKED ISSUE #245: rowsAsHash used in statement key evaluation
Sorry, if this is a stupid question, but I've noticed that you used `options.rowsAsHash` as a part of prepared statements key evaluation: 

* https://github.com/sidorares/node-mysql2/blob/7584df7543d475db1a1b96affc4c62671281d1e1/lib/connection.js#L472.
* https://github.com/sidorares/node-mysql2/blob/7584df7543d475db1a1b96affc4c62671281d1e1/lib/connection.js#L464

Shouldn't it be `rowsAsArray` instead? I can not find any mentions of `rowsAsHash` option variable except this piece of code.

dolphin278:
As I mentioned in #245, non-existent options field `rowsAsHash` was used for key evaluation for statement and fields.

This actually fixes some weird behavior when same prepared statement executed with `rowsAsArray` option variable set to `true` lead to mixed results, when sometime you've got rows as array, and other times – as `BinaryRow` objects, e.g.:

```js
var files = pool.executeAsync({
          sql: `SELECT id FROM ... `,
          rowsAsArray: true
        }, [id]);
```
produces mixed results when executed multiple times using pool of connections (console.log output on query result):
```
row [ [ 22500 ] ]
row [ BinaryRow { id: 1000307 } ]
row [ BinaryRow { id: 1000307 } ]
row [ [ 22500 ] ]
row [ [ 22500 ] ]
row [ [ 1000307 ] ]
row [ [ 1000302 ] ]
row [ [ 1000270 ] ]
row []
row [ BinaryRow { id: 1000274 } ]
row [ [ 1000274 ] ]
row [ BinaryRow { id: 1000274 } ]
row [ [ 1000270 ] ]
row [ [ 1000267 ] ]
```

Bug was initially observed on Node v 5.3.0.

It fixes #245.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -461,7 +461,7 @@ Connection.prototype.resume= function resume() {
 };
 
 Connection.prototype.keyFromFields = function keyFromFields(fields, options) {
-  var res = (typeof options.nestTables) + '/' + options.nestTables + '/' + options.rowsAsHash;
+  var res = (typeof options.nestTables) + '/' + options.nestTables + '/' + options.rowsAsArray;
   for (var i=0; i < fields.length; ++i)
     res += '/' + fields[i].name + ':' + fields[i].columnType + ':' + fields[i].flags;
   return res;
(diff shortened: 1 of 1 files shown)


sidorares:
Thanks! probably mistyped it because this is how it was named long time ago - https://github.com/sidorares/nodejs-mysql-native/blob/master/lib/mysql-native/socketclient.js#L27

dolphin278:
@sidorares Great, thank you :+1:

PR #269, opened 2016-02-19
initial version of built-in promise support

sidorares:
WIP

in light of https://github.com/nodejs/promises/issues thinking to have implementation of promise wrappers in this module ( this is ~50 lines of code anyway )

TODO:

- [ ] cc authors of co/promise mysql wrappers
- [x] discuss api with @dougwilson 
- [x] Pool / PoolCluster
- [x] allow to pass custom Promise implementation
- [ ] documentation 

current API:

```js
// es7 async/await
let mysql = require('mysql/promise');

// or: mysql = require('mysql'); conn = mysql.createConnectionPromise(...)

async function test() {
  const c = await mysql.createConnection({ port: 3307, user: 'root', namedPlaceholders: true });
  const [rows, fields] = await c.query('show databases');
  console.log(rows)
  console.log( await c.execute('select 1+:toAdd as qqq', {toAdd: 10}) );
  console.log( await c.execute('select 1+:toAdd as qqq', {toAdd: 20}) );
  await c.end();
}

```


examples/promise-co-await/.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["es2015", "stage-2"],
```
gajus: When using just a few features, it is best to use separate plugins. It keeps the compiled code a lot more like the original code.

gajus:
Looks great.

The only thing that I would like to be added is an ability to choose which Promise implementation to use, e.g. I might want to use Bluebird instead of the native Promise implementation. This could be done when creating connection, e.g.

```js
await mysql.createConnection({
    port: 3307,
    user: 'root',
    namedPlaceholders: true,
    Promise: require('bluebird')
});
```

gajus:
Sorry, just noticed that it is already in the todo list.

examples/promise-co-await/.babelrc
sidorares: can you help with this? I think I just threw here first version that worked with aync/await

gajus: Absolutely.

You need to be using:

* [`babel-plugin-transform-async-to-generator`](http://babeljs.io/docs/plugins/transform-async-to-generator/)

Unless I am missing something, thats it. Which Node.js versions are you targeting?

dougwilson:
Why add complexity to provide a promise constructor via a constructor vs simply using the global Promise? I have had to track down some weird bugs when modules using different promise implementations try to chain together, and have usually found that if someone wants to use Bluebird, the global Promise constructor can just be replaced at runtime, and all promises would stay consistent.

gajus:
> and have usually found that if someone wants to use Bluebird, the global Promise constructor can just be replaced at runtime, and all promises would stay consistent.

Because replacing setting anything globally is not reasonable, e.g. imagine if you added dependency `mysql2` and it replaced global promise implementation with Bluebird. Thats not what you are suggesting. I understand the suggestion that you are making. However, even at the project level replacing a global variable such as `Promise` can disrupt expectation of the dependencies.

Regardless, giving user a granular control of what Promise implementation to use for a specific instance of `mysql2` (or whatever else package for that matter) is a good balance of freedom and isolation.

dougwilson:
@gajus gotcha. Yea, definitely not saying a module should change `global.Promise` or anything :) I think the main issues I've seen came from when modules would do something like `var Promise = require('bluebird')` and after installing a given module tree, would end up with various different versions of Bluebird, some with bugs that have already been fixed. Usually it works out OK until you encounter one of those that have a bug interacting with another version of Bluebird or other promise implementation.

Providing an option to the contructor to provide the promise constructor is probably a god balance between these two concerns (yours and mine).

examples/promise-co-await/.babelrc
sidorares: probably at least node with native generators support, because promise api is optional and lazy-loaded.
This `.babelrc` is just to allow to run examples easily ( run npm install in examples folder + propably bablel-node for those  requiring transpilation )

sidorares:
yes, I was going to use something like `var PromiseImpl =  config.Promise || global.Promise || require('es6-promise')`

PR #276, opened 2016-03-18
Remove array memory leak

willfitch:
This change removes the unused connections property. It's a very bad memory leak.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -18,7 +18,6 @@ Server.prototype._handleConnection = function(socket) {
   var connectionConfig = new ConnectionConfig({ stream: socket, isServer: true});
   var connection = new Connection({ config: connectionConfig});
   this.emit('connection', connection);
-  this.connections.push(connection);
 };
 
 Server.prototype.listen = function(port, host, backlog, callback) {
--- a/test/integration/connection/test-disconnects.js
+++ b/test/integration/connection/test-disconnects.js
@@ -16,7 +16,7 @@ function test() {
     connection.on('error', function(_err) {
       err = _err;
     });
-    server.connections.forEach(function(conn) { conn.stream.end(); } );
+
     server._server.close(function() {
       assert.equal(err.code, 'PROTOCOL_CONNECTION_LOST');
     });
--- a/test/integration/connection/test-disconnects.js
+++ b/test/integration/connection/test-disconnects.js
@@ -6,6 +6,8 @@ var fields;
 var err;
 var server;
 
+var connections = [];
+
 function test() {
   var connection = common.createConnection({port: 3307});
   connection.query('SELECT 123', function(err, _rows, _fields) {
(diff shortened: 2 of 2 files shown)


sidorares:
maybe it's better to add a code that removes connection from the list on disconnect?

willfitch:
That property isn't used anywhere beyond the array push.  I would agree if maximum connections were being enforced, but I could see nowhere that property was being used.

willfitch:
One thing I would say - personally, I think this library should provide max connections/errors.  However, this change should be applied ASAP.  The Connection object is never cleaned up by the garbage collector as the push() reference is never decremented.  It takes about 3 minutes under 250 requests/second for Node to crash due to memory exhaustion with this leak.

After this change is applied and a release is pushed, then enforcing max connections/etc be considered.

sidorares:
It's property of a server, most users don't use it. 
mysql server exposes max connections limit because connection is quite expensive: it creates full os thread per connection ( 2-4mb of memory ) + sql context etc. With this library one connection is much cheaper, and I expect it to be used more like a timing/logging/instrumentation/federation proxy, not as a full db implementation.

sidorares:
So for max connections same consideration apply as for, say, node based http server: why would you want to limit based on a number? It could be based on free mem / cpu etc, and you can already do this in "connect" handler - just reject connection if you don't feel like it's good time to connect

willfitch:
I totally understand your perspective, and having intentions on keeping up with the connections.  However, adding the `this.connections.push(...` introduced a memory leak that crashes this library within a few minutes of heavy use.  I am currently using a version of 1.0.0-rc1 with this patch in order to achieve stability.  Right now, this small bug will crash any and all use of this library from a server perspective.

sidorares:
@willfitch I'm trying to understand if this can relate to users who don't create mysql server. Do you use this library as a server in your application?

willfitch:
How would this relate to users who aren't using the server functionality? The connections property is only in the Server class.  And yes, we're attempting to use it, but this along with a couple of protocol bugs are showstoppers.  We'll be submitting more PRs to fix the protocols shortly.

sidorares:
@willfitch how about my proposal about keeping track of all connections but remove from hash when disconnected?

willfitch:
Whatever you want.  I'd just like this library to be usable at this point.

sidorares:
after some thoughts - let's do your way. If users need to track other connections there is a way to add this association in a code that is using server

sidorares:
your change breaks this test - https://github.com/sidorares/node-mysql2/blob/e7b23d7c98d8bb71f377fe5782306123c2778405/test/integration/connection/test-disconnects.js#L19 - could you update it?

willfitch:
@sidorares updated

sidorares:
@willfitch could you keep test functionality? ( save connection to local array in `serverHandler(conn) {...`  and use this array  later instead of `server.connections`

willfitch:
Updated

sidorares:
thanks, happy to merge once CI finish

willfitch:
thanks @sidorares

Issue #270, opened 2016-02-24
1.0.0-rc.1 connection pool get exhausted using pool's query method


dolphin278:
I know this is bad bug report, since I am currently not able to provide reproducible test, but anyway:

* On 1.0.0-rc.1, using `Pool.query` method, you eat all connections and they are not released to pool ([callback is never invoked](https://github.com/sidorares/node-mysql2/blob/master/lib/pool.js#L134) and connection is not released)
* On 0.15.8, `Pool.query` works just fine, pool is not getting exhausted
* On 1.0.0-rc.1, `Pool.execute` works properly, connections are successfully returning to pool

Node v5.6.0. I'll try to prepare working example for debugging, but I am in a bit of a rush here, so I don't know exactly.

sidorares:
thanks for report @dolphin278 !

There was a number of changes in `Pool.query` implementation to allow to access query command, might be a bug there. example would be very helpful!

sidorares:
hi @dolphin278  - any update on this?

dolphin278:
@sidorares I've got narrowed it down – when `Pool.query` used in two argument form – `sql` and `cb` only, callback freeing pool connection (that is [passed to `conn.query`](https://github.com/sidorares/node-mysql2/blob/master/lib/pool.js#L134)) is never called. Moreover, even if it will be called, it will work incorrect – [`cb` argument here](https://github.com/sidorares/node-mysql2/blob/master/lib/pool.js#L136) is always a third argument regardless, whether there was two- or three-argument invocation (i.e. actual `cb` went to `values`).

I've put some clumsy ad-hoc solution for my own use, which is not 100% compatible with your original code (it does not return `cmdQuery` object):

```js
function (sql, values, cb) {
    if (typeof values === 'function') {
      cb = values;
      values = [];
    }
    this.getConnection(function (err, conn) {
      if (err) return cb(err);
      var cmdQuery = Connection.createQuery(sql, values, function () {
        conn.release();
        if (typeof cb === 'function') return cb.apply(this, arguments);
      });
      conn._resolveNamedPlaceholders(cmdQuery);
      var rawSql = conn.format(cmdQuery.sql, cmdQuery.values || []);
      cmdQuery.sql = rawSql;
      conn.query(cmdQuery);
    });
  };
```

sidorares:
oh. Yes, createQuery parameters are incorrect

[...]

Issue #218, opened 2015-11-12
connection/pool.query callback never executed on linux


Cherviakov:
Hello again. Another app deployed on server with debian 8, using mysql2@0.15.8
Got app infinitely long wait for callback from one query (on windows this app working alright), and memory leaking. In 8 minutes it eat around 1Gb memory, and then exit because cannot allocate more.

Adding snippet from previous last bug (Connection.prototype._notifyError) send no error signal either.

Here is code of method:
```
function getBets(cb){
    console.log("Checkpoint 3");
    var selectQueryCurrentWeek,selectQueryLastWeek;
    var currentWeek = moment().week(moment().week());
    var currentWeekEnd = moment(currentWeek).add(7,'day');
    var lastWeek = moment(currentWeek).subtract(1,'week');
    selectQueryCurrentWeek = "" +
        "SELECT * " +
        "FROM transactions " +
        "WHERE actionName='bet' " +
        "AND coin!='FUN' " +
        "AND accountId IN " +
        "(SELECT accountId " +
        "FROM users " +
        "WHERE transactionDate BETWEEN " + pool.escape(utils.refineDate(currentWeek)) + " AND " + pool.escape(utils.refineDate(currentWeekEnd)) + ")";
    //console.log(selectQueryCurrentWeek);
    selectQueryLastWeek = "" +
        "SELECT * " +
        "FROM transactions " +
        "WHERE actionName='bet' " +
        "AND coin!='FUN' " +
        "AND accountId IN " +
        "(SELECT accountId " +
        "FROM users " +
        "WHERE transactionDate BETWEEN " + pool.escape(utils.refineDate(lastWeek)) + " AND " + pool.escape(utils.refineDate(currentWeek)) + ")";
    //console.log(selectQueryLastWeek);
    console.log("Checkpoint 3.1");
    async.series([
        function currentWeek(callback){
            console.log("Checkpoint 3.2");
            pool.query(selectQueryCurrentWeek,function(err,rows,field){
                if(err) console.error("currentWeekError:",err);
                else console.log("Checkpoint 3.2.1");
                callback(err,rows,field);
            });
        },
        function lastWeek(callback){
            console.log("Checkpoint 3.3");
            pool.query(selectQueryLastWeek,function(err,rows,field){
                if(err) console.error("lastWeekError:",err);
                else console.log("Checkpoint 3.3.1");
                callback(err,rows,field);
            });
        }
    ],function done(err,result){
        console.log("Checkpoint 4");
[...]

Issue #182, opened 2015-07-10
Query callback is never called (using pool and stream)


equiet:
Hi,

I am using a pool with a stream option and occasionally (~5% of the time) `connectionPool.query` does not call the callback.

Here's the code:
```js
var nr = require('newrelic');

var express = require('express'),
    app = express(),
    path = require('path'),
    mysql = require('mysql2'),
    url = require('url'),
    SocksConnection = require('socksjs'),
    port = process.env.PORT || 8083,
    proxy = url.parse(process.env.QUOTAGUARDSTATIC_URL);

var connectionPool = mysql.createPool({
    connectionLimit: 10,
    host: process.env.MYSQL_HOST,
    user: process.env.MYSQL_USER,
    password: process.env.MYSQL_PASSWORD,
    database: process.env.MYSQL_DATABASE,
    ssl: 'Amazon RDS',
    stream: new SocksConnection({
        host: process.env.MYSQL_HOST,
        port: 3306
    }, {
        host: proxy.hostname,
        port: 1080,
        user: proxy.auth.split(':')[0],
        pass: proxy.auth.split(':')[1]
    })
});

connectionPool.on('enqueue', function() {
    console.log('Waiting for available MySQL connection slot');
});

app.get('/:id.js', function(req, res) {
    var timeoutDetector = setTimeout(function() {
        console.log('DEBUG: This transaction is taking over 40s', req.params.id);
    }, 40000);

    connectionPool.query('SELECT blabla FROM bla WHERE id = ? AND blabla IS NOT NULL', [req.params.id], nr.createTracer('db:query', function(err, results) {
        clearTimeout(timeoutDetector);

        if (err) {
            console.log(err);
            return;
        }

        res.send('');
    }));
});

app.listen(port);
```

According to New Relic tracer the query never takes more than 100ms. But the timeout detector regularly notifies me the request is taking over 40s (I guess it's safe to assume that callback will never be called at this point).

Sometimes I get this error and 40 seconds later I always get notified the transaction is taking too long (sometimes I don't get this error but still get a notification): 
```
(node) warning: possible EventEmitter memory leak detected. 11 end listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
at addListener (events.js:179:15)
at Readable.on (_stream_readable.js:671:33)
at PoolConnection.Connection (/app/node_modules/mysql2/lib/connection.js:99:15)
[...]

PR #281, opened 2016-04-04
allow namedParameters for queries on pool as well

bitcloud:
Currently it seems, that it is not possible to use namedPlaceholders directly on the pool.
Made a small pull request to pass through the namedPlaceholders config.


bitcloud:
Sorry for the commit mess. I can squash it to one commit if you want and make a new PR.

bitcloud:
tests are working now, but the query on the pool connection takes ages. There seems to be another issue involved.

gajus:
@bitcloud Can you please squash commits and fix the failing test case?

bitcloud:
hey @gajus, sorry for the late response. I squashed everything in one commit now and tracked down the issue in the testcase.

There was an issue that callbacks from pooled connection queries aren't correctly passed to the query object and thus get never called.

As far as I can see, my testcase seems to be the only one that is testing parts of the pooled connection??

I fixed that one now, but you should check on that. I fixed it so the function signature stays the same. But probably it would be better to not return a query object at that place as in execute. For now it is working, but I would consider it a hack.

Was broken in #230, but could be interesting for #218 and #182 as well perhaps.

bitcloud:
Just checked some other issues and found #270. Fixed the issues in the code you posted and have even one testcase that uses the code ;-)

The original intention was to test named parameters patch from me, but it tests that issue as well.

I don't know why they attached a noop function to the error event, i think the intention was to emit an error, because attaching a noop function doesn't make any sense -- at least none i know of.

sidorares:
thanks! especially for #270, I'll prepare rc2 release

bitcloud:
I just thought about the code again and think someone could just use the events instead of callbacks. Then we should check the callback before we try to invoke it, because it could be undefined.
Committed the change already.

PR #283, opened 2016-04-12
Update pool.js

parapanda:
Right now, pooled connections never execute query callbacks.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -123,7 +123,7 @@ Pool.prototype.end = function (cb) {
 };
 
 Pool.prototype.query = function (sql, values, cb) {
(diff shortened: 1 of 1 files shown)


sidorares:
I'm not sure this is correct fix, could you look at my comment here and compare? https://github.com/sidorares/node-mysql2/issues/270#issuecomment-190426878

if you look at https://github.com/parapanda/node-mysql2/blob/master/lib/connection.js#L334-L345
when you pass `Query` object as first parameter to `Connection.query()` second callback is not used (instead the one passed to Query constructor is used)

parapanda:
Yeah. I didn't actually mean to make a PR :(

I read your comment, and tested the code. Looks solid!

PR #284, opened 2016-04-12
Update pool.js

parapanda:
@sidorares This based upon your comment...however, I still have to add "cb" to the createQuery(), or nothing is ever returned.

If you see a better way, let me know, but I did quite a bit of testing.


bitcloud:
This seems to fix the same issue that just got fixed by PR #281. This should work now in master.

sidorares:
hi @parapanda sorry this got unnoticed :( and thanks for sending original PR

PR #292, opened 2016-04-28
PoolConnection: Add correct this-binding to callbacks

m-burst:
I don't know if this actually fixes anything -- but it certainly feels more correct and possibly fixes some connection-releasing bugs


sidorares:
Thanks!

would you mind adding a test that would trigger the code you touched? Create test mysql server, create pool, kill server

see similar example (but without pool) here - https://github.com/sidorares/node-mysql2/blob/cd16a0a7ad2d273fa5126135479d4698bd554cea/test/integration/connection/test-quit.js

aikar:
@m-burst @sidorares This PR can be closed as the issue was resolved with a wrapping function already in master.

Issue #291, opened 2016-04-28
Possible incorrect length calculation


vogievetsky:
I came in here to make a fix only to see that @jitendra beat me to it in the last commit: https://github.com/sidorares/node-mysql2/commit/cd16a0a7ad2d273fa5126135479d4698bd554cea

When I was looking for the bug I found another place where the same issue might exist:
https://github.com/sidorares/node-mysql2/blob/master/lib/packets/column_definition.js#L86

lengthCodedNumberLength is being called with a string.

Should I submit a PR?

sidorares:
Yes, that would work incorrectly with multibyte characters. 
( actually, that was my commit, I forgot to configure git while using work laptop )

Note that this only occurs when you _send_ column packets, that is likely in "server" mode.

Yes, please do PR! ( if you have time, a unit test would be great: prepare binary file serialized from some columns with unicode chars and check that `toPacket()` result matches that binary file and matches results of de-serializeing with `new ColumnDefinition(data)` )

sidorares:
ok, I see you are using server @vogievetsky , cool!
Let me know if you missing any functionality there ( server side prepare/execute is in plans but not started yet )

vogievetsky:
@sidorares,
Firstly let me say: I love this project - great work!
I am only using the server portion because I am working on a MySQL gateway to Druid.
I hope to stress the server code to the limits and PR any fixes to any issues I may find along the way.

PR #295, opened 2016-04-30
added Packet.lengthCodedStringLength

vogievetsky:
Fixes: https://github.com/sidorares/node-mysql2/issues/291

Also packet length was calculated incorrectly because `Packet.lengthCodedNumberLength(str)` was always returning `9`

Added some unit tests and tested this in my own project and it seems to work well.


sidorares:
code lgtm, thanks!

I'm not sure why some travis jobs, this is definitely not related to your changes. Until I find why I'll rust re-run failing ones

PR #297, opened 2016-05-02
emit packet event

vogievetsky:
Also:
- Made a server pattern of: emit event if there are listeners, or fallback to default
- Added some docs
- Updated CLIENT constants and made them hex for readability

### DIFF
--- a/lib/commands/server_handshake.js
+++ b/lib/commands/server_handshake.js
@@ -2,6 +2,7 @@ var util     = require('util');
 
 var ClientConstants = require('../constants/client.js');
 var CommandCode     = require('../constants/commands.js');
+var Errors          = require('../constants/errors.js');
 
 var Command  = require('./command.js');
 var Packets  = require('../packets/index.js');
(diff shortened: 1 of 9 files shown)


sidorares:
awesome!

what happened with PING command?

Suggestion: add `.cllone()` method to packet and pass cloned copy ( shared buffer, but cloned state, offset ) to "packet" handlers, so that if there are multiple they won't break each other

sidorares:
oops, sorry see PING at the bottom, never mind

lib/commands/server_handshake.js
```diff
@@ -69,38 +70,69 @@ ServerHandshake.prototype.readClientReply = function(packet, connection) {
...
+  var knownCommand = true;
   var commandCode = packet.readInt8();
   switch (commandCode) {
   case CommandCode.QUIT:
-    connection.emit('quit');
-    // connection.stream.end();
+    if (connection.listeners('quit').length) {
+      connection.emit('quit');
+    } else {
+      connection.stream.end();
```
vogievetsky: Is this OK to do? You had it commented out.

vogievetsky:
Note: I just used this code in my own project and something is failing. Will investigate and fix.

vogievetsky:
Regarding `PING` I ordered the `case` statements the same as the order in the [docs](https://dev.mysql.com/doc/internals/en/text-protocol.html), so it is easier to track what goes where.

lib/commands/server_handshake.js
sidorares: should be ok. This just ends underlying stream ( tcp/unix socket etc). I think this is what mysql server does, no extra data sent in response

vogievetsky:
OK, so it seems that my previous PR (that you merged) caused some problem. I can no longer run queries via the `mysql` CLI client it errors. Before my PR it worked ok. I have to go to sleep now, will investigate and fix tomorrow.

sidorares:
no worries, let's have a look tomorrow.
I'll keep this one open as well ( pending discussion re passing clone of a packet )

vogievetsky:
was missing 2 byte filler https://dev.mysql.com/doc/internals/en/protocoltext-resultset.html (filler_2) will update rest of PR tomorrow and add a few more tests.

vogievetsky:
@sidorares I think this is good to go (or at least good for more feedback) I also fixed that problem that I created with my previous PR :-p. I am not sure why some of travis tests are failing. Is there a way to re-run?

sidorares:
yes, I can re-run them ( could you check if you are able to do this? not sure if travis permissions are same as repo collaborators )

vogievetsky:
Yes I can re-run, just noticed the icon. Also just noticed that you made me a committer :+1: Let's take this server to the moon!

sidorares:
not sure what causes it, would be good to find original reason ( maybe bumping note versions might help? seems to fail only for 4.x and 5.x occasionally )

sidorares:
```
 ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2 "No such file or directory")
The command "mysql -e 'create database /*M!50701 IF NOT EXISTS */ test;'" failed and exited with 1 during .
```

other solution: make some of the tests ( especially unit tests ) use local mysql2-based server and only have small number of end-to-end tests connect to real db

vogievetsky:
Yes, my preference is always to push tests down as far as they can go but I am no stranger to running tests against a real DB (in docker).

PR #303, opened 2016-05-06
added server example to readme

vogievetsky:
Hi @sidorares,
Wanted to share what I am using mysql2 for. Also I just published a [blog post](http://imply.io/post/2016/05/04/programmatic-plyql.html) about it.

Basically for a while now I have been developing [Plywood](http://plywood.imply.io/) which is an OLAP focused query planner. It provides a query layer to make it convenient to write [interactive applications](https://github.com/implydata/pivot/). While Plywood can and will eventually support many backends, currently it supports MySQL, Druid, and native JS.

[PlyQL](http://plywood.imply.io/plyql) is a MySQL compatible(ish) language that gets translated into Plywood. Recently I integrated mysql2 into [plyql](https://github.com/implydata/plyql), a command line tool, that effectively provides a SQL interface to Druid - Druid does not speak SQL natively.

plyql goes far to support all sorts of metadata queries that typical ODBC and JDBC drivers make, you know stuff like [this](https://github.com/implydata/plyql/blob/master/test/mysql-gateway.mocha.js#L84), [this](https://github.com/implydata/plyql/blob/master/test/mysql-gateway.mocha.js#L96), and [this](https://github.com/implydata/plywood/blob/master/test/parser/plyqlParser.mocha.js#L1363). I plan to continue developing plyql (and mysql2) to a point where most read-only operations in most MySQL clients will be supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Every command packet received by the server will be emitted as a **packet** even
 
 In addition special events are emitted for [commands](https://dev.mysql.com/doc/internals/en/text-protocol.html) received from the client. If no listener is present a fallback behavior will be invoked.
 
-  *  **quit** - Default: close the connection
+  *  **quit**() - Default: close the connection
   *  **init_db**(schemaName: string) - Default: return OK
   *  **query**(sql: string) - Please attach a listener to this. Default: return HA_ERR_INTERNAL_ERROR
   *  **field_list**(table: string, fields: string) - Default: return ER_WARN_DEPRECATED_SYNTAX
@@ -322,8 +322,9 @@ In addition special events are emitted for [commands](https://dev.mysql.com/doc/
 
   - [Mysql-pg-proxy](https://github.com/sidorares/mysql-pg-proxy)  - mysql to postgres proxy server.
   - [Mysqlite.js](https://github.com/sidorares/mysqlite.js) - mysql server with JS-only (emscripten compiled) sqlite backend.
-  - [sql-engine](https://github.com/eugeneware/sql-engine) - mysql server with leveldb backend.
-  - [mysql-osquery-proxy](https://github.com/sidorares/mysql-osquery-proxy) - connect to [facebook osquery](https://osquery.io/) using mysql client
+  - [sql-engine](https://github.com/eugeneware/sql-engine) - mysql server with LevelDB backend.
+  - [mysql-osquery-proxy](https://github.com/sidorares/mysql-osquery-proxy) - connect to [facebook osquery](https://osquery.io/) using MySQL client
+  - [PlyQL](https://github.com/implydata/plyql) - connect to [Druid](http://druid.io/) using MySQL client 
 
 ## See also:
 


sidorares:
👍 
one small nitpick - in future commits reference subsystem ( such as build/tests/protocol/server/docs ) and add some context, for example "docs: add PlyQL to a list of server users"

sidorares:
something I was thinking about lately and maybe you could have a good input @vogievetsky :

At the moment server api is very low level, even lower then client, you operate on packet level and not "commands". While this gives a great flexibility, for most user this is too low. Would be good to discuss some better api ( maybe as separate library, maybe even include sql parser ) Use cases this library might help to cover:
   - direct 1-1 proxy server, log some stats about commands
   - federation server ( need to decide something based on a query )
   - automatic read/write split between seperate outgoing servers ( need to detect if query is read or r/w )
   - auth gateway ( use ldap / ntlm / ssh to authenticate incoming mysql connections

vogievetsky:
I am not sure I follow 100% regarding the `you operate on packet level and not "commands"` : the events emitted by the server like `query` and `field_list` correspond to commands no? I think once the other commands are added in like fashion it will be a very consistent API.

Regarding your use cases:

> direct 1-1 proxy server, log some stats about commands
Isn't this is totally possible right now? This is actually how I started using/testing mysql2

> federation server ( need to decide something based on a query )
This needs to be paired with an SQL parser. Is there a good generic MySQL parser in JavaScript?
I looked at https://github.com/alibaba/nquery but could not use it for PlyQL because of GPLv2 license (PlyQL uses the more permissive Apache 2.0). I decided to roll my own. I licensing would allow I think we could make a cool example for mysql2 using nquery.

> automatic read/write split between seperate outgoing servers ( need to detect if query is read or r/w )
Same as above

> auth gateway ( use ldap / ntlm / ssh to authenticate incoming mysql connections
Again, I think this is pretty easy to do with the current server API (maybe once the handshake is more flushed out - I want to do that btw).

I think that what needs to be done now is that the server needs to become complete: compression, proper, handshake, capability flag management, other commands that are not 'known' yet, e.t.c

Could elaborate on what the server API is missing (other than a SQL parser)?

sidorares:
yes, all use cases I mentioned are possible to handle now (for last one AuthSwitch request would be helpful - see https://github.com/felixge/node-mysql/issues/1396 ) , I'm talking about good high level api to make top 80% very easy to implement.

For example, it's relatively easy to build http proxy using built in node http module, but using https://github.com/nodejitsu/node-http-proxy most common use case ( pass request as is ) is very simple api call `proxy.web(req, res);`

> the events emitted by the server like query and field_list correspond to commands no?

The "commands" in the client are state machines, and they change state based on the packet received: value returned from `command.execute(packet)` becomes next `execute()` function. [query](https://github.com/sidorares/node-mysql2/blob/e83efa4c561eb28f7de6aa9349764a265c1f1148/lib/commands/query.js) command, for example, is responsible for all transitions described by this diagram (well, not all, "more data" is currently not supported, but there are parts that supported and not on the diagram - multiple statements, select infile etc): 

graphviz-3ab2ba81081a7f3cc556d11fd09f50341bba6f15

- http://dev.mysql.com/doc/internals/en/com-query-response.html

what I'm talking is that we currently don't have server-side version of "query command" and currently in the server handler one must implement all this logic manually

Issue #237, opened 2015-12-14
API incompatibility regarding DECIMAL values


selaux:
DECIMAL values are returned differently in node-mysql and node-mysql2. In node-mysql they are returned as type `Number` while in node-mysql2 they are returned as type `String`, which seems wrong to me.

Minimal test case:

``` javascript
var mysql = require('mysql'),
    mysql2 = require('mysql2'),
    connection;

connection = mysql.createConnection(JSON.parse(process.env.MYSQL_CONFIG));
connection2 = mysql2.createConnection(JSON.parse(process.env.MYSQL_CONFIG));

connection.query('CREATE TABLE decimal_table (d DECIMAL(4,2))', function() {
  connection.query('INSERT INTO decimal_table VALUES (0.1)', function() {
    var query = connection.query('SELECT * FROM decimal_table LIMIT 1');
    var query2 = connection2.query('SELECT * FROM decimal_table LIMIT 1');

    query.on('result', function(r) {
      console.log('node-mysql', r.d, typeof r.d);
      connection.end();
    });
    query2.on('result', function(r) {
      console.log('node-mysql2', r.d, typeof r.d);
      connection2.end();
    });
  });
});
```

On my machine (OS X, MySQL 5.7) this prints:

```
node-mysql2 0.10 string
node-mysql 0.1 number
```

sidorares:
hi, this is intended ( but I'm open for discussion if that should be that way )

DECIMAL values are stored and transmitted as strings, and often used where exact precision is required ( accounting etc ), so I assume in that case some precision arithmetic library would be used ( Bignumber, bn.js etc ) and data would be passed as string to that library

Is this incompatibility causing any particular problems in your code?

mlarcher:
Is there  way to apply the same library transformation to all decimal fields before they reach the consuming code ?

sushantdhiman:
You can use `typeCast` option as used in test [here](https://github.com/sidorares/node-mysql2/blob/2cf5ee3cc7447fccb6913205e33ba361e78178d9/test/integration/config/test-typecast-global-option.js)

mlarcher:
thanks, looks like precisely what we need :)

jaredjenkins:
Related issue for Sequelize: https://github.com/sequelize/sequelize/issues/7465

sidorares:
[...]

PR #317, opened 2016-05-26
compability with integer, biginteger, double, decimal

sonsonz:
returned from the query always `string` value, the problem is when we use the value in `Numeric`,
my problem seems like this issue https://github.com/sidorares/node-mysql2/issues/237


sidorares:
@sonsonz could you elaborate a bit more on what & why is in this PR?

sidorares:
I'd prefer this functionality to be solved via `typeCast`, see examples in the issue you linked

PR #324, opened 2016-06-13
Fix #323.

### LINKED ISSUE #323: [BUG]query with ? placeholder got worng result
I tested the exact same query, mysql works well. mysql2 doesn't.
pool.query('UPDATE table SET col1=?,col2=? WHERE Id = ?', [1,'http://domain.com/?query=', 3])

It will produce `1, http://domain.com/1query=`
It replace the ? in the value instead of parameter.

But test both mysql.format and mysql2.format they out put the same. That very strange.

dsonet:
shouldn't format SQL twice since already did in connection.query.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -137,10 +137,6 @@ Pool.prototype.query = function (sql, values, cb) {
       return;
     }
 
-    conn._resolveNamedPlaceholders(cmdQuery);
-    var rawSql = conn.format(cmdQuery.sql, cmdQuery.values || []);
-    cmdQuery.sql = rawSql;
-
     conn.query(cmdQuery).once('end', function () {
       conn.release();
     });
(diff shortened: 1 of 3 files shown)


sidorares:
you close pool in your test before other tests finish. Maybe have test for this particular issue in own file? "test-parameters-questionmark.js" and have non-pool and scenarios there

sidorares:
you close pool in your test before other tests finish. Maybe have test for this particular issue in own file? "test-parameters-questionmark.js" and have non-pool and scenarios there

dsonet:
Ya I see it. I'll correct it right now.

dsonet:
It was wrong at beginning. I didn't realized that mysql TEMPORARY table can't shared across different connection.

sidorares:
thanks @dsonet !

sidorares:
published as v1.0.0-rc.4

dsonet:
That's great.

Issue #325, opened 2016-06-15
Dead connections not removed from pool


ctlajoie:
I'm experiencing a problem where connections that the server has dropped (due to mysql server restart or hitting the wait_timeout) do not get removed from the pool. I wrote a short script that reproduces the problem.

The script opens 5 pooled connections and then waits for something to happen. I set `wait_timeout = 30` for testing. When the timeout is hit, the errors are displayed and then it shows you how many connections are in the pool. It always stays at 5, even after the connections are dropped. I also get the same result if I just restart mysqld after it opens the connections.

I am using mysql2 1.0.0-rc.4

```javascript
var mysql = require('mysql2');
var pool = mysql.createPool('mysql://localhost/test?connectionLimit=5');
//connectionLimit param in the uri doesn't work.. have to set it manually
pool.config.connectionLimit = 5;

pool.reportStatus = function() {
	console.log('Pool connections: %d', this._allConnections.length);
};

pool.on('connection', function(conn) {
	console.log('db connection opened');
	conn.on('error', function(err) {
		console.error(err);
		setTimeout(() => pool.reportStatus(), 3000);
	});
});

// open connections
for (var i=0; i<10; i++)
	pool.query('select 1', ()=>{});

// show connections are opened
setTimeout(() => pool.reportStatus(), 500);


// don't let the script die
var server = require('http').createServer();
server.listen(5151, ()=>{});
```

ctlajoie:
BTW this is the output I get from running the script.

```
db connection opened
db connection opened
db connection opened
db connection opened
db connection opened
Pool connections: 5
{ Error: Connection lost: The server closed the connection.
    at Socket.<anonymous> (/home/chris/dev/ett-mail/node_modules/mysql2/lib/connection.js:110:15)
    at emitNone (events.js:91:20)
    at Socket.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:926:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9) fatal: true, code: 'PROTOCOL_CONNECTION_LOST' }
{ Error: Connection lost: The server closed the connection.
    at Socket.<anonymous> (/home/chris/dev/ett-mail/node_modules/mysql2/lib/connection.js:110:15)
    at emitNone (events.js:91:20)
    at Socket.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:926:12)
[...]

PR #326, opened 2016-06-16
Fix a bug that caused dead connections to remain in a connection pool

ctlajoie:
According to [the docs](https://github.com/petkaantonov/deque#getint-index---dynamic) for Dequeue, accessing the indexes directly is incorrect (and will produce inconsistent results). I replaced the indexers with calls to the `get()` method. This fixes the bug I reported in #325.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -187,11 +187,11 @@ Pool.prototype.escapeId = function escapeId (value) {
 function spliceConnection (queue, connection) {
   var len = queue.length;
   if (len) {
-    if (queue[len - 1] === connection) {
+    if (queue.get(len - 1) === connection) {
       queue.pop();
     } else {
       for (; --len;) {
-        if (queue[0] === connection) {
+        if (queue.get(0) === connection) {
           queue.shift();
           break;
         }


sidorares:
this is great! I'll see if I can add simple test for this

sidorares:
I was going to test with real server in a following way: capture threadId for each connection, and issue "kill [threatId]" on all ids from separate ( not pool ) connection. It's not 100% reliable test ( sometime all connections released, sometimes not - without your pool patch applied ) but it should always pass if connections released correctly

One problem with this approach is that pool "connect" event is fired before connection handshake is finished ( and there is no threadId there yet ). Would you be able to look at this?

https://github.com/sidorares/node-mysql2/blob/a4dfd41e4e38dfa81e814e17fa8deaa3ecbbbe2f/lib/connection.js#L121

https://github.com/sidorares/node-mysql2/blob/a4dfd41e4e38dfa81e814e17fa8deaa3ecbbbe2f/lib/commands/client_handshake.js#L113

https://github.com/sidorares/node-mysql2/blob/a4dfd41e4e38dfa81e814e17fa8deaa3ecbbbe2f/lib/connection.js#L580-L593

https://github.com/sidorares/node-mysql2/blob/a4dfd41e4e38dfa81e814e17fa8deaa3ecbbbe2f/lib/pool.js#L53

I'll be able to look later today unless you want to do this

ctlajoie:
> I'll be able to look later today unless you want to do this
I'm going to have a beer and watch some TV... it's getting late where I live :)
If you haven't had a chance to do it by tomorrow I'll look into it though.

sidorares:
@ctlajoie I fixed events order and added tests - could you fetch commits from pool-tests into your branch? Also I think we need to replace `[]` with `get()`  here - https://github.com/sidorares/node-mysql2/blob/pool-tests/lib/pool.js#L122

The test is very basic, would be good to add "killed in the middle of query", "do some commands on pool, kill some connections, do some more commands" etc - but I think it tests the problem you fixed

sidorares:
merged via b41603924a395e8d49e63081457d4dde1a15029e

ctlajoie:
Good call on replacing that other `[]`.. I didn't see that one. Cheers!

PR #331, opened 2016-06-21
Plugin based authentication support

sidorares:
#### Packets documentation:
http://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest

#### Protocol examples: 
http://dev.mysql.com/doc/internals/en/auth-method-switch.html

related node-mysql discussion: https://github.com/mysqljs/mysql/issues/1396

/cc @twocode

## API:

authSwitchHandler connection option. Function, accepts data as parameter and callback to pass `(err, authSwitchResponseData)`.Note that it can be invoked multiple times if server respond with AuthMoreData. At first invocation data.pluginName contains name of auth plugin, data.pluginData - buffer with data. Next invocations contain only data as `data.pluginData`.

```js
  var conn = mysql.createConnection({
    user: 'test_user',
    password: 'test',
    database: 'test_database',
    port: port,
    authSwitchHandler: function(data, cb) {
      cb(null, response) // response: string or buffer
    }
  });
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ language: node_js
 before_script:
   - mysql -e 'create database /*M!50701 IF NOT EXISTS */ test;'
 node_js:
-  - '0.10'
-  - '0.12'
-  - '4.4'
-  - '5.11'
+#  - '0.10'
+#  - '0.12'
+#  - '4.4'
+#  - '5.11'
   - '6.2'
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ language: node_js
 before_script:
   - mysql -e 'create database /*M!50701 IF NOT EXISTS */ test;'
 node_js:
-#  - '0.10'
-#  - '0.12'
-#  - '4.4'
-#  - '5.11'
+  - '0.10'
+  - '0.12'
+  - '4.4'
+  - '5.12'
   - '6.2'
 
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+1.0.0-rc-6 ( 29/06/2016 )
+ - AuthSwitch support and partial support for
+   plugin-based authentication                               #331
+
 1.0.0-rc-5 ( 16/06/2016 )
  - Fix incorrect releasing of dead pool connections          #326, #325
  - Allow pool options to be specified as URL params          #327
--- a/README.md
+++ b/README.md
@@ -80,6 +80,40 @@ co(function * () {
 ```
 see examples in [/examples/promise-co-await](/examples/promise-co-await)
 
+### Authentication switch request
+
+During connection phase the server may ask client to switch to a different auth method.
+If `authSwitchHandler` connection config option is set it must be a function that receive
+switch request data and respond via callback. Note that if `mysql_native_password` method is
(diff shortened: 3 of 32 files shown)


twocode:
Hey @sidorares 

This is fantastic. I have noticed that mysql2 can now adapt mysql authentication-switch method. Bravo!

But it seems to be not correct when the authentication switch request is "mysql_native_password", could you help to verify?


Authentication switch request (plugin method: mysql_native_password, salt):
image


Authentication switch response (0-length payload)
image

sidorares:
Yeah, looks strange

Could you connect client with `debug: true` and post debug log?

The client supposed to send response to `mysql_native_password` here: https://github.com/sidorares/node-mysql2/blob/7b4f535f5977537e639785f822fd88ab7cb9aaa2/lib/commands/client_handshake.js#L133

could you also log authToken from l133? If it's empty for some reason the resulting packet would be empty

twocode:
Hey @sidorares 

I digged a little and it turned out that in line:
https://github.com/sidorares/node-mysql2/blob/7b4f535f5977537e639785f822fd88ab7cb9aaa2/lib/commands/client_handshake.js#L63

`this.password` is undefined, thus underneath it will return a `new Buffer(0)`

If I hard-code a valid password string here, it will work correct.

Thanks

sidorares:
Thanks! I'll have a look, should be easy to fix

twocode:
@sidorares Hey man, any updates? :p

Do we need to post an issue to track this?

Best

sidorares:
@twocode hi! The fix is actually very simple, just did not have time to implement

The problem is following: When you have ChangeUser request the same AuthSwitch is performed as with initial ClientHandshake command. I'm mixing in this bit of functionality here: https://github.com/sidorares/node-mysql2/blob/b534e15faa0dbeefe466ad5a53e4e7fdf378aee8/lib/commands/change_user.js#L21 but I think `ChangeUser.prototype.handshakeResult` expects password/username fields differently

If you able to debug that would be great, otherwise wait few more days (or weeks, too busy :( )

sidorares:
yes, issue with detailed documentation and debug log would be helpful as well ( just to confirm that it's ChangeUser command etc )

sidorares:
Hi @twocode , could you try head of https://github.com/sidorares/node-mysql2/tree/auth-switch2 branch?

twocode:
@sidorares Cool man, let me give a try. Will keep you updated!

twocode:
@sidorares Hey dude, this fix has been verified. Cheers!

sidorares:
hi @twocode as I was able to replicate exactly your behaviour, just for the record - what's your server version?

twocode:
@sidorares I was using 5.6.26.

By the way, are we having this fix in rc8? Will we have this in node-MySQL as well?

sidorares:
Yes, I'll try to release another rc today and hope to have final 1.0 this week

I'm not sure about node-MySQL timeline and priorities. If everyone is happy about AuthSwitch api I might try to implement them there but afaik there are more important features on todo list there

twocode:
Hey @sidorares , I understand. No worries. Thanks so much:)

sidorares:
@twocode the fix for your issue was published with rc-9

twocode:
@sidorares Gotcha!

Issue #336, opened 2016-07-01
Packet length causes issues with bn.js 


sheeley:
Hey, I'm not certain whether this is actually an issue here or with bn.js, but I figured I'd throw this at you to see what your thoughts are.

In our codebase, using knex and mysql2, I get this stacktrace when running our seeds:
```
/Users/jsheeley/git/lumen/api/node_modules/bn.js/lib/bn.js:6
    if (!val) throw new Error(msg || 'Assertion failed');
              ^

Error: Assertion failed
    at assert (/Users/jsheeley/git/lumen/api/node_modules/bn.js/lib/bn.js:6:21)
    at BN.iaddn (/Users/jsheeley/git/lumen/api/node_modules/bn.js/lib/bn.js:2121:5)
    at Packet.readLengthCodedNumberExt (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/packets/packet.js:145:37)
    at Packet.readLengthCodedNumber (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/packets/packet.js:114:15)
    at new ResultSetHeader (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/packets/resultset_header.js:12:28)
    at Query.resultsetHeader (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/commands/query.js:87:12)
    at Query.Command.execute (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/commands/command.js:38:20)
    at Connection.handlePacket (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/connection.js:263:28)
    at PacketParser.onPacket (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/connection.js:95:66)
    at PacketParser.executeStart (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/packet_parser.js:44:12)
    at Socket.<anonymous> (/Users/jsheeley/git/lumen/api/node_modules/mysql2/lib/connection.js:97:29)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:172:18)
    at Socket.Readable.push (_stream_readable.js:130:10)
    at TCP.onread (net.js:542:20)
```

[...]

PR #341, opened 2016-07-06
Replace bn.js with Long

sheeley:
For issue #336. Not sure if this was the intended change, but tests continue to pass (though lint fails on 0.10 and 0.12 for some reason?!).

### DIFF
--- a/lib/packets/resultset_header.js
+++ b/lib/packets/resultset_header.js
@@ -9,7 +9,7 @@ function ResultSetHeader (packet, bigNumberStrings)
   } else {
     this.fieldCount = packet.readInt8(); // skip OK byte
     this.affectedRows = packet.readLengthCodedNumber(bigNumberStrings);
-    this.insertId = packet.readLengthCodedNumber(bigNumberStrings);
+    this.insertId = packet.readLengthCodedNumberSigned(bigNumberStrings);
     this.serverStatus = packet.readInt16();
     this.warningStatus = packet.readInt16();
   }
--- a/test/integration/connection/test-insert-negative-ai.js
+++ b/test/integration/connection/test-insert-negative-ai.js
@@ -1,0 +1,35 @@
+var common = require('../../common');
+var connection = common.createConnection();
+var assert = require('assert');
+
+// common.useTestDb(connection);
+
+var table = 'insert_negative_ai_test';
+var text = ' test test test ';
+connection.query([
+  'CREATE TEMPORARY TABLE `' + table + '` (',
+  '`id` int(11) unsigned NOT NULL AUTO_INCREMENT,',
+  '`title` varchar(255),',
+  'PRIMARY KEY (`id`)',
+  ') ENGINE=InnoDB DEFAULT CHARSET=utf8'
+].join('\n'));
+
+var result, result2;
+connection.query('INSERT INTO ' + table + ' (id, title) values (-999, "' + text + '")', function (err, _result) {
+  if (err) {
+    throw err;
+  }
+  result = _result;
+  connection.query('SELECT * FROM ' + table + ' WHERE id = ' + result.insertId, function (err, _result2) {
+    result2 = _result2;
+    connection.end();
+  });
+});
+
+process.on('exit', function () {
+  assert.strictEqual(result.insertId, 1);
+  assert.strictEqual(result2.length, 1);
+  // TODO: type conversions
+  assert.equal(result2[0].id, String(result.insertId));
+  assert.equal(result2[0].title, text);
+});
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -1,7 +1,7 @@
 // var BigNumber = require("bignumber.js");
 var ErrorCodeToName = require('../constants/errors.js');
 
-var bn = require('bn.js');
+var Long = require('long');
 
 function Packet (id, buffer, start, end)
 {
@@ -106,15 +106,19 @@ Packet.prototype.eofWarningCount = function () {
   return this.buffer.readInt16LE(this.offset + 1);
 };
 
-Packet.prototype.readLengthCodedNumber = function (bigNumberStrings) {
+Packet.prototype.readLengthCodedNumber = function (bigNumberStrings, signed) {
   var byte1 = this.buffer[this.offset++];
   if (byte1 < 251) {
     return byte1;
   }
-  return this.readLengthCodedNumberExt(byte1, bigNumberStrings);
+  return this.readLengthCodedNumberExt(byte1, bigNumberStrings, signed);
 };
 
-Packet.prototype.readLengthCodedNumberExt = function (tag, bigNumberStrings) {
+Packet.prototype.readLengthCodedNumberSigned = function (bigNumberStrings) {
+  return this.readLengthCodedNumber(bigNumberStrings, true);
+};
+
+Packet.prototype.readLengthCodedNumberExt = function (tag, bigNumberStrings, signed) {
   var word0, word1;
   var res;
   if (tag == 0xfb) {
@@ -142,7 +146,7 @@ Packet.prototype.readLengthCodedNumberExt = function (tag, bigNumberStrings) {
       return word1 * 0x100000000 + word0;
     }
 
-    res = (new bn(word1)).ishln(32).iaddn(word0);
+    res = new Long(word0, word1, signed);
 
     return bigNumberStrings ? res.toString() : res;
   }
--- a/package.json
+++ b/package.json
@@ -28,9 +28,9 @@
   "author": "Andrey Sidorov <sidorares@yandex.ru>",
   "license": "MIT",
   "dependencies": {
-    "bn.js": "4.11.4",
     "cardinal": "0.7.0",
     "double-ended-queue": "2.1.0-0",
+    "long": "^3.1.0",
     "named-placeholders": "1.1.1",
     "readable-stream": "2.1.4",
     "seq-queue": "0.0.5",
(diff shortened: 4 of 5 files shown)


sheeley:
@sidorares any tips on why linting fails on 0.10 / 0.12?

sheeley:
Oh, and to continue my spam - this solves the issue in #336

sidorares:
I think recent eslint started to use arrow functions, 3.0 definitely worked. Can you try to lock exact version ( without caret or ~ ) ?

sidorares:
Did that help with original issue?

sidorares:
eslint 3.0.1 pr from greenkeeper failed today: https://github.com/sidorares/node-mysql2/pull/340

If 3.0.0 won't help, could you try to roll back to latest eslint 2.x?

sheeley:
Locked to 2.13.1 and it appears to be working (though these tests... 15 minutes?!). Hopefully will see some green checkboxes shortly. :)

sidorares:
tests run about 1.5 minute on both travis and appveyor but often there is a big waiting time in the queue ( especially for appveyor )

sidorares:
I want to do a bit more research about allowing negative value in length encoded int. I think everywhere in the protocol 'length encoded int' assumed to be unsigned ( it's mostly used as "how much data we expect?" value - string length prefixes, packet size etc etc ). Maybe we should restrict usage to lastInsertId only ( e.i call it readLengthCodedNumberSigned or something like this and use only at https://github.com/sidorares/node-mysql2/blob/7b4f535f5977537e639785f822fd88ab7cb9aaa2/lib/packets/resultset_header.js#L11 )

sheeley:
I'm pretty unfamiliar with the protocol; are you saying remove the `Long` within `readLengthCodedNumberExt`, or just that it should be unsigned? Sorry I'm not much help, but happy to keep sending commits if you can direct me. :)

sidorares:
Maybe have unsigned Long in standard readLengthCodedNumberExt and signed in a special version `readLengthCodedNumberSigned`, just because I think that particular field ( next insert id ) is exception to protocol ( or a bug in protocol documentation, I don't know )

sheeley:
Does this look like what you were thinking?

sidorares:
yes, I think it's much safer to start with this ( only scoped to resultset_header packet )
not sure if readLengthCodedNumberSigned should be used for affectedRows. Can you think of any scenario when negative value is returned here?

sheeley:
nope - that was just me reading wrong. updated. :)

sidorares:
oh, one more thing

I think the test should be updated to reflect original issue you were having ( e.i negative value in insertId )

Could you change it so that initial value is say -999, then something is inserted, we expect -999, then set to actually big number ( outside of 2^53, as in current test). The last part is already there, we need to add 'negative insert id' to test. Could be a new file or addition to existing `test/integration/connection/test-insert-bigint.js` ( Maybe new even better - `test/integration/connection/test-negative-insert-id.js` )

sheeley:
I'm not sure if I understand - our `AUTO_INCREMENT` initial value isn't negative, but `0`; we insert values like `-999` to provide non-null foreign keys, so the `AUTO_INCREMENT` shouldn't be impacted. I added a test for inserting negative ids working, but it doesn't look like you get back the `LAST_INSERT_ID` when you specify the id.

```sql
INSERT INTO audience_lkp(aud_id, aud_nm, ins_user)
VALUES
(-998, 'Unknown', 'seed-data');

select LAST_INSERT_ID(); -- => 0

insert into audience_lkp(aud_nm, ins_user) values ('wat', 'huh');
select LAST_INSERT_ID(); -- => 8 (or whatever id)
```

sidorares:
The flag in Long should be true for unsigned - see https://github.com/dcodeIO/long.js#longfrombitslowbits-highbits-unsigned

and for test we need to have something like 

```js
c.query("INSERT INTO audience_lkp(aud_id, aud_nm, ins_user) VALUES (-123, 'Unknown123', 'seed-data555')", function (err, res) {
  assert.equal(res.insertId, -123);
});
```

lib/packets/resultset_header.js
```diff
@@ -9,7 +9,7 @@ function ResultSetHeader (packet, bigNumberStrings)
   } else {
     this.fieldCount = packet.readInt8(); // skip OK byte
     this.affectedRows = packet.readLengthCodedNumber(bigNumberStrings);
-    this.insertId = packet.readLengthCodedNumber(bigNumberStrings);
```
sidorares: for values that fit into JS number ( like `-123` )we need to convert to js number, something like this:

```
    var insertIdString = insertId.toString(10);
    var insertIdNumber = insertId.toNumber();
    if (bigNumberStrings) {
      this.insertId = insertIdString;
    } else if (insertIdNumber.toString() === insertIdString) {
      this.insertId = insertIdNumber;
    } else {
      this.insertId = insertIdString;
    }
```

test/integration/connection/test-insert-negative-ai.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+  result = _result;
+  connection.query('SELECT * FROM ' + table + ' WHERE id = ' + result.insertId, function (err, _result2) {
+    result2 = _result2;
+    connection.end();
+  });
+});
+
+process.on('exit', function () {
+  assert.strictEqual(result.insertId, 1);
```
sidorares: we should expect `-999` here (as js number)

sheeley:
I don't think that we should expect `-123` or `-999` - that's why I posted the SQL snippet above. This is what the mysql docs say:

> When a new AUTO_INCREMENT value has been generated, you can also obtain it by executing a SELECT LAST_INSERT_ID() statement with mysql_query() and retrieving the value from the result set returned by the statement.

My reading of that says "because you're specifying `-123` or `-999` as the id, it's not creating a new AUTO_INCREMENT value. Is that what gets returned? If not, can you point me at the right docs? 

Will update the `signed` thing, silly miss on my part.

sidorares:
>  Is that what gets returned?
Yes, this is what actually returned in insertId field: if the value is auto increment (and not specified) it'll be the same as  LAST_INSERT_ID(), and when you specify it's returned as is, as in my example

```js
c.query("INSERT INTO audience_lkp(aud_id, aud_nm, ins_user) VALUES (-123, 'Unknown123', 'seed-data555')", function (err, res) {
  // I can actually see length encoded int:  0xfe tag (8 byte long) and 8 bytes representing -123
  // resultset_header packet: 0001fe85ffffffffffffff02000000 - http://dev.mysql.com/doc/internals/en/packet-OK_Packet.html
  // Long.js words1 & word2: 4294967173, 4294967295
  // new Long(4294967173, 4294967295, false) => -123 . 'false' means "not unsigned"
  assert.equal(res.insertId, -123);
});
```

sidorares:
doing this ( note BIGINT for aud_id )
```
CREATE TABLE audience_lkp1 (
  aud_id                BIGINT                NOT NULL PRIMARY KEY AUTO_INCREMENT,
  aud_nm                VARCHAR(50)        NOT NULL,

  ins_ts                TIMESTAMP          DEFAULT CURRENT_TIMESTAMP,
  ins_user              VARCHAR(80)        NOT NULL,
  upd_ts                TIMESTAMP          ON UPDATE CURRENT_TIMESTAMP,
  upd_user              VARCHAR(80)
)

INSERT INTO audience_lkp1(aud_id, aud_nm, ins_user) VALUES (18446744073709551493, 'Unknown123', 'seed-data555')
```
returns 9223372036854775807 for insertId, and next time I do `INSERT INTO audience_lkp1(aud_nm, ins_user) VALUES ('Unknown123', 'seed-data555')` ( without aud_id ) it errors with `Duplicate entry '9223372036854775807' for key 'PRIMARY'` ( and last_insert_id() returns 0 )

sidorares:
http://dev.mysql.com/doc/refman/5.7/en/example-auto-increment.html

http://stackoverflow.com/questions/11264447/what-is-the-limit-of-auto-increment-integer-in-mysql

sheeley:
Interesting - there are a couple of spots I'll update solo, but will likely need to check back in on the insertId stuff (let me read the mysql docs first though). Thanks!

sidorares:
thanks @sheeley ! sorry it goes back and forth but the issue itself is in the dark area of bad documentation and edge cases :)

sidorares:
note that #342 changed eslint in package.json to be able to pass tests so you might need to rebase on top

sidorares:
hi @sheeley - do you need a hand with this? I'd like to have your changes in 1.0 release and also have plans for similar fix in binary protocol ( `.execute()` with longlong results )

sheeley:
hey @sidorares, sorry about the delay on this; been a busy month. i can keep chugging away at it but you may be much faster - if i get time, i'll try to fix up this week

sidorares:
No problems at all @sheeley

sidorares:
@sushantdhiman - maybe you can also help with this. Main issue was in incorrect handling of negative insertId, this PR fixes it but I think needs a bit more work on test. I also might be able to look at this later today or tomorrow

sushantdhiman:
@sidorares I have a tight schedule today :) I'll see if I can work on this at night.

sidorares:
I'm not pushing :)

sushantdhiman:
@sidorares It seems `MySQL 5.7` will not accept the negative AI numbers.

```js
Error: Out of range value for column 'id' at row 1
      at Packet.asError (/var/www/node-mysql2/lib/packets/packet.js:547:13)
      at Query.Command.execute (/var/www/node-mysql2/lib/commands/command.js:28:22)
      at Connection.handlePacket (/var/www/node-mysql2/lib/connection.js:261:28)
```

sushantdhiman:
Ok my bad, MySQL 5.7 need the signed version of `id` :)

PR #349, opened 2016-07-21
Failing test for geometry typeCast issue

sushantdhiman:
`Work in Progress + Test Case`

While parsing the Geometry values using the `typeCast` some odd error occurs.

```js
> node ./test/run.js && MYSQL_USE_COMPRESSION=1 node ./test/run.js

[0:00:00 0 0/1 0.0% node test/integration/connection/test-typecast-geometry.js]
  
  evalmachine.<anonymous>:11
        geometry: function () { return this.parser.parseGeometryValue(); }
                                                  ^
  
  TypeError: Cannot read property 'parseGeometryValue' of undefined
      at Object.geometry (evalmachine.<anonymous>:11:49)
      at Object.connection.query.typeCast (/var/www/node-mysql2/test/integration/connection/test-typecast-geometry.js:9:20)
      at new TextRow (evalmachine.<anonymous>:15:25)
      at Query.row (/var/www/node-mysql2/lib/commands/query.js:197:13)
      at Query.Command.execute (/var/www/node-mysql2/lib/commands/command.js:38:20)
      at Connection.handlePacket (/var/www/node-mysql2/lib/connection.js:263:28)
      at PacketParser.onPacket (/var/www/node-mysql2/lib/connection.js:95:66)
      at PacketParser.executeStart (/var/www/node-mysql2/lib/packet_parser.js:44:12)
      at Socket.<anonymous> (/var/www/node-mysql2/lib/connection.js:97:29)
      at emitOne (events.js:96:13)
  
[0:00:00 1 0/1 100.0% node test/integration/connection/test-typecast-geometry.js]
```

As per my understandings, this error occurs when we try to call `field.value(), field.buffer(), field.geometry()` more than one time. But as you can see in test case I have called `field.geometry()` exactly one time.

This issue doesn't appear with other datatypes afaik.

### DIFF
--- a/lib/compile_text_parser.js
+++ b/lib/compile_text_parser.js
@@ -21,8 +21,8 @@ function compile (fields, options, config) {
       table: field.table,
       name: field.name,
       string: function() { return packet.readLengthCodedString(); },
-      buffer: function () { return this.parser.parseLengthCodedBuffer(); },
-      geometry: function () { return this.parser.parseGeometryValue(); }
+      buffer: function () { return packet.parseLengthCodedBuffer(); },
+      geometry: function () { return packet.parseGeometryValue(); }
     };
   };
 
--- a/lib/compile_text_parser.js
+++ b/lib/compile_text_parser.js
@@ -21,7 +21,7 @@ function compile (fields, options, config) {
       table: field.table,
       name: field.name,
       string: function() { return packet.readLengthCodedString(); },
-      buffer: function () { return packet.parseLengthCodedBuffer(); },
+      buffer: function () { return packet.readLengthCodedBuffer(); },
       geometry: function () { return packet.parseGeometryValue(); }
     };
   };
--- a/test/integration/connection/test-typecast-geometry.js
+++ b/test/integration/connection/test-typecast-geometry.js
@@ -15,4 +15,17 @@ connection.query({
   assert.deepEqual(res[0].foo, { x: 11, y: 0 });
 });
 
+connection.query({
+  sql: 'select GeomFromText(\'POINT(11 0)\') as foo',
+  typeCast: function (field, next) {
+    if (field.type === 'GEOMETRY') {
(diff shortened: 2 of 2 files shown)


sidorares:
Thabks @sushantdhiman ! I think the problem is here - https://github.com/sidorares/node-mysql2/blob/b534e15faa0dbeefe466ad5a53e4e7fdf378aee8/lib/compile_text_parser.js#L23-L25

```diff
       name: field.name,
       string: function() { return packet.readLengthCodedString(); },
-      buffer: function () { return this.parser.parseLengthCodedBuffer(); },
-      geometry: function () { return this.parser.parseGeometryValue(); }
+      buffer: function () { return packet.parseLengthCodedBuffer(); },
+      geometry: function () { debugger; return packet.parseGeometryValue(); }
     };
   };
```

Not sure the test is correct though - you expect buffer but typeCast function parses into point object

sushantdhiman:
Thanks for quick reply and fix :) @sidorares 

Test is just a demonstration, Ultimately I want to use buffer value and pass it to a `wkb` library in my `typeCast` function. I'll make suggested changes soon.

sidorares:
LGTM!

sushantdhiman:
One change it should be `readLengthCodedBuffer` instead of `parseLengthCodedBuffer` , just commited. Lets wait for CI

sushantdhiman:
Just to be sure, for buffer parsing there is no method `parseLengthCodedBuffer` so I have changed it to `readLengthCodedBuffer`, otherwise calling `field.buffer()` will fail (test included).

If thats right way, then its ready for merge @sidorares

sidorares:
yes, that was my mistake copy-pasting from node-mysql
I have functions named "parse*" when they do some complicated parsing and "read*" when just simple read from incoming buffer

sushantdhiman:
Its ready for merge then :)

Issue #348, opened 2016-07-20
JSON_EXTRACT comparison/casting issue


KevinMartin:
Using MySQL 5.7, it seems like calling `JSON_EXTRACT(field, path+)` for an integer from a json array isn't allowing me to compare with an integer.

Query:
```sql
SELECT *
FROM table
WHERE JSON_EXTRACT(json, "$[2]") >= ?;
```

The value in the `json` field is just a simple `[1,2,3,4,5]`.

I try executing the query with both a string and integer value. Both result in now results.

If I insert the query into MySQL Workbench, it works fine. If I screw the prepared statement route and just put the integer in the query in place of the `?`, it works fine.

They way around it so far is casting the `JSON_EXTRACT` value into `UNSIGNED` for the comparison to work:
```sql
SELECT *
FROM table
WHERE CAST(JSON_EXTRACT(json, "$[2]") AS UNSIGNED) >= ?;
```

sidorares:
could you post schema creation code for table? ( or just `show create` sql)?

sidorares:
also, do are you using `query()` or `execute()` workbench is probably not using prepared statements

KevinMartin:
I'm using `execute()`.

```sql
CREATE TABLE `table` (
  `json` json NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

sidorares:
cold you also help me to populate table? Can you post simple insert queries to match your data?

KevinMartin:
```sql
INSERT INTO `table` (`json`) VALUES ('[1,2,3,4,5]');
```

KevinMartin:
The values is always an array of integers.

sidorares:
works for me:

```js
conn.query('SELECT * FROM tttable WHERE JSON_EXTRACT(json, "$[1]") >= ?', [5], (err, rows) => {
  console.log(err, rows);
});
```

prints `null [ TextRow { json: [ 100, 100, 3, 4, 5 ] } ]`

```
mysql> select * from tttable;
+---------------------+
| json                |
+---------------------+
| [1, 2, 3, 4, 5]     |
| [100, 100, 3, 4, 5] |
+---------------------+
2 rows in set (0.00 sec)

```

KevinMartin:
Weird. Any other info I can provide you? Doesn't work for me.

sidorares:
I'm using mysql server version 5.7.10. Do you have issues when doing same queries from command line?

```
[...]

PR #353, opened 2016-07-28
Add some parameter type checking for prepared statements

KevinMartin:
Whenever I use anything besides `Types.DECIMAL` for number type, I get `Error: Incorrect arguments to mysqld_stmt_execute`.

Related to: #348

### DIFF
--- a/lib/packets/execute.js
+++ b/lib/packets/execute.js
@@ -99,8 +99,13 @@ Execute.prototype.toPacket = function ()
         case 'object':
           if (this.parameters instanceof Date) {
             packet.writeInt16(Types.DATETIME);
+          } else if (
+            Array.isArray(this.parameter[i]) ||
+            this.parameters[i].constructor === Object ||
+            typeof this.parameters[i].toJSON === 'function'
+          ) {
+            packet.writeInt16(Types.JSON);
           } else {
-            // packet.writeInt16(Types.JSON); - Requires MySQL 5.7+
             packet.writeInt16(Types.VAR_STRING);
           }
           break;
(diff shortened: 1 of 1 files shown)


lib/packets/execute.js
```diff
@@ -82,11 +82,37 @@ Execute.prototype.toPacket = function ()
...
+        case 'boolean':
+          this.parameters[i] = +this.parameters[i];
+          packet.writeInt16(Types.TINY);
+          break;
+
+        case 'object':
+          if (this.parameters instanceof Date) {
+            packet.writeInt16(Types.DATETIME);
+          } else if (
+            Array.isArray(this.parameter[i]) ||
```
sidorares: typo here

sidorares:
Thanks @KevinMartin !
Could you add a test ( with `execute` +  number + string + array + date parameters ) ?

KevinMartin:
Any guidance?

sidorares:
Unfortunately mysql server version on TravisCI does not support json ( maybe this can be overriden, but right now .travis.yml uses default setup ). Maybe you add simple select with parameters, using at least numeric and string parameter types
Test examples:
https://github.com/sidorares/node-mysql2/blob/master/test/integration/connection/test-execute-newdecimal.js
https://github.com/sidorares/node-mysql2/blob/master/test/integration/connection/test-execute-null-bitmap.js
https://github.com/sidorares/node-mysql2/blob/master/test/integration/connection/test-execute-signed.js

Actually it looks like it's already covered ( with exception for JSON case ) Maybe you add your original JSON test, but wrap it with a sql check to execute only when JSON type supported?

PR #356, opened 2016-07-30
Catching errors in once crashes process

### LINKED ISSUE #355: Properly catching invalid connection init level errors
`createConnection` connect with database as soon its called. So you can query directly like

```js
let connection = mysql2.createConnection({});

connection.query('SELECT 1+1');
```

But this pose an issue, say when you pass invalid credentials. It will throw an async error which cannot be caught unless we use `process.uncaughtException` or `domains`.

Is there any way we can handle this error properly without using hacks. This is also a deviation 
from `node-mysql` API, which allow creating connections without actually connecting.

May be we should add a version of `createConnection` which does not connect with server unless we explicitly call `connect`. So if `connect` fails we can handle any errors properly.

sushantdhiman:
This will NOT crash the process
```js
var connection = mysql2.createConnection({ invalid credentials });
connection.on('error', handler);
```

But this will
```js
var connection = mysql2.createConnection({ invalid credentials });
connection.once('error', handler);
```

When error are handled in `once` they still bubble up to process level.

**Fixes**

- Handling errors in either `on` or `once` work now
- No multiple error events if handshake resulted in error

Closes #355

### DIFF
--- a/lib/commands/client_handshake.js
+++ b/lib/commands/client_handshake.js
@@ -68,10 +68,10 @@ ClientHandshake.prototype.calculateNativePasswordAuthToken = function (authPlugi
 ClientHandshake.prototype.handshakeInit = function (helloPacket, connection) {
   var command = this;
 
-  this.on('error', function (err) {
-    connection._protocolError = err;
-    connection.emit('error', err);
+  this.on('error', function (e) {
+    connection._protocolError = e;
   });
+
   this.handshake = Packets.Handshake.fromPacket(helloPacket);
   if (connection.config.debug) {
     console.log('Server hello packet: capability flags:%d=(%s)', this.handshake.capabilityFlags,
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -103,19 +103,17 @@ function Connection (opts)
       return;
     }
 
-    // TODO: move to protocolError()
     if (!connection._protocolError) { // no particular error message before disconnect
-      connection._protocolError = 'PROTOCOL_CONNECTION_LOST';
+      connection._protocolError = new Error('Connection lost: The server closed the connection.');
+      connection._protocolError.fatal = true;
+      connection._protocolError.code = 'PROTOCOL_CONNECTION_LOST';
     }
-    var err = new Error('Connection lost: The server closed the connection.');
-    err.fatal = true;
-    err.code = connection._protocolError;
-    connection._notifyError(err);
+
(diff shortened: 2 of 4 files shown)


sidorares:
yes, if I add `on.error((e) => console.log('err'))` I can see it printed 3 times from different places. Need to clean up this.

Try the following example and you'll see who is emitting error:

```js
var connection = mysql2.createConnection({ password: 'foo' });

connection.on('error', (e) => { console.trace() });
```

sidorares:
1) Handshake error is reported more than one - this needs to be fixed
2) Currently with mysql2 the only reliable way to catch errors is to attach 'error' handler to connection. I'm not sure what's the good way to avoid this restriction. Potentially, if there is an error and there is no command currently being processed ( or in the queue ) the connection could go into 'invalid' state and trigger error on next command ( to a command callback ). Advantage: no crash even if users don't listen for connection errors. Disadvantage: errors reported later and not at the moment they occur. Not sure what's current node-mysql in this regard

sushantdhiman:
1) Done
2) `createConnection` start connecting when called, so we have to use `.on` handlers. `node-mysql` `createConnection` doesn't start connecting immediately, we need to call `connect` function. `connect` accepts a callback which reports errors otherwise pass connection. We can look for similar implementation so we can handle handshake errors in a better way.

In my case I expected promise wrapper to work like this

```js
mysql2.createConnectionPromise( { invalid credentails })
.then((connection) => {
   console.log(connection);
})
.catch((err) => {
//Access Denied Error
})
```

But this wasn't the case, I still get `uncaughtException`. But now issue with `error` event upon handshake are resolved so there is no hurry for these changes

sushantdhiman:
Also this PR can be merged. Let me know if there are any changes

sidorares:
Looks good to me! Do you have any thoughts re handling errors in between queries?

sushantdhiman:
`connect` is only good for handling the handshake / init level errors. But if our connection is lost in between queries, it better to handle that at `connection.on` level.

sidorares:
This is the current behaviour (and it assumes you must attach error handler to connection if it's a long running server and you don't want process to crash at some point and not using connection pool)

sushantdhiman:
And this is the correct behavior. Plus now error event will not be called multiple time (on handshake) and will work with `once` too.

sidorares:
great, thanks for your work @sushantdhiman !

Issue #315, opened 2016-05-25
Named Placeholders not working in Pool.Execute


ericvolp12:
Hey, I've been having issues with pooled connections using the standard method of requesting a connection from the pool, setting named placeholders of the connection, executing the query with that connection, then releasing the connection after error or success. The issue I've got is that eventually I get an error when performing any mysql command from that pool that gives me a: `[Error: This socket has been ended by the other party] code: 'EPIPE'`

Now I'm trying to use pool.execute() to achieve the same results but using the pool itself instead of grabbing a single connection from the pool. For some reason, despite my use of:
`pool.config.namedPlaceholders = true;`
I'm getting issues trying to execute my queries where the data is not being prepared into the query before it is sent.

My database accessor class looks like this:
```javascript
//MySQL Configuration
var mysql = require('mysql2');

//Configure the mysql pools for the web database
var pool = mysql.createPool({
    host     : creds.mysql.host,
    user     : creds.mysql.user,
    password : creds.mysql.password,
    database : creds.mysql.db
});

pool.config.namedPlaceholders = true;

var winston = require('winston');


/**
 * Function to prepare and execute a query in the DB
 * @param query - Query to execute
 * @param data - Data to pass into the query's placeholders
 */
function execute(query, data) {
    return new Promise(function (resolve, reject) {
        //Execute the query and pass in the data
        pool.execute(query, data, function (err, rows) {
            if (err) {
                console.log(err);
                winston.log(err);
                reject(err);
            } else {
                resolve(rows);
            }
        });
    });
}
```

And the error I'm getting with a standard query that worked with the old method (but not after like 36 hours when the pool timed out or something), is this:
```
{ Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ':asin LIMIT 1' at line 1
```

The query is this - `"SELECT * FROM listings WHERE asin=:asin LIMIT 1;";`
And the value I'm passing is - `{asin: asin}` where asin is actually a value and I know it is.

Is there anyone who might be able to tell me what's going on? I've been reading the source to see if there's something different about the config but I can't seem to see what I'm doing wrong or whether this is an error.

Thanks!

- Eric

sidorares:
What's the version of module? The issue might be related to #281 - this was merged recently, could you try master?

ericvolp12:
I guess in my package.json I'm using "mysql2": "^1.0.0-rc.1"

ChiperSoft:
[...]

PR #369, opened 2016-08-09
Fix pool namedplaceholders on execute

ChiperSoft:
Fix for issue #315 

`this.config` on a pool does not contain the namedPlaceholders option, it now lives inside `this.config.connectionConfig`. Thus, namedPlaceholders was always being set to undefined.

I also removed the argument swapping in Pool.execute. This conversion occurs inside connection.execute and does not need to be performed at the pool level. Doing it there actually breaks passing an options object instead of a query.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -145,11 +145,6 @@ Pool.prototype.query = function (sql, values, cb) {
 };
 
 Pool.prototype.execute = function (sql, values, cb) {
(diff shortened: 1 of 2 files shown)


CacheControl:
👍

ChiperSoft:
Working on the CI break. I didn't think there were any tests on pools...

sidorares:
👍

PR #372, opened 2016-08-13
pool connection time error is not propagated correctly to users and as a result process crashes

sidorares:
see test

might be a side effect of #356,  CCing @sushantdhiman

during `pool.query` or `pool.getConnection` if connection is created (e.i not reused) this part of code is executed: https://github.com/sidorares/node-mysql2/blob/26748034feef277d56f5d97907c4790591f307ec/lib/pool.js#L46

but for some reason, if there is connect time error (bad password or 'too many connections' error from ClientHandshake command is not returned - "error" is emitted on ClientHandshake and there is no error listener there, thus crashing process 

see also:
commands error handling: https://github.com/sidorares/node-mysql2/blob/26748034feef277d56f5d97907c4790591f307ec/lib/commands/command.js#L29-L33
Initial ClientHandshake command: https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L105-L110

I think previously for commands with no error handler (and no onResult callback) error would be routed to connection

### DIFF
--- a/lib/commands/client_handshake.js
+++ b/lib/commands/client_handshake.js
@@ -75,6 +75,7 @@ ClientHandshake.prototype.handshakeInit = function (helloPacket, connection) {
   var command = this;
 
   this.on('error', function (e) {
+    connection._fatalError = e;
     connection._protocolError = e;
   });
 
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -108,6 +108,9 @@ function Connection (opts)
       connection.threadId = handshakeCommand.handshake.connectionId;
       connection.emit('connect', handshakeCommand.handshake);
     });
+    handshakeCommand.on('error', function(err) {
+      connection._notifyError(err);
+    });
     this.addCommand(handshakeCommand);
   }
 }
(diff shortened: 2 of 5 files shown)


sushantdhiman:
We might need to handle errors in Command just like the Client Handshake.(https://github.com/sidorares/node-mysql2/blob/26748034feef277d56f5d97907c4790591f307ec/lib/commands/client_handshake.js#L78)

sidorares:
Looks like main cause was this ( `pool.query` shortcut ): https://github.com/sidorares/node-mysql2/blob/26748034feef277d56f5d97907c4790591f307ec/lib/pool.js#L133

If you have callback but not explicitly attach error to command this causes crash. Fix in commit https://github.com/sidorares/node-mysql2/pull/372/commits/33617dc8804cc03c7d96f104666cdfe736a39193

sidorares:
@sushantdhiman I added you as collaborator if you don't mind ( and to npm as well )

sushantdhiman:
Thanks @sidorares :)

sushantdhiman:
LGTM

Kennytian:
`SET GLOBAL wait_timeout = 60; `

run this command in mysql client command terminal, the timeout unit is seconds

PR #374, opened 2016-08-15
Fixes #302

### LINKED ISSUE #302: fix character encoding handling and use UTF8MB4_UNICODE_CI by default
Basically, we assume everything is utf8 because this is what client tells server during connection.
This is not enough, because
  - server is able to ignore / override this
  - different columns in result set could be using own charsets. Column charset is part of [column definition packet](http://dev.mysql.com/doc/internals/en/com-query-response.html#column-definition)

see https://github.com/felixge/node-mysql/pull/1408
and https://github.com/felixge/node-mysql/issues/804

would really like to have this as part of 1.0 because this change is semver-major

sushantdhiman:
**Tasks**
- [x] Test for encoding issues
- [x] Grab `charset` and install `iconv-lite`
- [x] Finding reason for corrupt `INSERT INTO` and fixing them
- [x] Use iconv for parsing string to proper mysql format (`cesu8`)
- [x] Prepare map between charset and encodings
- [x] Use iconv for parsing string to proper mysql format (actual mapped charset)
- [x] Use charset from config (connection, server) for proper encoding
- [x] Integration tests for connection, server charset


sidorares:
@sushantdhiman see also https://github.com/mysqljs/mysql/pull/1408#issuecomment-224798992

( and https://github.com/sidorares/node-mysql2/commit/488e0e6e5e71aa057f65535730e4fa79755bfa80 )

sushantdhiman:
`UTF8MB4_UNICODE_CI` is working as expected. In previous tests table was created with `latin1` and we tried to insert non-latin characters into it thus resulting `????`

Although one problem remains, `fieldName` ( [failing test](https://github.com/sidorares/node-mysql2/blob/313bd02d984ae9a07d82eeaacda92b14c666e5e5/test/integration/connection/test-select-utf8mb4.js) ) are not parsed properly, to solve this I need to access `charset` and pass it to `readLengthCodedString`, where I can use `iconv-lite` to do the conversion if `charset` is passed. Main issue is packet data is accessed in some serial offset based protocol [here](https://github.com/sidorares/node-mysql2/blob/313bd02d984ae9a07d82eeaacda92b14c666e5e5/lib/packets/column_definition.js#L43), I need to access `this.charCode` a few steps back [here](https://github.com/sidorares/node-mysql2/blob/313bd02d984ae9a07d82eeaacda92b14c666e5e5/lib/packets/column_definition.js#L36)

sidorares:
"some serial offset based protocol" - this actually helps here ( to overcome the fact that charset is further down packet and we need to calculate it's position before we start decoding schema/catalog/table/orgTable ) The lazy accessor is used to avoid unnecessary buffer->string conversion if schema/catalog/table/orgTable are not used

We still to use `characterSet` to decode name, so I suggest we'll have something like this:

```js

  function readString(buffer, start, end, charset) {
     // add iconv charset conversion and respect charset parameter. Optimise in a way that it's likely called with utf8 most of the time
     return buffer.utf8Slice(start, end);
  }

  // name is always used, don't make it lazy
  // this.name = packet.readLengthCodedString();
  var _nameLength = packet.readLengthCodedNumber();
  var _nameStart = packet.offset;

  this._orgNameLength = packet.readLengthCodedNumber();
  this._orgNameStart = packet.offset;
  packet.offset += this._orgNameLength;

  packet.skip(1); //  length of the following fields (always 0x0c)
  this.characterSet = packet.readInt16();

  // now decode name once we have characterSet
  this.name = readString(this._buf, _nameStart, _nameStart + _nameLengt, this.characterSet);

  // modify getters code to respect charset as well

  var addString = function (name) {
  Object.defineProperty(ColumnDefinition.prototype, name, {get: function () {
    var start = this['_' + name + 'Start'];
    var end = start + this['_' + name + 'Length'];
    return readString(this._buf, start, end, this.characterSet);
  }});
};
```

sushantdhiman:
Now I just need to prepare a proper map between [charset codes](https://github.com/sidorares/node-mysql2/blob/master/lib/constants/charsets.js) and [iconv-encodings](https://github.com/ashtuchkin/iconv-lite/wiki/Supported-Encodings), this one will need a bit of research work to see which charset belongs to which encoding

sidorares:
we need to add conversion to column value itself:

https://github.com/sidorares/node-mysql2/blob/b2010f99180d8b49eb7e9c2b6aeb8cb441362835/lib/compile_text_parser.js#L149
https://github.com/sidorares/node-mysql2/blob/b2010f99180d8b49eb7e9c2b6aeb8cb441362835/lib/compile_binary_parser.js#L153

Also maybe worth creating decoder only once in row parser? (might be tricky as we potentially need to have one decoder per field - not sure if overhead worth complexity, and this is just for non-utf8 users)

https://github.com/ashtuchkin/iconv-lite/blob/69a25dc19c67de1b022ccfafbc5e41c2114f3486/lib/index.js#L36

sushantdhiman:
After testing with Wireshark and studying MySQL docs. Here are a few points to consider

MySQL doesn't support `UTF8MB4` for table meta data ([Their Docs](https://dev.mysql.com/doc/refman/5.6/en/charset-metadata.html))

Every meta data is internally stored in form of `UTF8`, thus [this test](https://github.com/sidorares/node-mysql2/blob/313bd02d984ae9a07d82eeaacda92b14c666e5e5/test/integration/connection/test-select-utf8mb4.js) will always fail on UTF8MB4. Which I think try to convert multibyte data to some `UTF8` representation and comes up with `????`

**Using CLI [MySQL 5.5 , 'UTF8MB4']**
mysql utf8mb4 multibyte select

**Using CLI [MySQL 5.5 , 'UTF8']**
utf8 cli

Apart from field meta data all selection will work fine. Everything is handled by MySQL. But this field meta data conversion is done at MySQL level we can't do anything about it.

Even if we connect with `UTF8` charset and switch to `UTF8MB4` later on this effect will remains, thus converting any multibyte data to `????` in all cases (for field meta data with `utf8mb4`)

I have modified said test so we use some proper field name to access the data. 

Need your thoughts on this @sidorares , Unless MySQL doesn't start handling this case properly can we do anything about it ?

sidorares:
@dougwilson - would be really great if you can add your opinion, I think you have much more experience on mysql side

@sushantdhiman I'll try to have close look tomorrow and give some feedback

sidorares:
@sushantdhiman not sure if meta data ([Their Docs](https://dev.mysql.com/doc/refman/5.6/en/charset-metadata.html)) docs are still valid. What I see is when UTF8_GENERAL_CI is set `SELECT "💩"` query returns `[ TextRow { '💩': '💩' } ]` row, and both column name and column content transmitted as 4 byte utf8 character - 0xF0 0x9F 0x92 0xA9. When `UTF8MB4_UNICODE_CI` is used the column data is sent as same 4 bytes, but metadata (column name) is replaced with `?` character.

sushantdhiman:
Thats what bugs me too, isnt `UTF8MB4` supposed to be a superset of `UTF8`, then why meta data is returned (converted) as `???`

sidorares:
I would expect opposite, this is how I read documentation

- Mysql UTF8 is not actually UTF8, it's a subset of utf8 for code points encoded with up to 3 bytes
- UTF8MB4 is "normal", standard utf8.
- Field names still not allowed to contain chars outside of BMP ( 3 bytes encoding ), whether it's UTF8 or UTF8MB4

The reality seems to be very different

sidorares:
> then why meta data is returned (converted) as ???

This is kind of expected as per https://dev.mysql.com/doc/refman/5.6/en/charset-metadata.html (it's converted to ? because character is outside of BMP)

What I don't understand is why column name is _not_ converted to ??? when UTF8_GENERAL_CI charset is selected

sushantdhiman:
> What I don't understand is why column name is not converted to ??? when UTF8_GENERAL_CI charset is selected

Exactly, why UTF8 is able to convert it but UTF8MB4 isnt. Its exact opposite if we consider UTF8MB4 is capable of handling 4 byte where UTF8 isnt.

May be in case of UTF8 everything is already in 3 bytes so that data fits easily, in case of UTF8MB4 they try to convert 4 byte representation to 3 bytes , resulting loss of data thus giving us `???`

sidorares:
> May be in case of UTF8 everything is already in 3 bytes

No, I thought it might split it into 2 two bytes surrogate pairs but later realised that this is for utf-16 only. And bytes actually sent over the wire are  0xF0 0x9F 0x92 0xA9 for one single unicode character

sidorares:
I'll try to test this against older versions of server

sidorares:
iconv-lite encodings list: https://github.com/ashtuchkin/iconv-lite/wiki/Supported-Encodings

dougwilson:
Very odd indeed, though UTF-8 in MySQL is a mess in some regards... I have never personally even tried to use non-ASCII identifiers, partly because it's easier, and partly because I'm just English-centric :) It does seem odd that UTF8BM4 would not work in column metadata, though UTF8 would. If it helps as well, in MySQL, what is called "UTF8" is actually the "CESU-8" encoding and what is called "UTF8MB4" is actually the "UTF-8" encoding.

Both support characters outside the BMP just fine (not sure on metadata), though of course if the charset in MySQL is set to UTF8 then the driver will need to send CESU-8 encoded bytes rather than UTF-8 encoded bytes or non-BMP chars will get mangled. UTF8MB4 provides better support for non-BMP chars in that it understands them as one character instead of being made of two different characters (JavaScript, C#, and some other langs also consider non-BMP to be two characters).

sidorares:
here is what I think is going on:
1) with `utf8_unicode_ci` when character is outside of BMP server treats is like unknown 4 (or more) bytes stream. When received back on node side those 4 bytes assembled back into single codepoint. This applies to data and metadata. This is why field name chars are preserved.
2) with `utf8mb4_unicode_ci` server is aware of characters outside of BMP and treats them as individual codepoints. For metadata, characters outside of BMP are replaced with `?` ( 0x3f ) beacause of [some internal restrictions](https://dev.mysql.com/doc/refman/5.6/en/charset-metadata.html)

Example:

```js
var connection = require('mysql2').createConnection({
  charset: 'utf8mb4_unicode_ci'
});
connection.query('SELECT CHAR_LENGTH("𝌆") ', (err, rows, fields) => console.log(rows, fields));
```

returns ( note that CHAR_LENGTH("𝌆") is 1 )

```
[ TextRow { 'CHAR_LENGTH("?")': 1 } ] [ { catalog: 'def',
    schema: '',
    name: 'CHAR_LENGTH("?")',
    orgName: '',
    table: '',
    orgTable: '',
    characterSet: 63,
    columnLength: 10,
    columnType: 8,
    flags: 129,
    decimals: 0 } ]
```

While same script, but with `utf8_unicode_ci` charset connection option results in 

```
[ TextRow { 'CHAR_LENGTH("𝌆")': 4 } ] [ { catalog: 'def',
    schema: '',
    name: 'CHAR_LENGTH("𝌆")',
    orgName: '',
    table: '',
    orgTable: '',
    characterSet: 63,
    columnLength: 10,
    columnType: 8,
    flags: 129,
    decimals: 0 } ]
```
Note that [CHAR_LENGTH](http://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_char-length) returned is 4, same as [LENGTH](http://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_length) would return.

With this in mind `utf8mb4_unicode_ci` does look like a safer default option.

sidorares:
@dougwilson 
> Both support characters outside the BMP just fine (not sure on metadata), though of course if the charset in MySQL is set to UTF8 then the driver will need to send CESU-8 encoded bytes rather than UTF-8 encoded bytes or non-BMP chars will get mangled

I don't think this is true, looks like when non-mb4 utf8 charset is selected and character is encoded using more than 3 octets server gets confused and just treats them as 4 or more characters

dougwilson:
That is because you are not encoding properly over the wire. You need to encode in CESU-8 when using the UTF8 charset.

sidorares:
You mean when the data is sent to server? Because when I receive it back it looks like it's normal utf8

sidorares:
https://github.com/ashtuchkin/iconv-lite/issues/102 :)

dougwilson:
In CESU-8, a non-BMP character is encoded as six bytes, not four. This is because it is encoded as a pair of three byte sequences. CESU-8 is a dumb encoding used in the late 90s/early 00s and made it's way into various dbs like MySQL. They went back and defined a charaet that uses UTF-8 and called it UTF8MB4. But otherwise, when using the UTF8 charset, you need to use CESU-8 encoding over the wire.

dougwilson:
lol, yea, that iconv-lite issue was for talking to an existing MySQL server that contained CESU-8 data :)

sidorares:
This is what I get now when query is sent CESU-8 encoded:

`'SELECT CHAR_LENGTH("𝌆") '` =>

```
[ TextRow { 'CHAR_LENGTH("������")': 2 } ]
```

`'SELECT "test 𝌆 💩" '` =>

```
[ TextRow { 'test ������ ������': 'test ������ ������' } ]
```

dougwilson:
The CHAR_LENGTH looks right. Are you using a CESU-8 decoder to decode as well?

sidorares:
oops, no. utf8

sidorares:
yay, `[ TextRow { 'test 𝌆 💩': 'test 𝌆 💩' } ]` cesu8 <-> cesu8

sidorares:
@sushantdhiman so when non-mb4 utf8 is selected the correct way to write string is

```js 
  var buf = require("iconv-lite").encode(s,"cesu8");
  this.buffer.copy(buf, this.offset);
  this.offset += buf.lengh;
```

and read as

```js
Packet.prototype.readLengthCodedString = function () {
  var len = this.readLengthCodedNumber();
  // TODO: check manually first byte here to avoid polymorphic return type?
  if (len === null) {
    return null;
  }
  this.offset += len;
  var buf = this.buffer.slice(this.offset - len, this.offset);
  return require("iconv-lite").decode(buf,"cesu8");
};
```

sidorares:
> The CHAR_LENGTH looks right

why? It's one code point, 2 surrogate pairs, 6 cesu-8 bytes. Should it return 1?

dougwilson:
> > The CHAR_LENGTH looks right

> why? It's one code point, 2 surrogate pairs, 6 cesu-8 bytes. Should it return 1?

That is because in CESU-8 (and the MySQL UTF8 charset), it works similar to how things like C3 & JavaScript work, where it is seeing surrogate pairs and counting them as two separate characters. This is because a lot of these things using UCS-2 internally instead of UTF-16. Typically systems that were created in the days of UCS-2 also use CUES-8 encoding for this reason.

sushantdhiman:
@sidorares Need your review on this, Now I have used `iconv-lite` to properly encode strings

sidorares:
@sushantdhiman was just going to start adding code myself and you seem to complete it :) I'll have a look now

sidorares:
so currently you always encode/decode as cesu8?

I'll need to check, but I think utf8mb8 does not understand cesu8 , int's "normal" utf8 ( maybe with exception of metadata strings )

Do you think we'll add other encodings later? 
https://dev.mysql.com/doc/refman/5.7/en/charset-connection.html

as far as I understand we need to 1) always encode JS strings using encoding that client specified during connection 2) always decode using encoding specified in result set header (It's always present) 3) would be good to set default to utf8mb4 only if LC_LANG env is not set (or set to utf8)
4) need to add some integration tests by setting `character_set_client`/ `character_set_connection` / `character_set_results` mysql variables to something other than utf8

I'll prepare mapping mysql collation/charset number -> charset name ( it's partially available as result of `show charset` sql, I need to add numbers from constants/charsets.js and also map collations pointing to the same mysql charset -> same charset number )

Overall above might be too big for single pr, I'm happy to break it in two

sidorares:
@sushantdhiman let me know what you think and I can continue from where you stopped

sushantdhiman:
> so currently you always encode/decode as cesu8?
> 
> I'll need to check, but I think utf8mb8 does not understand cesu8 , int's "normal" utf8 ( maybe with exception of metadata strings )

Currently all strings are converted to cesu8 and mysql seems to understand that. But I think we need to implement per charset based parsing as you explained in point (1) and (2)

> Do you think we'll add other encodings later? 
> https://dev.mysql.com/doc/refman/5.7/en/charset-connection.html

Yes, I think we may need different encoding for many charset, ParserCache based arch will allow us to reuse multiple encoding, given that `iconv` supports them.

 1) and 2) will need the charset <-> encoding mapping, then we need to pass the correct encoding to `iconv`

3) Not sure
4) yes we need those integration tests

> Overall above might be too big for single pr, I'm happy to break it in two

Sounds good

lib/packets/packet.js
```diff
@@ -348,7 +349,10 @@ Packet.prototype.readLengthCodedString = function () {
     return null;
   }
   this.offset += len;
-  return this.buffer.utf8Slice(this.offset - len, this.offset);
+
+  // becuase MySQL utf8 is actually cesu8
```
sidorares: typo :)

sushantdhiman: oh :)

sidorares:
maybe add StringParser.encode to prepare statement packet as well? Similar to `Query` packet

https://github.com/sidorares/node-mysql2/blob/722958e6d6383cd4b87243e8f15b494a42c0160a/lib/packets/prepare_statement.js#L12

sidorares:
So I was thinking over weekend about possible scenarios with different combinations of incoming and outgoing encodings

1) Connection charset: currently used completely incorrectly. We always send utf8 ( with maybe cesu8 after this pr). From a server point of view it's client saying "all textual data from now on is encoded with this table". Server options are 1) accept it and translate from client table to internal encoding or 2) reject connection - there is no more negotiations after that. I need to double check if connection encoding applies to hello packet itself ( user, database - what if those contain non-latin1 ? are they encoded ? do we encode the values? ) So currently if I connect with 1251 and insert "привет" the data would be broken

2) server encoding - we probably can ignore this. The only value I see in this data: it's indication of charset server definitely support and safe to use in hello packet

3) Results encoding: just use what's set in the column charset and decode into JS string. Does not matter what client connection or server connection was used. Usually does not reflect original table charset, server converts automatically into connection charset

I tested a bit what concerned me about utf8mb4 vs cesu8 and it seems to work fine ( tested writing cesu8 and reading cesu8. I'll do another test now: write normal utf8 4byte char (non cesu8 encded) and read it with client connected as utf8mb4 ) So I think we can merge this PR ( but also happy to continue and finish with actually using result charset and connection charset )

sidorares:
regarding 2) - this is too edge case, should be manual step IMO. If you have really old legacy server that does not speak utf8 and you have non-latin chars in the user name - just configure your client to use that exotic encoding

sidorares:
Also would be interesting to see if there is observable performance hit with iconv introduction. I'll try to run benchmarks of master vs fix-char-encoding

sidorares:
@sushantdhiman I added mapping table

sushantdhiman:
Thanks for that mapping table @sidorares 

Here are my thoughts 

1) Connection encoding, It seems data can lost in some cases like setting connection at `latin1` and trying to insert `восемь`. Not sure we can do anything about it if the user is specifying it themselves

2) Server encoding, if initial hello packet fails to recognize `utf8` / `ceus8` / `other user defined charset` then we should warn users and / or switch to `latin1`. It should be noted that even `utf8mb4` is returned as `45 Not supported`, we should handle that case too (This may be because we have been using `utf8` in past for packets, with packets prepared with `ceus8` server should properly understand all encodings now)

3) Result set encoding, if we have the result set encoding, we should probably use it rather than relying on auto conversion.

sidorares:
1) If user is specifying koi8r for example we need to serialise JS string using koi8r encoding. Currently ces8 is used regardless of client setting

2) I'm pretty sure if client encoding is not in the list server can understand the connection is closed by server

3) yes

sidorares:
maybe I misunderstood your point about 1) . If for example encoding is latin1 and data is `восемь` ( outside of latin1 range ) there is nothing we can do ( maybe display warnings or replace characters with ? - this is probably default iconv behaviour anyway )

sidorares:
I was wrong about 2). When I connect with non-existing charset ( I used 250 as charset number  )  the server does not complain at all. All results sent back with 63 charset ( `BINARY` ) and probably incoming data is treated as binary

sidorares:
@sushantdhiman I added code to use encodings. It's fairly big change, could you look?

sidorares:
looks like I broke everything :)

sushantdhiman:
`charset to encoding` might be giving wrong encoding types ?

sushantdhiman:
` Error: Cannot find module './constants/charset_encodings.js'` :-p

sidorares:
:p one second

sidorares:
getting "index out of range errors". I suspect this is `offset += buf.copy()` used together with MockBuffer - the part where packet first serialized without actual writing to buffer to get required lenhg

sidorares:
I'll get back to this in the morning

sidorares:
phew, that was much more writeString/readString to wire up to correct encoding than I originally expected. I'll add tests now and we should be good to merge this

sidorares:
@sushantdhiman @dougwilson I'm trying various clients and a bit lost with utf8/cesu8 and results/metadata

When I use standard mysql command client, `select "💩"` is sent as `<Buffer 0e 00 00 00 03 73 65 6c 65 63 74 20 22 f0 9f 92 a9 22>` packet over the wire - with utf8 encoded  f0 9f 92 a9. If we treat UTF8MB4 as cesu8, the node client would not be able to decode this char ( `����` is displayed )

Should we assume that  UTF8MB4 is 'utf8 for results and queries, cesu8 for metadata?' ? Would like to make good tests to see what's actually used in existing clients

sushantdhiman:
> Should we assume that UTF8MB4 is 'utf8 for results and queries, cesu8 for metadata?'

That seems to be correct as per our previous discussion and by wireshark results

sidorares:
@sushantdhiman  does not look like metadata is encoded by server as cesu8 in that case - just '?' symbol ( 0x3f )

sidorares:
what I see now: cesu8 works ok (rows and fields) for non-bmp chars if read and write by node client. mysql client fails to read them ( `set names utf8` or `set names utf8mb4` )

utf8mb4 work for data ( written by node client, read by mysql command line clien ) nut not for metadata ( converted to ? server side )

Looks like only practical advise for new data: use UTF8MB4 for connection encoding but never use non-mb4 in field names. Old data: figure out what's there, if you have cesu-8 encoded chars connect with UTF8_GENERAL_CI otherwise connect as UTF8MB4_UNICODE_CI .

sidorares:
@sushantdhiman I think this is ready. Maybe you can have one final look again just to be sure an I'll mere it (or you merge)

sushantdhiman:
:clap: :clap: :tada:

sushantdhiman:
@sidorares `rc-12` is not on `npm` yet, right ?

sidorares:
@sushantdhiman it is now, wanted tag to pass CI before publishing

I think this is last rc before 1.0.0 - this encoding stuff was the main reason I was holding release, hopefully after we'll have more frequent patch/major versions

Issue #380, opened 2016-08-18
`Buffer()` without a `new` keyword is going to be hard-deprecated soon


ChALkeR:
Refs: https://github.com/nodejs/node/pull/8169, https://github.com/nodejs/node/pull/7152.

Two options here:
 * Use the new Buffer API — `Buffer.alloc()`/`Buffer.from()`/`Buffer.allocUnsafe()` (requires [a shim](https://github.com/feross/safe-buffer) for v0.10/v0.12 and older v4.x versions prior to v4.5.0).
 * Use `new Buffer()` for the time being — just add the `new` keyword everywhere. You should manually check that everything is safe in this case, see the links below for more explanation. That _might_ be hard-deprecated at some later point.

More background:
 * https://github.com/ChALkeR/notes/blob/master/Lets-fix-Buffer-API.md
 * https://github.com/ChALkeR/notes/blob/master/Buffer-knows-everything.md
 * https://github.com/nodejs/node/pull/7152#issuecomment-240880317

Quick grep (you should better re-check):
```
mysql2-1.0.0-rc.9.tgz/examples/pass-sha.js:4:  passwordSha1: Buffer('8bb6118f8fd6935ad0876a3be34a717d32708ffd', 'hex')
mysql2-1.0.0-rc.9.tgz/lib/packets/auth_switch_response.js:8:    data = Buffer(data);
mysql2-1.0.0-rc.9.tgz/lib/packets/handshake_response.js:91:    packet.writeNullTerminatedString(Buffer(this.authToken));
mysql2-1.0.0-rc.9.tgz/lib/packets/packet.js:685:  var res = Buffer(0);
mysql2-1.0.0-rc.9.tgz/test/integration/connection/test-binary-charset-string.js:74:  assert.deepEqual(rows, [{'x\'010203\'': Buffer([1, 2, 3])}]);
mysql2-1.0.0-rc.9.tgz/test/integration/connection/test-binary-charset-string.js:78:  assert.deepEqual(rows2, [{'x\'010203\'': Buffer([1, 2, 3])}]);
mysql2-1.0.0-rc.9.tgz/test/integration/connection/test-binary-charset-string.js:83:  assert.deepEqual(rows4, [{'x\'010203\'': Buffer([1, 2, 3])}]);
[...]

PR #381, opened 2016-08-19
replace deprecated Buffer apis with new API

sidorares:
Ref: #380

`Buffer(number)` and new `Buffer(number)` replaced with `Buffer.unsafeAlloc(number)`
`new Buffer(string)` and `new Buffer(array)` replaced with `Buffer.from`

some `new Buffer()` still left in tools and benchmarks


sidorares:
@sushantdhiman would you be able to review?

sushantdhiman:
We are using `allocUnsafe` a lot (I guess for performance), its good but we need to be sure those buffer are completely filled.

sidorares:
yes, everywhere it's used the intention is to fill all bytes of buffer. There is one exception, in handshake packet there is 10 bytes zero filler so I'm using `Buffer.alloc(size, 0)` + `packet.skip(10)` there (instead of filling 10 bytes manually). I'll try to double check again myself. I wander maybe we could use some version of Buffer in tests that allows to check that there is no uninitialised space in resulting buffers.

sushantdhiman:
It appears most of time `unsafeAlloc` is used to build packets or headers that are filled to the end. So it LGTM.

We might be able to add unit tests for each `Header/Packet` type and try initialize them with mock data, at the end we could inspect if `Header/Packet` was properly prepared and fully filled. But that will be a big test suite to add, we should add it in separate PR(s)

sidorares:
Yes agree ( in general, having unit tests for all packets would be good)

PR #383, opened 2016-08-19
Add test for creating functions

lillem4n:
I added this test as it seems this functionality is missing in the implementation or maybe I am doing something wrong?

It would come to great use for me if it was possible to create a function from node-mysql2.

### DIFF
--- a/test/integration/connection/test-create-function.js
+++ b/test/integration/connection/test-create-function.js
@@ -1,0 +1,37 @@
+var common = require('../../common');
+var connection = common.createConnection({multipleStatements: true});
+var assert = require('assert');
+
+var _stmt1;
+var query1 = '';
+var query2 = 'SELECT multi_two(4) AS foo;';
+
+query1 += 'DROP FUNCTION IF EXISTS multi_two;\n';
+query1 += 'delimiter $$\n';
+query1 += 'CREATE FUNCTION multi_two (x INT) RETURNS INT\n';
+query1 += '  DETERMINISTIC\n';
+query1 += '  BEGIN\n';
+query1 += '    return x * 2;\n';
+query1 += '  end\n';
+query1 += '$$\n';
+query1 += 'delimiter ;\n';
+
+connection.query(query1, function (err) {
+  if (err) {
+    throw err;
+  }
+
+  connection.query(query2, function (err, _rows) {
+    if (err) {
+      throw err;
+    }
+
+    _stmt1 = _rows;
+    connection.end();
+  });
+});
+
+process.on('exit', function () {
+  assert(_stmt1 instanceof Array);
+  assert.equal(_stmt1[0].foo, 8);
+});


sidorares:
Hi @lillem4n ! The tests not supposed to cover functionality of mysql server, only possible ways the server can interact via protocol. What's the result of creating function? If it's normal result set then in general this is already covered ( diver sends query and gets rows in response ). Though I'm not against slight duplication in tests, the test itself is good also for documentation/example purpose.

Could you double check the query? Test fails with `Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use` error

lillem4n:
Hey @sidorares 
I think I understand the purpose of the tests (not testing server functionality). This, however have something to do with the connection or similar. The exact same SQL works well when ran through the mysql shell client and adminer (https://www.adminer.org/). 

I know to little about the inner workings of MySQL/MariaDB and the raw connections to it to know if this is very different. :)

I encoutered this issue when I was trying to run database migrations via node-mysql2.

sidorares:
I think there are two problems in your query:
1) You need to enable multiple statements flag or make two separate queries - "DROP FUNCTION IF EXISTS multi_two" and "CREATE FUNCTION ..."
2) (this is probably not critical) No need to change delimiter when you create function from library call, this is to help enter function from command line repl.

I have one example here - https://github.com/sidorares/node-mysql2/blob/efb4af102db5f96ea97e50559904605808b35226/test/integration/connection/test-binary-multiple-results.js#L76-L84

lillem4n:
1) I think I have activated multiple statements?
```javascript
var connection = common.createConnection({multipleStatements: true});
```
2) This might be the critical issue, since the test is choking on the delimiter setting on my end. I'll try a way without changing delimiters now and see if that works. :)

lillem4n:
It seems it actually is the "delimiter" command that breaks. The really strange thing is that this works on the same database server over another connection (I tried mysql CLI client and PHP client)

This is an issue, but might not be very big. I'll try my larger function that caused this issue without the delimiter and see if I get it running.

Thanks a lot for the help! Feel free to abandon this PR and maybe create an issue about the delimiter thingie?

lillem4n:
Sadly the delimiter change is needed since the ; character is used in the function body. Without the delimiter change it seems to be impossible to create functions with more than one statement. :(

sidorares:
this example works fine for me (with and without `multipleStatements` flag set):

```js
connection.query('DROP FUNCTION IF EXISTS multi_two');
connection.query(`
  CREATE FUNCTION multi_two (x INT) RETURNS INT
  DETERMINISTIC
  BEGIN
    set @a=1;
    set @b=2;
    return @a + @a + @b * x;
  END
`);
connection.query('SELECT multi_two(?) AS foo', 20, (err, rows) => console.log(rows[0].foo));
```

lillem4n:
Yes! Very nice!

That works great as well for me. Thank you. :)

sidorares:
@lillem4n I'll close your PR, but if you find time to improve documentation feel free to submit another PR, it seems that what we discussed here is not so trivial and might need to be reflected in docs. Overall this is mysq side of things, not driver so probably links to good article (client agnostic) might be enough

Issue #82, opened 2014-02-18
ORDER BY on VIEW gives unexpected results


wshager:
Steps to reproduce:

* create 2 tables
* create a view1 selecting fields from both tables with a (inner) join
* create a view2 selecting fields from view1
* execute("SELECT * FROM view2 ORDER BY somefield")

PR #395, opened 2016-09-10
added test for #82

sushantdhiman:
Just a proof of concept that #82 was an userland problem, we can keep the test if we want to prevent future regressions


sidorares:
👍

PR #396, opened 2016-09-10
added connect timeout event and tests

sushantdhiman:
Closes https://github.com/sidorares/node-mysql2/issues/376

PR #397, opened 2016-09-11
Save correctly compiled parser for re-use

sidorares:
It looks like text protocol parser cache was not used at all: https://github.com/sidorares/node-mysql2/pull/396#issuecomment-246134184

After fix, few tests failed. The reason was over-optimization in parser: encoding was part of compiled parser and if encoding changed between parser invocation the result was incorrect. This is changed to dynamically use charset from field parameter. Another problem was in re-use parser with typeCast options, first time where it's a function and second time where it's `false`. Now typeCast added to parser cache key, when typeCast values differ new parser is generated

TODO: compare benchmarks before/after caching

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -8,6 +8,7 @@ var Command = require('./command.js');
 var Packets = require('../packets/index.js');
 var compileParser = require('../compile_text_parser.js');
 var ServerStatus = require('../constants/server_status.js');
+var CharsetToEncoding = require('../constants/charset_encodings.js');
 
 var EmptyPacket = new Packets.Packet(0, Buffer.allocUnsafe(4), 0, 4);
 
@@ -16,7 +17,7 @@ function Query (options, callback)
   Command.call(this);
   this.sql = options.sql;
   this.values = options.values;
-  this.options = options;
+  this._queryOptions = options;
   this.onResult = callback;
   this._fieldCount = 0;
   this._rowParser = null;
@@ -35,6 +36,7 @@ Query.prototype.start = function (packet, connection) {
     console.log('        Sending query command: %s', this.sql);
   }
   this._connection = connection;
+  this.options = Object.assign({}, connection.config, this._queryOptions);
   var cmdPacket = new Packets.Query(this.sql, connection.config.charsetNumber);
   connection.writePacket(cmdPacket.toPacket(1));
   return Query.prototype.resultsetHeader;
@@ -168,7 +170,7 @@ Query.prototype.readField = function (packet, connection) {
     this._rowParser = connection.textProtocolParsers[parserKey];
     if (!this._rowParser) {
       this._rowParser = compileParser(fields, this.options, connection.config);
-      connection.textProtocolParsers[parserKey] = this.rowParser;
+      connection.textProtocolParsers[parserKey] = this._rowParser;
     }
     return Query.prototype.fieldsEOF;
   }
@@ -195,7 +197,8 @@ Query.prototype.row = function (packet)
     return this.done();
   }
 
-  var row = new this._rowParser(packet, this._fields[this._resultIndex], this.options);
+  debugger;
+  var row = new this._rowParser(packet, this._fields[this._resultIndex], this.options, CharsetToEncoding);
   if (this.onResult) {
     this._rows[this._resultIndex].push(row);
   } else {
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -197,7 +197,6 @@ Query.prototype.row = function (packet)
     return this.done();
   }
 
-  debugger;
   var row = new this._rowParser(packet, this._fields[this._resultIndex], this.options, CharsetToEncoding);
   if (this.onResult) {
     this._rows[this._resultIndex].push(row);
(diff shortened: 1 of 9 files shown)


sushantdhiman:
> Now typeCast added to parser cache key, when typeCast values differ new parser is generated

Yep, my logs did show that different config is used with typeCast being `false`, glad to know its fixed.

>  compare benchmarks before/after caching

Definitely there will be some improvements :) Do you have that EC2 benchmark suite ready ?

lib/commands/query.js
```diff
@@ -35,6 +36,7 @@ Query.prototype.start = function (packet, connection) {
     console.log('        Sending query command: %s', this.sql);
   }
   this._connection = connection;
+  this.options = Object.assign({}, connection.config, this._queryOptions);
```
sushantdhiman: Is `Object.assign` supported by all node version we test with ?

sidorares:
> Definitely there will be some improvements :) Do you have that EC2 benchmark suite ready ?

Initial setup is a bit manual, I'll try to start it overnight. Also interesting to see iconv impact

lib/commands/query.js
sidorares: probably no, if >0.10 tests pass on CI I'll add shim for 0.10

lib/connection.js
```diff
@@ -442,7 +442,7 @@ Connection.prototype.resume = function resume () {
 
 Connection.prototype.keyFromFields = function keyFromFields (fields, options) {
   var res = (typeof options.nestTables) + '/' + options.nestTables + '/' + options.rowsAsArray
-    + options.supportBigNumbers + '/' + options.bigNumberStrings;
+    + options.supportBigNumbers + '/' + options.bigNumberStrings + '/' + typeof options.typeCast;
```
sushantdhiman: any chances this key will face collision when `typeCast` function is changed on runtime. I think we need to generate a hash from `typeCast` function and use that for key field. What do you think ?

sushantdhiman:
> Definitely there will be some improvements :) Do you have that EC2 benchmark suite ready ?
> > Initial setup is a bit manual, I'll try to start it overnight. Also interesting to see iconv impact

Yeah it will be great if we could release v1.0 with fresh benchmarks against current `node-mysql` version

test/integration/connection/encoding/test-client-encodings.js
```diff
@@ -2,8 +2,9 @@ var mysql = require('../../../../index.js');
 var common = require('../../../common');
 var assert = require('assert');
 
-var connection = common.createConnection({charset: 'UTF8_GENERAL_CI'});
-connection.query('create table if not exists __test_client_encodings (name VARCHAR(200))', function(err) {
+var connection = common.createConnection({charset: 'UTF8MB4_GENERAL_CI'});
+connection.query('drop table __test_client_encodings');
```
sushantdhiman: `drop table if exists`

lib/connection.js
sidorares: function itself is passed as a parameter, here I'm avoiding extra if/else in the parser if function is not callable at all

sushantdhiman: Oh then as long as the correct func is being used we can keep it in same way

sushantdhiman:
LGTM

sidorares:
screen shot 2016-09-14 at 11 33 45 am

@sushantdhiman here are benchmarks
two graphs at the top look too close, I might made a mistake so I have to double check if this is similar performance of mysql and mysql2-rc12 or they both results against same driver. Bottom one is result of starting using cache. "select * from mysql.user" - 8 rows, 47 fields - improvement is very visible. ( Y axis is response time, X - latency histogram ). I'll run again the range 9000rps - 15000

sushantdhiman:
I like those graphs @sidorares :+1:, good improvements.

`mysql` and `mysql2-rc12` are painfully close to each other :stuck_out_tongue_closed_eyes: , I hope 9000-15000 rps tests will use correct driver versions :)

sidorares:
I have a version with a slider where you can see how latency changes as you increase load, I might push it to gh-pages

sidorares:
I pushed rc.13, really hope it to become final release (only diffeence with previous is cache fix)

sidorares:
@sushantdhiman 

http://sidorares.github.io/node-mysql2/benchmarks/results/14092016/chart.html
http://sidorares.github.io/node-mysql2/benchmarks/results/14092016/chart3d.html

chart3d is same as chart where 3d axis doing what slider does on chart

sushantdhiman:
So it turns out `rc12` was slower for some cases (iconv hit), but `master` has better performance.

Please update the graphs with `x` and `y` axis labels so everyone can understand what we are plotting against. We should also update graphs with the system specs and `mysql server` version used for testing.

Now I think we should release `v1` officially :tada:

sidorares:
@sushantdhiman I had some strange errors during benchmark due to undefined encoding, and I'm struggling to reproduce errors now. I'll play a little bit more but if there is a bug we can always do patch 1.0.x

sidorares:
v1 published!

( the issue I was getting fixed here https://github.com/sidorares/node-mysql2/commit/d8e3b05bbeca8b1b87428600071b82e9cbbb6b90 )

Issue #389, opened 2016-08-31
when client updates client charset via query, update in on the client as well


sidorares:
depends on #388 

if character_set_client variable is updated we need to update connection.charsetNumber as well

This is improvement on top of #374 and depends on it

Also need to add inverted version of https://github.com/sidorares/node-mysql2/blob/9fea6c784bf349b26bc5596bc5887f737efd3861/lib/constants/charset_encodings.js lookup table to be able to get charset id from encoding name.

Issue #388, opened 2016-08-31
add support of CLIENT_SESSION_TRACK in ok packet


sidorares:
https://dev.mysql.com/doc/internals/en/packet-OK_Packet.html#cs-sect-packet-ok-sessioninfo

PR #400, opened 2016-09-17
feat: issue#288

plusmancn:
Request Code Review.

### DIFF
--- a/lib/packets/resultset_header.js
+++ b/lib/packets/resultset_header.js
@@ -18,6 +18,15 @@ function ResultSetHeader (packet, bigNumberStrings, encoding)
     this.infileName = packet.readString(undefined, encoding);
   }
 
+  // issue#288: Add changedRow if exists, reference to https://github.com/sidorares/node-mysql2/issues/288
+  if (encoding) {
+    var message = packet.readString(undefined, encoding);
+    var m = message.match(/\schanged:\s*(\d+)/i);
+    if (m !== null) {
+      this.changedRows = parseInt(m[1], 10);
+    }
+  }
+
   // snippet from mysql-native:
   // res.affected_rows = this.lcnum();
   // res.insert_id = this.lcnum();
--- a/test/integration/connection/test-update-changed-rows.js
+++ b/test/integration/connection/test-update-changed-rows.js
@@ -1,0 +1,47 @@
+/**
+ * <plusmancn@gmail.com> created at 2016.09.17 15:24:34
+ *
+ * issue#288: https://github.com/sidorares/node-mysql2/issues/288
+ */
+var common = require('../../common');
+var connection = common.createConnection();
+var assert = require('assert');
+
+var result1 = undefined;
+var result2 = undefined;
+
+// connection.
+connection.query([
+  'CREATE TEMPORARY TABLE `changed_rows` (',
+  '`id` int(11) unsigned NOT NULL AUTO_INCREMENT,',
+  '`value` int(5) NOT NULL,',
+  'PRIMARY KEY (`id`)',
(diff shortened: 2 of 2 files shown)


test/integration/connection/test-update-changed-rows.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * <plusmancn@gmail.com> created at 2016.09.17 15:24:34
```
sushantdhiman: ?

test/integration/connection/test-update-changed-rows.js
```diff
@@ -0,0 +1,47 @@
...
+  ') ENGINE=InnoDB DEFAULT CHARSET=utf8'
+].join('\n'));
+connection.query('insert into changed_rows(value) values(1)');
+connection.query('insert into changed_rows(value) values(1)');
+connection.query('insert into changed_rows(value) values(2)');
+connection.query('insert into changed_rows(value) values(3)');
+
+connection.execute('update changed_rows set value=1', [], function (err, _result) {
+  if (err) {
+    throw err;
```
sushantdhiman: `assert.ifErr` can be used

lib/packets/resultset_header.js
```diff
@@ -18,6 +18,15 @@ function ResultSetHeader (packet, bigNumberStrings, encoding)
     this.infileName = packet.readString(undefined, encoding);
   }
 
+  // issue#288: Add changedRow if exists, reference to https://github.com/sidorares/node-mysql2/issues/288
+  if (encoding) {
+    var message = packet.readString(undefined, encoding);
+    var m = message.match(/\schanged:\s*(\d+)/i);
```
sushantdhiman: not sure if its the best way to do this, need @sidorares comments :)

sidorares:
lgtm, thanks!
I guess no need to update docs as this covers node-mysql behaviour

sidorares:
oops I should have waited for you @sushantdhiman

lib/packets/resultset_header.js
sidorares: I'm going to change a lot here anyway to support #388 and #389 but this is still good start
`if (encoding)` probably unnecessary as it must be defined here

sushantdhiman:
Everything else looks good but extracting changed value like [this](https://github.com/sidorares/node-mysql2/pull/400/files#diff-2e7a1c3222ba746fc6605a26c5c0c79fR24) , is it good? can we read packet serially like we do in other cases ? we are converting entire packet to string and extracting, costly ?

lib/packets/resultset_header.js
plusmancn: @sushantdhiman According to [https://dev.mysql.com/doc/internals/en/packet-OK_Packet.html]，the`EOF` is `human readable status information`。In other words, the `EOF` is not in one good format.

sidorares: I'd argue that `changedRows` should not be part of api ( up to client how to parse message) but since it's already in node-mysql I'm OK to have it as well

sidorares:
no, message is not entire packet but only part from current position until end

lib/packets/resultset_header.js
plusmancn: @sidorares `test/integration/connection/test-load-infile.js` can't pass the test without `if (encoding)`。

sidorares:
@sushantdhiman should we save it and expose as `info`? `query("insert ...", (err, res) => { console.log(res.info) })` ?

sushantdhiman:
@sidorares no I think we should not diverge away from `node-mysql` public api, beside number of result parameters are not that large yet that they need sub sectioning.

Issue #57, opened 2013-10-25
Socket Close Fatal Errors


crh3675:
Getting this message once in a while:

```
Error: This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (net.js:274:12)
    at Connection.write (/project/path/node_modules/mysql2/lib/connection.js:94:15)
    at Connection.writePacket (/project/path/node_modules/mysql2/lib/connection.js:103:10)
    at Query.start (/project/path/node_modules/mysql2/lib/commands/query.js:23:14)
    at Query.Command.execute (/project/path/node_modules/mysql2/lib/commands/command.js:26:20)
    at Connection.handlePacket (/project/path/node_modules/mysql2/lib/connection.js:233:28)
    at Connection.addCommand (/project/path/node_modules/mysql2/lib/connection.js:245:10)
    at Connection.query (/project/path/node_modules/mysql2/lib/connection.js:285:15)
    at setSession (/project/path/lib/session/index.js:87:23)
    at SessionStore.set (/project/path/lib/session/index.js:102:13)
    at Session.save (/project/path/node_modules/express/node_modules/connect/lib/middleware/session/session.js:63:25)
```

Not sure why this is occuring but it completely stops the NodeJS server from running.  How can I catch this error and keep the server alive?

Issue #38, opened 2013-08-28
Connection dying


mscdex:
The tcp connection with 'mysql2' appears to die after a short period of time (just a few seconds or so), which causes problems if you try to then perform a query some time later (namely a "This socket has been ended by the other party" exception from net.js).

I briefly looked at porting ping() from 'mysql' to try and see if occasional pinging (while no other queries were in progress) would help things, but the two modules are not similar enough for me to know what needs to be done.

PR #404, opened 2016-09-17
Don't write to stream when socket is closed

### LINKED ISSUE #289: This socket has been ended by the other party
Hey, I just got an error `This socket has been ended by the other party ` on the version **0.15.8**. I found similar issue https://github.com/sidorares/node-mysql2/issues/57.

Not sure if I can reproduce it again to grab some example. Please let me know if it's needed I'll try.
Thanks.

>This socket has been ended by the other party     at Socket.writeAfterFIN [as write] (net.js:270:12)     at Connection.write (/app/node_modules/mysql2/lib/connection.js:136:15)     at Connection.writePacket (/app/node_modules/mysql2/lib/connection.js:150:10)     at Query.start (/app/node_modules/mysql2/lib/commands/query.js:39:14)     at Query.Command.execute (/app/node_modules/mysql2/lib/commands/command.js:34:20)     at Connection.handlePacket (/app/node_modules/mysql2/lib/connection.js:310:28)     at Connection.addCommand (/app/node_modules/mysql2/lib/connection.js:326:10)     at Connection.query (/app/node_modules/mysql2/lib/connection.js:397:15)     at Connection.wrapped (/app/node_modules/newrelic/lib/transaction/tracer/index.js:155:28)     at Connection.wrappedFunction [as query] (/app/node_modules/newrelic/lib/transaction/tracer/index.js:356:55)     at /app/node_modules/knex/lib/dialects/mysql/index.js:92:18     at tryCatcher (/app/node_modules/knex/node_modules/bluebird/js/main/util.js:26:23)     at Promise._resolveFromResolver (/app/node_modules/knex/node_modules/bluebird/js/main/promise.js:480:31)     at new Promise (/app/node_modules/knex/node_modules/bluebird/js/main/promise.js:70:37)     at Client._query (/app/node_modules/knex/lib/dialects/mysql/index.js:88:12)     at Client.query (/app/node_modules/knex/lib/client.js:136:24)     at Runner.<anonymous> (/app/node_modules/knex/lib/runner.js:115:24)     at Runner.tryCatcher (/app/node_modules/knex/node_modules/bluebird/js/main/util.js:26:23)     at Runner.query (/app/node_modules/knex/node_modules/bluebird/js/main/method.js:15:34)     at /app/node_modules/knex/lib/runner.js:43:21     at /app/node_modules/knex/node_modules/bluebird/js/main/using.js:176:30     at wrapped (/app/node_modules/newrelic/lib/transaction/tracer/index.js:155:28)     at tryCatcher (/app/node_modules/knex/node_modules/bluebird/js/main/util.js:26:23)     at Promise._settlePromiseFromHandler (/app/node_modules/knex/node_modules/bluebird/js/main/promise.js:507:31)     at Promise._settlePromiseAt (/app/node_modules/knex/node_modules/bluebird/js/main/promise.js:581:18)     at Promise._settlePromises (/app/node_modules/knex/node_modules/bluebird/js/main/promise.js:697:14)     at Async._drainQueue (/app/node_modules/knex/node_modules/bluebird/js/main/async.js:123:16)     at Async._drainQueues (/app/node_modules/knex/node_modules/bluebird/js/main/async.js:133:10)

sidorares:
network error handler already closes commands in the queue on network error, hovewer there still seems to be possibility to when code is writing to socket after it was already closed:
 - connection closed in the middle of wrtiting command
 - there was long running multi packet command such as `LOAD INFILE` packets from client to server

Just listening `error` events on socket is not enough, in the scenarios above error is thrown from event loop and it's not possible to handle it

hopefully fixes #289 #57 and #38

solution: 
 - add callback to `socket.write` and treat callback error as network error
 - stop listening `LOAD INFILE` stream in case of network error
 - guard from adding new commands and writing packets if client is in the closed state

### DIFF
--- a/lib/commands/execute.js
+++ b/lib/commands/execute.js
@@ -27,6 +27,7 @@ function Execute (options, callback)
   this._executeOptions = options;
   this._resultIndex = 0;
   this._localStream = null;
+  this._unpipeStream = function () {};
   this._streamFactory = options.infileStreamFactory;
 }
 util.inherits(Execute, Command);
--- a/lib/commands/prepare.js
+++ b/lib/commands/prepare.js
@@ -21,6 +21,17 @@ function Prepare (options, callback)
 util.inherits(Prepare, Command);
 
 Prepare.prototype.start = function (packet, connection) {
+  var Connection = connection.constructor;
+  this.key = Connection.statementKey(this.options);
+
+  var statement = connection._statements.get(this.key);
+  if (statement) {
+    if (this.onResult) {
+      this.onResult(null, statement);
+    }
+    return null;
+  }
+
   var cmdPacket = new Packets.PrepareStatement(this.query, connection.config.charsetNumber);
   connection.writePacket(cmdPacket.toPacket(1));
   return Prepare.prototype.prepareHeader;
@@ -101,19 +112,18 @@ Prepare.prototype.fieldsEOF = function (packet, connection) {
 Prepare.prototype.prepareDone = function (connection)
 {
   var self = this;
+  var statement = new PreparedStatementInfo(
+    self.query,
+    self.id,
+    self.fields,
+    self.parameterDefinitions,
+    connection
+  );
+
+  connection._statements.set(this.key, statement);
+
   if (this.onResult) {
-    process.nextTick(function () {
-      self.onResult(
-        null,
-        new PreparedStatementInfo(
-          self.query,
-          self.id,
-          self.fields,
-          self.parameterDefinitions,
-          connection
-        )
-      );
-    });
+    self.onResult(null, statement);
   }
   return null;
 };
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -146,6 +146,7 @@ Query.prototype._streamLocalInfile = function (connection) {
 
   var onError = function (err) {
     command._localStreamError = err;
+    connection.writePacket(EmptyPacket);
   };
 
   command._unpipeStream = function () {
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -27,6 +27,7 @@ function Query (options, callback)
   this._receivedFieldsCount = 0;
   this._resultIndex = 0;
   this._localStream = null;
+  this._unpipeStream = function () {};
   this._streamFactory = options.infileStreamFactory;
   this._connection = null;
 }
@@ -45,6 +46,7 @@ Query.prototype.start = function (packet, connection) {
 
 Query.prototype.done = function () {
   var self = this;
+  this._unpipeStream();
   if (this.onResult) {
     var rows, fields;
     if (this._resultIndex === 0) {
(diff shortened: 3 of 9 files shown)


sidorares:
WIP - would like to add test for this

sushantdhiman:
@sidorares funny that we both start work on our branches almost at same time

sidorares:
found pretty bad bug: when you add commands added after `conn.end()` were just ignored. `execute()` is special in a way that it's actually 2 commands ( prepare + execute ) and if you do code like this: 

```js
   conn.query(q1);
   conn.execute(q2);
   conn.query(q3);
```
the order of execution is query(q1), prepare(q2), query(q3), execute(q2) - because execute is inserted at the end of the queue. 

Previously if you add `.end()`

```js
   conn.query(q1);
   conn.execute(q2);
   conn.query(q3);
   conn.end()
```

then execute(q2) command was ignored. In general I'd advise not to use code like this (though in short scripts I often rely on sequential order if I enqueue commands). Not sure if this "non-atomic" behaviour should be fixed (we could insert `execute` command into queue immediately after `prepare`)

sushantdhiman:
Do we consider `conn.execute()` atomic op?

Internally as you said `execute = prepare + query`, but that may not be obvious to users and they might consider it `atomic`.

Order of execution in queue should be kept same even if its a mix of `execute` and `query`. We should fix this issue

sushantdhiman:
@sidorares Any way to kill the build at Appveyor, [this one](https://ci.appveyor.com/project/sidorares/node-mysql2/build/804) has been running from 3 hours and blocking next builds

sidorares:
@sushantdhiman killed it ( not sure if there is a way a can give you permission. With travis it mirrors github access rules )

sidorares:
> Order of execution in queue should be kept same even if its a mix of execute and query. We should fix this issue

https://github.com/sidorares/node-mysql2/blob/f37c25e23857a42ff4ad8a4e8e6dd515313c33ad/lib/connection.js#L542-L559

maybe have something like `addCommandAfter` and do `connection.addCommandAfter(prepareCommand, executeCommand);`

sushantdhiman:
@sidorares They have the concept of collaborators and teams ([docs](https://www.appveyor.com/docs/team-setup/#collaborator)), but I think you have to add me as collaborator over there too, it wont pick Github access roles

sushantdhiman:
May be something like `addCommands` which accept `n` commands to be added sequentially. Will that make sense ? If doing that is something that will make code more complex, then we should just add `addCommandAfter` as you suggested.

sidorares:
maybe, I'll think about this. The problem here is that `prepare()` connection function does not always add prepare command - only on first use when it's not in the cache. So sometimes it's 2 commands, sometimes 1

sidorares:
@sushantdhiman now prepare & execute always added together one after another. Cache logic moved to prepare command itself - see https://github.com/sidorares/node-mysql2/pull/404/commits/22b2f6edd575078152cd1bfb47efdebc9fb24274

sushantdhiman:
LGTM

PR #412, opened 2016-09-24
Parser Rewrite with generate-function

sushantdhiman:
Closes https://github.com/sidorares/node-mysql2/issues/167

**Changes**

- Used `generate-function` for creating function
- Now using concatenation over array join for some performance gains


sidorares:
good to merge

sidorares:
would be interesting to see if there is any difference between `return vm.runInThisContext(parserFn.toString());` and `return parserFn.toFunction()` ( the latter is basically [eval](https://github.com/mafintosh/generate-function/blob/af11020b5280972156c46876d9779c0e448fdfcb/index.js#L55))

sushantdhiman:
Yea, I will have this tested in next pr, coming in a few minutes :)

Issue #414, opened 2016-09-26
Error in connecting to a proxy address for a Heroku add-on


mm-wang:
Hello,

I am looking to try to connect to a SQL database via a Heroku app, but the database restricts IP addresses, and due to the nature of Heroku's reliance on an EC2 instance with changing IP addresses, I need to use a proxy. I looked through the comments on [this issue](https://github.com/mysqljs/mysql/issues/725) in mysql, and I used the configuration in one of the last comments. I'm repeating it here, but it was originally posted by @pierrickchabi 2 years ago:

```
var mysql2 = require('mysql2');
var url = require("url");
var SocksConnection = require('socksjs');

var mysql_server_options = {
  host: 'database.ho.st',
  port: 3306
};

var socks_options = {
  host: 'proxy.ho.st',
  port: 1080,
  user: 'proxy_username',
  pass: 'proxy_password'
};

var socksConn = new SocksConnection(mysql_server_options, socks_options);

var mysql_options =  {
  database: 'db_name',
  user: 'db_username',
  password: 'secret',
  stream: socksConn
}

var mysqlConn = mysql2.createConnection(mysql_options);

mysqlConn.query('SELECT 1+1 as test1;', function(err, rows, fields) {
  if (err) throw err;

  console.log('Result: ', rows);

  socksConn.dispose();
  mysqlConn.end();
});
```

However, the error I get is: 
```
TypeError: this.stream.destroy is not a function
at Connection._handleTimeoutError (/app/node_mudules/mysql2/lib/connection.js:164:17)
at tryOnTimeout (timers.js:224:198:5)
at Timer.listOnTimeout (timers.js:198:5)
...
```

I am not sure what the issue is, but it seems as if the connection is timing out. Any thoughts or help you can provide?

sidorares:
Looks like a side effect of #396

Can you try with really big `connectTimeout` value @mm-wang ?

@sushantdhiman - I think we should not use `this.stream.destroy()` - we need only assume DuplexStream interface for `connection.stream` ( maybe have special case for Net.Stream if it helps to cleanup immediately at this point )

mm-wang:
With a big connectTimeout, I get an ECONNREFUSED error like so:

```
2016-09-26T14:51:21.466380+00:00 app[web.1]: ERROR:  { [Error: connect ECONNREFU
SED i.p.add.ress:3306]
2016-09-26T14:51:21.466382+00:00 app[web.1]:   code: 'ECONNREFUSED',
[...]

PR #417, opened 2016-09-30
dont destroy stream, only clear timeout

sushantdhiman:
As discussed in

https://github.com/sidorares/node-mysql2/issues/414#issuecomment-249584274
https://github.com/sidorares/node-mysql2/issues/414#issuecomment-250388825


sidorares:
thanks!

Issue #248, opened 2016-01-08
TypeError: value is out of bounds when Insert 1M rows from .sql file


terrywong2:
```
TypeError: value is out of bounds
    at checkInt (buffer.js:825:11)
    at Buffer.writeUInt16LE (buffer.js:883:5)
    at Packet.writeInt16 (/var/www/public/node_modules/mysql2/lib/packets/packet.js:527:15)
    at Packet.writeInt24 (/var/www/public/node_modules/mysql2/lib/packets/packet.js:523:8)
    at Packet.writeHeader (/var/www/public/node_modules/mysql2/lib/packets/packet.js:593:8)
    at Connection.writePacket (/var/www/public/node_modules/mysql2/lib/connection.js:142:10)
    at Query.start (/var/www/public/node_modules/mysql2/lib/commands/query.js:39:14)
    at Query.Command.execute (/var/www/public/node_modules/mysql2/lib/commands/command.js:34:20)
    at Connection.handlePacket (/var/www/public/node_modules/mysql2/lib/connection.js:310:28)
    at Connection.addCommand (/var/www/public/node_modules/mysql2/lib/connection.js:326:10)
```
any ideas?

terrywong2:
I figured out it is a mysql config problem, set max_allowed_packet the value to 1073741824 to solved this problem.

Thanks!

sidorares:
was it accidentally reopened?

terrywong2:
i just set the `max_allowed_packet = 128M`, node-mysql is works fine, but node-mysql2 still shows this error.

here is my code:
<https://gist.github.com/terrywong2/56fd19a1ee81d50f77e7>

lillem4n:
Related, when inserting (large) binary blobs I'm getting similar issues:

> uncaughtException: value is out of bounds date=Wed Sep 28 2016 10:29:56 GMT+0200 (CEST), 
pid=9977, 
uid=1005, 
gid=1005, 
cwd=/srv/filmcentrum/admin, 
execPath=/usr/bin/nodejs, 
version=v4.4.3, 
argv=[
	/usr/bin/nodejs, 
	/usr/lib/node_modules/pm2/lib/ProcessContainerFork.js
], 
rss=196702208, 
heapTotal=93235296, 
heapUsed=89848920, 
loadavg=[0.0029296875, 0.0146484375, 0.04541015625], 
uptime=13724341, 
trace=[
[...]

PR #419, opened 2016-10-01
Added test to trigger value out of bounds bug

lillem4n:
Referenced to #248 - More commits to come


lib/commands/command.js
```diff
@@ -35,7 +35,14 @@ Command.prototype.execute = function (packet, connection) {
   }
 
   // TODO: don't return anything from execute, it's ugly and error-prone. Listen for 'end' event in connection
-  this.next = this.next(packet, connection);
+  try {
```
sushantdhiman: wrapping it to try catch feels dirty to me, may be it can return error in some other way, a bit clean approach? I think right now we should solve actual cause as suggested in https://github.com/sidorares/node-mysql2/issues/248#issuecomment-250903209

sidorares:
@lillem4n unfortunately try/catch are expensive and often cause v8 to deoptimize functions. As a workaround for your particular case we could add check to packet length and if > 16M propagate error to callback, until big packet support is landed

lillem4n:
As long as the library does not throw errors on async calls, I'm perfectly happy. :)

I'm not good enough with this library to make a good enough solution that returns an error, since I do not fully understand the consequences of fiddling with the inside of this.next().

I might dive deeper, but not today.

Hope this helped with showing a bit on what I need on the "client" side.

lib/commands/command.js
lillem4n: I do not really like try/catches either, but I do not know another safe way to make sure no throwed errors are breaking the application implementing this library.

Maybe accept that bugs like this WILL break the app and that's that?

sidorares:
@lillem4n thanks a lot for your input, I do share the same - async system library under no circumstance can throw error in a way client is unable to catch. I'll try to evaluate other approaches, maybe try/catch is indeed the only safe solution, even at performance cost

sidorares:
@lillem4n updates:
1) main issue should be fixed by #421 , we will try to land it first
2) I'll do benchmarking with try/catch at top level of sync code paths. Based on performance cost we might add (or not) them

I'll keep this PR open as placeholder for "how to avoid crashes that are not possible to handle in user code" discussion, but pr itself is unlikely to be merged as is (your test though went to #421)

sidorares:
try/catch performance research:

while V8 with torbofan does some optimisations for try/catch ( http://v8project.blogspot.com.au/2015/07/v8-45-release.html ) on my simple tests there was significant (up to 10x) penalty for a function with try/catch block, even when try catch is around loop (not inside loop / hot path). Also see  https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax

I did very simple benchmark ( 100k rows from real production data copy - 51 field in a row) and it seems that in practice it's not that bad, ~3% decrease while having 5% std deviation (so it's less than statistical error)

current mysql2:
Milliseconds:  2091 rows per sec: 47824.007651841224  fields per sec: 2439024.3902439023
Average fields per sec over ~30 runs, standard deviation: 2464686,  113277

mysql2 + try/catch around packetParser.execute() only
Milliseconds:  1967 rows per sec: 50838.84087442806  fields per sec: 2592780.8845958314
Average fields per sec over ~30 runs,  standard deviation: 2545764 102012

mysql:
Milliseconds:  7957 rows per sec: 12567.550584391101  fields per sec: 640945.0798039462
Average fields per sec over ~30 runs,  standard deviation: 667522 191383

mysql is 3.8 times slower at 100k rows, crashing with memory error at 150k ( mysql2 doing fine with full results, 449800 rows)

obviously this is some extreme test where mysql2 is doing especially well ( very large result set ). When you query 1-5 rows at a time perf difference is not so huge

lillem4n:
@sidorares Thats great! Does that mean try/catch is actually an option after all?

sidorares:
> Does that mean try/catch is actually an option after all?

probably yes. Note that wrapping `packetParser.execute()` should be enough as it does result callback synchronously (which in turn makes all protocol parsing code up to query/execute callback). Exception in packets or commands (as in `/lib/packets/*`, `/lib/commands` ) would be caught by this try/catch (unless compression is used, because compression leaves execution flow to event loop [here](https://github.com/sidorares/node-mysql2/blob/f88eb6e8aa33e9f9703396f0de6ee06c9b4d5a4b/lib/compressed_protocol.js#L39-L43)

PR #420, opened 2016-10-01
Using Denque and Connection Timeout Fix

sushantdhiman:
- `denque` is [much faster](https://github.com/Salakar/denque#benchmarks) than `double-ended-queue` and both are API compatible
- Fixed the `connectTimeout` issue, we need to destroy stream, because its blocking the event reporting. Any error set before destroying are not reported to receiver. We don't need to use timers as well, because we are now implementing it only if stream supports `setTimeout`

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1,9 +1,9 @@
 var Net = require('net');
 var util = require('util');
-var Tls = require('tls');
+var TLS = require('tls');
 var Timers = require('timers');
 var EventEmitter = require('events').EventEmitter;
-var Queue = require('double-ended-queue');
+var Queue = require('denque');
 var SqlString = require('sqlstring');
 var LRU = require('lru-cache');
 
@@ -227,7 +227,7 @@ Connection.prototype.writePacket = function (packet) {
   this.write(packet.buffer);
 };
 
-if (Tls.TLSSocket) {
+if (TLS.TLSSocket) {
   // 0.11+ environment
   Connection.prototype.startTLS = function _startTLS (onSecure) {
     if (this.config.debug) {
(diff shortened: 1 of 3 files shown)


sushantdhiman:
Hmm it seems `denque` use `class`, so we have to wait till Node v4, or may be I could add PR to get tid of Class and other ES6 stuff

Let me know what do you think about timeout `destroy`, I can add different PR to resolve the issue

cc @sidorares

sidorares:
> because its blocking the event reporting. Any error set before destroying are not reported to receiver. 

can you give simple example illustrating this? Not sure I follow

> we are now implementing it only if stream supports setTimeout

so if custom stream is used timeout is silently ignored? This is a bit worrying as 'custom stream' users are probably the ones caring most about what to do if server does not respond for a long time ( likely not a local mysql server: heroky dyno + socks proxy addon to allow whitelist ip, node-ssh2 based stream etc )

sushantdhiman:
`Net.connect` is blocking propagation of events to connection listener, or may be its blocking the `process.exit` (need to check that). So connect timeout test currently run for about 3 minutes. This can be solved by destroying socket.

> so if custom stream is used timeout is silently ignored? This is a bit worrying as 'custom stream' users are probably the ones caring most about what to do if server does not respond for a long time ( likely not a local mysql server: heroky dyno + socks proxy addon to allow whitelist ip, node-ssh2 based stream etc )

Right, ignoring them silently is not something I would like as well. We can

- Use timers and try to call `destroy` if it exists, thus implementing it for both cases (custom and Net)
- Warn users if `destroy` is not found on their custom stream implementation

What do you think ?

PR #421, opened 2016-10-03
Add support for reading and writing packets over 16m

sidorares:
see related #248 and #419

Integration test cherry-picked from #419 

I think main implementation is complete (please review) but would like to add more unit tests for packet parser


sidorares:
still trying to get packet logic working correctly under various normal packet / big packet / compressed packet / compressed packet without compression combination. Need to read `net->pkt_nr` and `net->compress_pkt_nr`logic carefully -  https://github.com/mysql/mysql-server/blob/c28e258157f39f25e044bb72e8bae1ff00989a3d/sql/net_serv.cc#L788-L791

sidorares:
so this meant to be "add big packets support" pr, but ended up in near rewrite of compression protocol and the way sequence ids are handled

Compression protocol:
instead of doing 1:1 packet -> compress & wrap into compressed packet -> write by replacing `writePacket` now a lower level `connection.write()` is replaced with version that handles compression, this way "big" multi-packets are automatically accounted for. Also previously compression was actually never used due to a bug in compressedSize <> originalSize comparison.

Packet now stores `sequenceId` from first packet in sequence and number of multi-packets if size > 2^24. 

`sequenceId` reset moved to `Command::start`, that way commands where start is the only state ( "close statement" as example - no response, just one single one way packet ) reset sequenceId correctly. sequenceId increment after incoming packet now handles overflow correctly ( also increments moved to `connection._bumpSequenceId(n)` and `connection._bumpCompressedSequenceId(n)` helpers )

@sushantdhiman @lillem4n could you look at the changes again and leave feedback? This part proved to be very hard to debug and understand, I'm keen to add more comments and documentation where required

lillem4n:
Thanks for all the great work! I'm a bit held up with a lot of stuff atm, I'll look in to the changes as soon as I can, but expect a couple of days.

sidorares:
@lillem4n merging this, but when you have time feel free to comment on merged commits if you feel anything can be improved

lillem4n:
I've looked through the merge and I see nothing immediate. Must've taken quite a few hours, thanks a lot!

I do not have a solution for this, but, it would be usable if the lib threw an error when the package size grew over what the server could handle. As I understand it the server have a max_allowed_packet setting that can only be changed with a server restart, so it would be awesome if the client lib respected this limit and called back with an error if the size is to large.

sidorares:
there is no need for restart, `'SET GLOBAL max_allowed_packet=[size]` is enough. We could read it at connection time and use value somehow but I'd prefer this to be task for higher level wrappers as there might be some edge cases difficult to handle (other connection change variable, etc - though that can be handled similarly to 744e84eb34797728f34c5c84dd7e7e91c91f1195)

lillem4n:
Agreed :)

PR #425, opened 2016-10-04
Fixed co example

odino:
;-)

### DIFF
--- a/examples/promise-co-await/co.js
+++ b/examples/promise-co-await/co.js
@@ -1,4 +1,4 @@
-var mysql = require('../../promise.js');
+var mysql = require('mysql/promise');
 var co = require('co')
 
 co(function * () {
--- a/examples/promise-co-await/co.js
+++ b/examples/promise-co-await/co.js
@@ -1,8 +1,8 @@
-var mysql = require('../promise.js');
+var mysql = require('../../promise.js');
 var co = require('co')
 
 co(function * () {
-  var c = yield mysql.createConnection({ port: 3307, user: 'root', namedPlaceholders: true });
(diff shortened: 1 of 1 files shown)


examples/promise-co-await/co.js
```diff
@@ -1,8 +1,8 @@
-var mysql = require('../promise.js');
+var mysql = require('../../promise.js');
```
sushantdhiman: Actually change it to `require('mysql/promise')`

odino: then it should be changed [everywhere](https://github.com/sidorares/node-mysql2/blob/master/examples/mysqlproxy.js)?

sushantdhiman: Right, you can fix this example, I'll handle the rest :)

odino: though one thing is that the examples wont work out of the box -- ie I just cloned the repo and ran node examples/promise-co-await/co.js -- works like a charm. If we use absolute requires you wont be able to do that. Happy to change stuff as long as you think that's alright!

sushantdhiman: I think these examples are meant to be copied when user have installed the `node-mysql2` package. So please go ahead and commit the change as discussed

sushantdhiman:
@sidorares What do you think about making all `examples` standalone, right now they cant be copied and tested. I think most developers would like to quickly copy the code and do `npm install mysql2`, we should make all of these examples standalone

sidorares:
@sushantdhiman sorry forgot about this
shouldn't examples `require('mysql2/promise')` instead of mysql/promise ?

sushantdhiman:
I realized that and fixed all case I found in #429

sidorares:
yeah, saw that after commented, 👍

PR #430, opened 2016-10-09
Fix and tests for garbled column names

### LINKED ISSUE #422: Chinese column names become garbled in mysql2^1.0.0
We have some pretty old tables with some Chinese column names like this:
````sql
CREATE TABLE `商城` (
 `商品类型` varchar(255) NOT NULL,
 `商品说明` varchar(255) NOT NULL,
 `价格` int(11) NOT NULL,
 `剩余` int(11) NOT NULL,
 PRIMARY KEY (`商品类型`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8
````

We used to be using **mysql2 1.0.0-rc.11** and everything is OK:
qq 20161003224851

Today I was migrating the app to another server and npm install'd **version 1.1.0** and everything is broken:
qq 20161003225318

I also tested version 1.0.0 and no luck, the column names are still garbled, and I have to roll back to 1.0.0-rc.11

Some details:
Only varchar columns was good, name of other columns (like the int ones) all become garbled.
The charset of my database, table and varchar columns are all utf8.
The js file being executed is also utf8.
The charset of the connection is also utf8. ("set names utf8")

### PR DESCRIPTION
**Pull request by sushantdhiman**
Closes #422

When encoding is unknown we try to parse name as `binary`. I have added check which will force name parsing to `utf8` in such case. I don't know if its a proper fix. Test case included

### DIFF
--- a/test/integration/regressions/test-#442.js
+++ b/test/integration/regressions/test-#442.js
@@ -17,7 +17,7 @@ connection.query([
   ' `' + testFields[1] + '` varchar(255) NOT NULL,',
   ' `' + testFields[2] + '` int(11) NOT NULL,',
   ' `' + testFields[3] + '` int(11) NOT NULL,',
-  ' PRIMARY KEY (`商品类型`)',
+  ' PRIMARY KEY (`' + testFields[0] + '`)',
   ') ENGINE=InnoDB DEFAULT CHARSET=utf8'
 ].join(' '), function (err) {
   assert.ifError(err);
--- a/test/integration/regressions/test-#442.js
+++ b/test/integration/regressions/test-#442.js
@@ -1,0 +1,49 @@
+var common = require('../../common');
+var connection = common.createConnection();
+var assert = require('assert');
+
+var tableName = '商城';
+var testFields = ['商品类型', '商品说明', '价格', '剩余'];
+var testRows = [
+  ['商类型', '商品型', 47, 7],
+  ['类商型', '商型品', 11, 108]
+];
+
+var actualRows = null;
+
+connection.query([
+  'CREATE TEMPORARY TABLE `' + tableName + '` (',
+  ' `' + testFields[0] + '` varchar(255) NOT NULL,',
+  ' `' + testFields[1] + '` varchar(255) NOT NULL,',
+  ' `' + testFields[2] + '` int(11) NOT NULL,',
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/integration/regressions/test-#442.js [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+
+var actualRows = null;
+
+connection.query([
+  'CREATE TEMPORARY TABLE `' + tableName + '` (',
+  ' `' + testFields[0] + '` varchar(255) NOT NULL,',
+  ' `' + testFields[1] + '` varchar(255) NOT NULL,',
+  ' `' + testFields[2] + '` int(11) NOT NULL,',
+  ' `' + testFields[3] + '` int(11) NOT NULL,',
+  ' PRIMARY KEY (`商品类型`)',
```
sushantdhiman: oops

sushantdhiman:
@sidorares this can be merged, may be released if you find the fix acceptable.

sidorares:
the results are definitely better, trying to understand why encoding is binary (63) in the first place

also added similar "treat binary as utf8" for catalog / schema / table / orgTable / orgName

sidorares:
catalog / schema / table / orgTable / orgName:

```
 { catalog: 'def',
    schema: 'test',
    name: '剩余',
    orgName: 'å©ä½',
    table: 'åå',
    orgTable: 'åå',
    characterSet: 63,
    columnLength: 11,
    columnType: 3,
    flags: 4097,
    decimals: 0 }
``` 
vs 

```
{ catalog: 'def',
    schema: 'test',
    name: '剩余',
    orgName: '剩余',
    table: '商城',
    orgTable: '商城',
    characterSet: 63,
    columnLength: 11,
    columnType: 3,
    flags: 4097,
    decimals: 0 }
```

sidorares:
interesting that charset is 'utf8' for text fields but binary for binary for numeric. Does that mean that charset is for content only and not for field itself?

sushantdhiman:
> interesting that charset is 'utf8' for text fields but binary for numeric. Does that mean that charset is for content only and not for field itself?

honestly I dont know, I was surprised by code 63 as well, I think fields should be parsed with `utf8` in all cases and content should be parsed by encoding type.

sidorares:
I think we merge this now (it's for metadata only - if you want column content to be binary it'll still work fine) but I'll attempt to do more research

PR #444, opened 2016-10-22
Using Denque

sushantdhiman:
There is good performance increase as per the benchmarks

**With Denque**

```
packet_parser.execute() toSpeed() 17463.53125 24954.22432172604
           = 44900000000 ±64200000000 packets/second
read 43 column definitions (select * from mysql.user) x 10000: 342000000  ±15000000
           = 1260000 ±60000 columns/second
read 5 chars strings from 10000 bytes buffer x 10000: 1664000000  ±8300000
           = 300500000 ±1500000 chars/second
```

**Master**

```
packet_parser.execute() toSpeed() 23182.71875 25260.35895211405
           = 33900000000 ±36900000000 packets/second
read 43 column definitions (select * from mysql.user) x 10000: 374600000  ±4300000
           = 1150000 ±10000 columns/second
Slower then prev result: 1258983.2138447813 ±columns/second %s
read 5 chars strings from 10000 bytes buffer x 10000: 1989000000  ±44000000
           = 251400000 ±5500000 chars/second
Slower then prev result: 300484564.14630795 ±chars/second %s
```

### DIFF
--- a/benchmarks/unit/packets/column_definition.js
+++ b/benchmarks/unit/packets/column_definition.js
@@ -6,10 +6,10 @@ var fixtureFile = __dirname + '/../fixtures/column_definition';
 function prepareFixture() {
   var Packets = require('../../../lib/packets/index.js');
   var packetInd = 0;
-  Packets.ColumnDefinition = function(packet) {
+  Packets.ColumnDefinition = function(packet, clientEncoding) {
     fs.writeFileSync(fixtureFile + packetInd, packet.buffer.slice(packet.start, packet.end));
     packetInd++;
-    var c = new ColumnDefinition(packet);
+    var c = new ColumnDefinition(packet, clientEncoding);
     //console.log('packet', c);
     return c;
   };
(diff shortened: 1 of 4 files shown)


sidorares:
@sushantdhiman I think at the moment benchmark only measures just some packet deserealisation speed sequentially, queue is not used at all. `Queue` (or `Connection`) are not used. Need to improve benchmarks

sushantdhiman:
@sidorares That can be the case but `denque` is much faster than `double-ended-queue`. You can check benchmarks https://github.com/Salakar/denque#benchmarks , it will definitely help with performance

sushantdhiman:
@sidorares Can we merge this, we can add more benchmarks in different PR, what do you suggest?

sidorares:
looks like Denque is about 20% faster compared to double-ended-queue, should not hurt. ( queue operations performed on add / process command and get/release connection so is of order at most few thousands per sec, overall improvement is likely below 1% )  

```
> denque@1.1.0 benchmark_two_million /private/tmp/denque
> node benchmark/two_million

Platform info:
Darwin 15.5.0 x64
Node.JS 7.0.0
V8 5.4.500.36
Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz × 8

denque x 24,089,339 ops/sec ±0.77% (89 runs sampled)
double-ended-queue x 19,073,767 ops/sec ±0.82% (90 runs sampled)
```

PR #451, opened 2016-11-02
feat: remove use of domains (fixes #449)

### LINKED ISSUE #449: Remove use of domains
Given that [domains have been deprecated](https://nodejs.org/api/domain.html), domains should be removed now.

This would predominantly reverting changes introduced in https://github.com/sidorares/node-mysql2/issues/73.

gajus:
Domains have been deprecated and are pending to be removed.

https://nodejs.org/api/domain.html


sidorares:
windows build became very flaky :( need to fix that

Issue #450, opened 2016-11-02
relying on connection.end() callback is causing process to exit early


gajus:
Here is the code:

```diff
import mysql from 'mysql2';
import Promise from 'bluebird';

const config = {"database":"showtime","host":"104.155.65.164","password":"your-database","port":"3306","user":"root"};

+ const makeItWork = true;

const createConnection = (options = {}) => {
  options.namedPlaceholders = true;
  options.decimalNumbers = false;

  return new Promise((resolve, reject) => {
    const connection = mysql.createConnection(options);

    connection.once('connect', () => {
      resolve(connection);
    });

    connection.once('error', (error) => {
      reject(error);
    });
  });
};

const createDisposableConnection = (connectionOptions) => {
  return createConnection(connectionOptions)
    .then((connection) => {
      return connection;
    })
    .disposer((connection) => {
      console.log('Disposing of connection...');
      return Promise
        .fromCallback((callback) => {
+          if (makeItWork) {
+            connection.end();
+
+            callback();
+          } else {
+            connection.end(callback);
+          }
        })
        .then(() => {
          console.log('This is never reached. (createDisposableConnection)');
        });
    });
};

Promise
  .using(createDisposableConnection(config), (connection) => {
    return new Promise((resolve, reject) => {
      connection
        .query('SELECT `id`, `nid`, ST_X(`coordinates`) `longitude`, ST_Y(`coordinates`) `latitude` FROM `location` WHERE `coordinates` != POINT(0, 0) ORDER BY `google_places_lookup_at` ASC')
        .on('result', (row) => {
          console.log('Hello');

          connection.pause();

          reject(new Error('...'));
        })
    })
    .catch(() => {
      console.log('I am caught!');

      throw new Error('Continue... 0');
    });
  })
  .then(() => {
    console.log('This is never called. (this)');
  })
  .catch(() => {
    console.log('This is never called. (catch)');
  });
```

Here is the program output when `makeItWork = true`:

```
Hello
I am caught!
Disposing of connection...
This is never reached. (createDisposableConnection)
This is never called. (catch)
```

Here is the program output when `makeItWork = false`:

```
Hello
I am caught!
Disposing of connection...
```

gajus:
I have tried debugging this using `--inspect`. However, for whatever reason the debugger is crashing.

I am wondering if this is related to the use of domains.

gajus:
[...]

PR #456, opened 2016-11-02
feat: make end resume command queue (partial #450)

gajus:
Enable `end` method to finish the command queue.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -803,6 +803,10 @@ Connection.prototype.serverHandshake = function serverHandshake (args) {
 Connection.prototype.end = function (callback) {
   var connection = this;
 
+  if (this._paused === true) {
+    this.resume();
+  }
+
   if (this.config.isServer) {
     connection._closing = true;
     var quitCmd = new EventEmitter();


dougwilson:
Would this have unintended consequences? What is `.end()` just destroyed the connection if it was paused? If you want to go with the auto resume, should it not auto resume on _any_ new packet added to the queue, not just end? i.e. if you did change user on a connection that was paused, you'd end up in the same situation as your original issue.

gajus:
@dougwilson I have added more information about the issue https://github.com/sidorares/node-mysql2/issues/450#issuecomment-257899247.

The only reason I am proposing to "resume" command queue on "end" is because:

> [Closing the connection is done using end() which makes sure all remaining queries are executed before sending a quit packet to the mysql server.](https://github.com/mysqljs/mysql/tree/5cc6c6801ba91c5f15114ccf0fd55efd5317febb#introduction)

Unless I misunderstood the code, at the moment, if the command queue is paused and you call "end" it will terminate the command queue at once.

sidorares:
IMO adding command (any, not just `end`) should not resume stream automatically.

What might help is after each callback the paused state is returned back to what it was before (even if exception was thrown from callback function)

gajus:
> IMO adding command (any, not just end) should not resume stream automatically.

Whats the distinction then between `end` and `destroy`? My understanding is that one is supposed to complete the outstanding command queue and the other one must abort as soon as possible.

I might be wrong though. My familiarity with the code base and with the streams is limited.

sidorares:
yes, `.end()` adds `[COM_QUIT](https://dev.mysql.com/doc/internals/en/com-quit.html)` command to the end of the queue, and connection is closed by server as a result when command is executed (commands in single connection are executed sequentially by server). `destroy()` closes client connection immediately and if there are connections in the queue left their completion callbacks called with error

gajus:
Therefore, doesn't it make sense that `end` resumes the command queue?

Alternatively, `end` could throw an error if command queue is paused when `end` is called. I think I prefer this approach.

sidorares:
if stream is paused, you never get to the point where `end()` command starts executing

> Alternatively, end could throw an error if command queue is paused when end is called.

I think it should not matter if stream is paused or not when you put commands to the queue. There is `end()`  method that adds command to the queue and there is `COM_QUIT` command itself

gajus:
I am closing this PR since it looks like I am overlooking something. Please re-open it if you'd like to discuss this further.

Issue #458, opened 2016-11-05
Huge memory use when loading huge data (compared to the old version)


NijiharaTsubasa:
I'm loading some static text from database into memory in my app booting process:
select * from some_huge_table_with_600_records_with_about_25_MB_of_text;

Sorry I can't provide the table content. It's a table with two columns, one VARCHAR(255) as primary key, and one MEDIUMTEXT.

Using mysql2@0.15.8: about 20MB+ of memory use
After updated to mysql2@^1.0.0: **740MB** of memory use (Out of memory in one of my user's sever)

sushantdhiman:
Try against `master` and report if you see same high memory usage

NijiharaTsubasa:
Tested:
1.1.1
1.0.0
master
All have high memory usage

sidorares:
@NijiharaTsubasa would you be able to use git bisect to find exact commit?

@sushantdhiman iconv maybe?

NijiharaTsubasa:
709719426d74f8da09a553dd3ff47ce9923b1562 is the first bad commit
commit 709719426d74f8da09a553dd3ff47ce9923b1562
Author: Sushant <sushantdhiman@outlook.com>
Date:   Sat Aug 27 16:52:07 2016 +0530

    added Iconv-lite based parser

:040000 040000 d0682134bcd4268fbf57d92a4291a5473d4ebcfd f388d40ccc42f24f2a392448186692277a3dcff7 M  lib

dougwilson:
Does removing the `ParserCache` stuff from that commit make it any better? I'm not sure you're supposed to be reusing the parsers after calling `.end()` like the string decoder is doing, especially because `.end()` does not reset any state in the decoder.

sidorares:
thanks @NijiharaTsubasa !

can you try to put `return new StringParser(encoding, options);` as first line in `prepareParser` so that ParserCache is not used?  https://github.com/sidorares/node-mysql2/commit/709719426d74f8da09a553dd3ff47ce9923b1562#diff-43063f93e383632dfd77260c2666c141R18

sushantdhiman:
@sidorares It seems iconv may be the issue and @dougwilson might be right. @NijiharaTsubasa if you can confirm https://github.com/sidorares/node-mysql2/issues/458#issuecomment-258658276 works then I can quickly throw in a PR with this fix

sushantdhiman:
[...]

PR #459, opened 2016-11-06
Removing parser cache

sushantdhiman:
Probably Fixes https://github.com/sidorares/node-mysql2/issues/458

### DIFF
--- a/lib/parsers/string.js
+++ b/lib/parsers/string.js
@@ -1,6 +1,21 @@
 var Iconv = require('iconv-lite');
 
+var NODE_ENCODING = [
+  'ascii',
+  'utf8',
+  'utf16le',
+  'ucs2',
+  'base64',
+  'latin1',
+  'binary',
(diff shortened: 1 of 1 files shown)


sidorares:
one thing I noticed: because iconv-lite tries to be compatible with browser bundlers it does not actually lazy load encodings you use - on first usage all tables are loaded into memory

maybe we should at least short-circuit built-in encodings case: if encoding is "utf8" or "binary" use `Buffer::toString(encoding)` instead of iconv. That way encoding tables only loaded for non-utf case. Or, alternatively use lower level iconv functions and require one of encodings from https://github.com/ashtuchkin/iconv-lite/blob/6043a4709bced434132dd392a25b15b672230334/encodings/index.js

sushantdhiman:
@sidorares AppVeyor build fails on `4.6` for unknown reasons, but I think it can be merged

PR #480, opened 2016-12-15
Fix for `TypeError: Cannot read property 'toString' of undefined` when missing a named parameter

ChiperSoft:
This addresses an exception that occurs when using namedParameters if the named param is missing from the values object. The code hits `this.parameters[i].toString()` and throws up because `this.parameters[i]` is undefined.

Sorry I don't have a test for it, I fixed it directly in my installed dependency. Prior to the change I couldn't even tell what query was causing the error. After the change I get a proper ER_WRONG_ARGUMENTS error out of the query and was able to trace it back to the offending code.

### DIFF
--- a/lib/packets/execute.js
+++ b/lib/packets/execute.js
@@ -32,7 +32,7 @@ Execute.prototype.toPacket = function ()
     length += 2 * this.parameters.length;  // type byte for each parameter if new-params-bound-flag is set
     for (i = 0; i < this.parameters.length; i++)
     {
-      if (this.parameters[i] !== null) {
+      if (this.parameters[i] !== null && this.parameters[i] !== undefined) {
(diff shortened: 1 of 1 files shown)


sidorares:
thanks! I'll need to fix travis setup first ( I think docker-based mysql started to conflict with default ) before merging this, hope to do that over weekend

vlasky:
Hi @sidorades, I'd be grateful if you could merge this fix. I and my colleagues often struggle to identify such problem queries.

sushantdhiman:
@vlasky can you resubmit this PR, its still using old travis setup which fails.

sushantdhiman:
Need to consider discussion https://github.com/sidorares/node-mysql2/issues/493 as well when working on fix

swftvsn:
Could we atleast what I propose in #687? This would allow to turn the logging on and see where the mismatch is.

PR #496, opened 2017-01-18
fix #495 : PromiseConnection missing interface functions from Connection

### LINKED ISSUE #495: promise wrapper `PromiseConnection` is missing `escape`
The `Promise` based wrapper on `Connection` does not fully implement the interface of `Connection`. eg: it lacks `escape` and `escapeId`.

So the provided connection object can not be a drop-in replacement.
I would expect the following to print the same for normal and for `Promise` based `Connection`:

```

const mysql = require("mysql2");
const mysqlPromise = require("mysql2/promise");

var connection1 = mysql.createConnection({ host: "localhost", user: 'root'});
console.log("1: " + connection1.escape("encoded 'test' data"));

mysqlPromise.createConnection({host: "localhost", user: 'root'})
    .then(function (connection2) {
        console.log("2: " + connection2.escape("encoded 'test' data"));
    })
    .catch(function (err) {
        console.log("FAILED 2: " + err.message);
    })
;

```
Unfortunately the output is:
```
1: 'encoded \'test\' data'
FAILED 2: connection2.escape is not a function
```

nros:
`PromiseConnection` was missing some functions `Connection` implements.
Hence it was unable to use the `Promise` wrapped connection like
the "real" connection.

This commit patches `PromiseConnection` and creates a facade
function for all functions that are missing from `PromiseConnection`
prototype.

fixes #495

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -68,25 +68,108 @@ PromiseConnection.prototype.end = function () {
   });
 };
 
+PromiseConnection.prototype.beginTransaction = function () {
+  var c = this.connection;
+  return new this.Promise(function (resolve, reject) {
+    var done = makeDoneCb(resolve, reject);
+    c.beginTransaction(done);
+  });
+};
+
+PromiseConnection.prototype.commit = function () {
+  var c = this.connection;
+  return new this.Promise(function (resolve, reject) {
+    var done = makeDoneCb(resolve, reject);
+    c.commit(done);
+  });
+};
+
+PromiseConnection.prototype.rollback = function () {
+  var c = this.connection;
+  return new this.Promise(function (resolve, reject) {
+    var done = makeDoneCb(resolve, reject);
+    c.rollback(done);
+  });
+};
+
+PromiseConnection.prototype.ping = function () {
+  var c = this.connection;
+  return new this.Promise(function (resolve, reject) {
+    c.ping(resolve);
+  });
+};
+
+PromiseConnection.prototype.connect = function () {
+  var c = this.connection;
+  return new this.Promise(function (resolve, reject) {
+    c.connect(function (error, param) {
+      if (error) {
+        reject(error);
+      } else {
+        resolve(param);
+      }
+    });
+  });
+};
+
+PromiseConnection.prototype.prepare = function () {
+  var c = this.connection;
+  return new this.Promise(function (resolve, reject) {
+    c.prepare(function (error, statement) {
+      if (error) {
+        reject(error);
+      } else {
+        resolve(statement);
+      }
+    });
+  });
+};
+
+// note: the callback of "changeUser" is not called on success
+// hence there is no possibility to call "resolve"
+
 // patching PromiseConnection
 // create facade functions for prototype functions on "Connection" that are not yet
 // implemented with PromiseConnection
-for (var func in core.Connection.prototype) {
 
-  // omit private functions starting with "_"
-  if (
-    func[0] !== '_'
-    && typeof core.Connection.prototype[func] === 'function'
-    && PromiseConnection.prototype[func] === undefined
-  ) {
-    PromiseConnection.prototype[func] = (function factory (funcName, connection) {
-      return function () {
-        return core.Connection
-          .prototype[funcName].apply(this.connection, arguments);
-      };
-    }(func));
+// proxy synchronous functions only
+(function (functionsToWrap) {
+
+  for (var i = 0; functionsToWrap && i < functionsToWrap.length; i++) {
+    var func = functionsToWrap[i];
+
+    if (
+      typeof core.Connection.prototype[func] === 'function'
+      && PromiseConnection.prototype[func] === undefined
+    ) {
+      PromiseConnection.prototype[func] = (function factory (funcName) {
+        return function () {
+          return core.Connection
+            .prototype[funcName].apply(this.connection, arguments);
+        };
+      })(func);
+    }
   }
-}
+
+})([
+// synchronous functions
+  'addCommand',
+  'close',
+  'createBinlogStream',
+  'destroy',
+  'escape',
+  'escapeId',
+  'format',
+  'handlePacket',
+  'keyFromFields',
+  'pause',
+  'pipe',
+  'protocolError',
+  'resume',
+  'serverHandshake',
+  'unprepare',
+  'write'
+]);
 
 
 function createPool (opts) {
--- a/promise.js
+++ b/promise.js
@@ -153,22 +153,16 @@ PromiseConnection.prototype.prepare = function () {
 
 })([
 // synchronous functions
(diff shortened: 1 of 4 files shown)


sidorares:
this is only valid for synchronous functions like `escape` and `escapeId`. Maybe just white list them instead of iterating all function properties of coreConnection (filtered by '_' prefix) ?

Async functions ( such as `beginTransaction` as requested by #497 ) have to return promise and resolve it when callback is invoked )

nros:
You may be right! I am investigating this. Thanks for drawing my attention to this point.

It depends whether asynchronous functions are really a special type of function or just return a promise. I admin, I really do not know, but I am currently checking that out. Hopefully async functions just return a Promise and are not specially typed, marked or handled otherwise. Then it seems sufficient to just pass the call as it is done now, because the return value is just passed to the caller unchanged.

On the other hand, if async functions are a special type of functions which need special marks, hints, etc., there maybe a possibility to get some meta information about the function to be called on `this.connection`. 

I simply love the idea to use a "catch all" ... :)

nros:
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncFunction) has some explanation. functions created like this have `AsyncFunction` as constructor.
```
 async function asyncFunc() {}
```
At the moment support for `async` keyword is limited. Hence asynchronous functions are currently just "emulated" as functions returning a `Promise`. So my patch will work.

Testing with `node` 7.4 (that support `async`/`await` with command line parameter `--harmony`, revealed, that my generator code works seamlessly with ES7. It does not matter whether the target function has been created with `async` keyword or not.

**Therefore I think it is possible to ignore `async` functions and treat them as ordinary functions.** 


FIY: this is the code I have used for research on this subject.
```
// taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function
function TestAsyncClass(x) {
    this.x = x;
}

TestAsyncClass.prototype.resolveAfter2Seconds = function resolveAfter2Seconds(x) {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve(x);
    }, 2000);
  });
};

TestAsyncClass.prototype.add1 = async function add1() {
  let a = await this.resolveAfter2Seconds(20);
  let b = await this.resolveAfter2Seconds(30);
  return this.x + a + b;
};

function TestAsyncClassTargetMixin(d) {
    this.wrappedTarget = new TestAsyncClass(d);
}

async function dummyForName() {}

for (func in TestAsyncClass.prototype) {
    if (typeof TestAsyncClass.prototype[func] === "function") {
        // remove "false" for true async wrapper
        if (false && TestAsyncClass.prototype[func].constructor.name === dummyForName.constructor.name) {
            console.log("found async function: " + func);
            TestAsyncClassTargetMixin.prototype[func] = (function factory(funcName) {
                return async function () {
                    return await TestAsyncClass.prototype[funcName].apply(this.wrappedTarget, arguments);
                };
            })(func);
        } else {
            TestAsyncClassTargetMixin.prototype[func] = (function factory(funcName) {
                return function () {
                    return TestAsyncClass.prototype[funcName].apply(this.wrappedTarget, arguments);
                };
            })(func);
        }
    }
}


async function testIt() {

    let t1 = new TestAsyncClass(2);
    console.log(" value t1: " + (await t1.add1()));


    let t2 = new TestAsyncClassTargetMixin(2);
    console.log(" value t2: " + (await t2.add1()));

}
testIt();

console.log(testIt.constructor.name);
console.log(testIt.constructor);
```

sidorares:
hi @nros - by async functions I didn't mean `async` keyword, but just regular functions that return data via callback call.

this is for example how one might wrap `ping()` ( which is currently missing from promise wrappers )

```js
  PromiseConnection.prototype.ping = function (query, params) {
    var c = this.connection;
    return new this.Promise(function (resolve, reject) {
     c.ping(makeDoneCb(resolve, reject));
   });
  };
```

This is the whole point of promise wrapper - to convert from callback-style interface to Promise-based

`escape` returns result directly via normal function return, not using callback so it can be wrapped in a way you did:

```js
  PromiseConnection.prototype.escape = function (value) {
    return this.connection.escape(value)
  };
```

nros:
Ah, you had that in mind!

That is no problem at all. All functions with a callback ( = async functions) need to be implemented separately before the mixin generator runs.  The `if`-clause will check for already existing implementations.
```
&& PromiseConnection.prototype[func] === undefined
```
will ignore previously, manually created functions. So my additional code only adds functions not yet implemented.

Yes, such functions should be added to PromiseConnection, otherwise they would not be converted to a `Promise` returning version.

sidorares:
oh, missed that check. Might be fine then - anything that is not added to prototype is assumed to be  sync helper function

@sushantdhiman wdyt?

sushantdhiman:
`PromiseConnection` APIs are supposed to return `Promise`, Although escape is helper, I think many will still try to call `PromiseConn.escape().then()`. I think this unbalances the promise API, where a few functions are sync, others aren't.

Some alternatives to catch all patching

- If you want to use helper or extra function, just use `PromiseConnection.connection` directly
- As already suggested, use whitelist, I see many functions without `_` which are clearly not helpers. I suggest keeping two whitelists `async` and `sync`. Each async whitelist method will be named exactly as its in `connection.proto`, sync list methods will get `Sync` suffixed with their names in `Promise.connection`

Even if we dont want to rename functions when copying them, I would still recommend whitelist approach, only expose those function which are actually needed.

promise.js
```diff
@@ -68,6 +68,110 @@ PromiseConnection.prototype.end = function () {
...
+  'handlePacket',
+  'keyFromFields',
+  'pause',
+  'pipe',
+  'protocolError',
+  'resume',
+  'serverHandshake',
+  'unprepare',
+  'write'
+]);
```
sidorares: I'd probably remove addCommand, addCommand, keyFromFields, protocolError, serverHandshake, write - they are internal helpers ( note to me - add underscore prefix to them ). Could you remove them from list @nros?

nros:
done. I remove internal Functions to handle MySQL server communication.

sushantdhiman:
Needs a rebase, lots of ghost commits atm

nros:
Fine, I tried a rebase but it did not work as expected because the branch for this pull request was already pushed to the github server.

Do you see any possibility to exchange the branch of this pull request for another one? Then I could create a clean, rebased branch. I tried to figure this out with the github web interface but did not find a way. Maybe I have just overseen something. If there is no such possibility, I'll vote for closing this pull request and creating a new one - unless more additions, more commits are needed.

Please advise.

sushantdhiman:
Yeah, create a fresh PR. We can close this one when other is finally merged

nros:
see pull request #531

PR #508, opened 2017-02-08
Fix - release the connection if there is and error in the query

AyhamAlzoubi:
Currently when we use `pool.query` and the query returns an error, the connection is not released from the pool.

This fix will release the connection from the pool even if there was an error in the query.


tangxinfa:
Encountered this bug these days, hopes this problem fix asap, a example to show this bug: https://gist.github.com/tangxinfa/c2f27bf8681521dd907f6c0de0d051a8

sidorares:
fixed in #579

PR #516, opened 2017-02-25
Don't try to write undefined parameters.

vlasky:
This is a re-submission of @ChiperSoft's fix, as [requested](https://github.com/sidorares/node-mysql2/pull/480#issuecomment-277293179) by @sushantdhiman: 

> This addresses an exception that occurs when using namedParameters if the named param is missing from the values object. The code hits `this.parameters[i].toString()` and throws up because `this.parameters[i]` is undefined.

### DIFF
--- a/lib/packets/execute.js
+++ b/lib/packets/execute.js
@@ -32,7 +32,7 @@ Execute.prototype.toPacket = function ()
     length += 2 * this.parameters.length;  // type byte for each parameter if new-params-bound-flag is set
     for (i = 0; i < this.parameters.length; i++)
     {
-      if (this.parameters[i] !== null) {
+      if (this.parameters[i] !== null && this.parameters[i] !== undefined) {
         if (Object.prototype.toString.call(this.parameters[i]) == '[object Date]') {
           var d = this.parameters[i];
           // TODO: move to asMysqlDateTime()
@@ -95,7 +95,7 @@ Execute.prototype.toPacket = function ()
 
     for (i = 0; i < this.parameters.length; i++)
     {
-      if (this.parameters[i] !== null) {
+      if (this.parameters[i] !== null && this.parameters[i] !== undefined) {
         if (Buffer.isBuffer(this.parameters[i])) {
           packet.writeLengthCodedBuffer(this.parameters[i]);
         } else {


sidorares:
I understand what problem you trying to solve ( driver should not crash regardless of the input !!! ) but a bit concerned with your solution

` !== null ` check was about building 'null bitmask' - in binary protocol, [all NULL parameters are compressed](https://dev.mysql.com/doc/internals/en/null-bitmap.html), this part was not about checking if params are "defined"

with extra `!==undefined` check if you mistype field it's just silently ignored (hopefully you get back "incorrect number of parameters" sql error)

Is there any reason we want to send execute packet in that case? Maybe document "undefined variables are not allowed" in js<->mysql types documentation and make code error early if undefined parameter is passed?

vlasky:
At the moment, we are using prepared statements invoked with connection.execute(). When the wrong number of arguments is passed to execute(), the exception message does not include the actual SQL statement or the line of code from where connection.execute() was called.

This is making it virtually impossible to identify and debug the code that is causing the problem.

This is an example of what we currently get without the fix:

`W20161118-19:49:32.812(11)? (STDERR) /home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/packets/execute.js:47
W20161118-19:49:32.814(11)? (STDERR)           var str = this.parameters[i].toString();
W20161118-19:49:32.815(11)? (STDERR)                                       ^
W20161118-19:49:32.815(11)? (STDERR)
W20161118-19:49:32.816(11)? (STDERR) TypeError: Cannot read property 'toString' of undefined
W20161118-19:49:32.817(11)? (STDERR)     at Execute.toPacket (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/packets/execute.js:47:39)
W20161118-19:49:32.819(11)? (STDERR)     at Execute.start (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/commands/execute.js:50:40)
W20161118-19:49:32.819(11)? (STDERR)     at Execute.Command.execute (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/commands/command.js:39:20)
W20161118-19:49:32.819(11)? (STDERR)     at PoolConnection.Connection.handlePacket (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/connection.js:417:28)
W20161118-19:49:32.820(11)? (STDERR)     at PoolConnection.Connection.handlePacket (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/connection.js:424:12)
W20161118-19:49:32.820(11)? (STDERR)     at PacketParser.onPacket (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/connection.js:93:16)
W20161118-19:49:32.821(11)? (STDERR)     at PacketParser.executeStart (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/packet_parser.js:73:14)
W20161118-19:49:32.821(11)? (STDERR)     at Socket.<anonymous> (/home/apache/.meteor/packages/vlasky_mysql/.1.2.4.15e9rxz++os+web.browser+web.cordova/npm/node_modules/mysql2/lib/connection.js:101:29)
W20161118-19:49:32.822(11)? (STDERR)     at emitOne (events.js:77:13)
W20161118-19:49:32.823(11)? (STDERR)     at Socket.emit (events.js:169:7)`

We (and others) urgently need a runtime mechanism that will allow us to identify the specific SQL statement (and lines of code) for which undefined parameters are being passed.

As you correctly say, it would technically not be necessary to send the execute packet - it would be more efficient to have code to detect this beforehand and throw an exception.

However, this is not a situation that would happen very often - passing undefined arguments would almost always be programmer error and therefore would be fixed very quickly once discovered.

Therefore, I don't consider the efficiency difference to be a strong reason to not merge this pull request which implements an extremely simple, straightforward fix.

I would prefer this pull request to be merged so that everyone in a situation like ours can immediately debug and fix their code and let the more efficient approach be implemented later.

sidorares:
> strong reason to not merge this pull request which implements an extremely simple, straightforward fix.

I'm just suggesting even more simple fix. For example, add here https://github.com/vlasky/node-mysql2/blob/d347710630b49df709942dccd252f9630a11439f/lib/commands/execute.js#L49 something like this:

```js
if (!this.parameters.every(p => typeof p !== 'undefined')) {
  const err = new Error('Passed undefined parameter to prepared statement, query is ${this.sql}')
  if (this.onResult) {
      this.onResult(err);
  } else {
    this.emit('error', err);
  }
  return;
}
```

vlasky:
Yes that looks promising.

Can we make it so that the error object includes both the query and the array of parameters that was passed as separate properties? This information would help narrow down the cause.

This might require creating a custom ExecuteError object that derives from Error.

sidorares:
> Can we make it so that the error object includes both the query and the array of parameters that was passed as separate properties?

yes (probably using util.inspect? - so that large Buffer/arrays are trimmed automatically and not blow up logs)

> This might require creating a custom ExecuteError object that derives from Error.
I guess we could just put everything in error message ( standard `Error` constructor parameter )

The code I posted wasn't tested at all would you want to explore that approach yourself and create another PR? Also, need to add unit test

Sorry for nitpicking, just think this is better than applying temp solution first and refactor it after

n8ores:
What happened with this? We are still getting these errors which provide no context as to the query or parameters which caused the query to fail.

The code provided by @sidorares looked promising, why was it not incorporated? Even better if parameters can be returned in the error also as suggested by @vlasky, but understand that may cause an issue with large objects. Just having the query would probably be sufficient for us to at least know where the problem is occuring and thus work on a fix.

sidorares:
@vlasky would you have time to go through my suggestions? If not that's ok, I might be able to find some time over weekend to do that myself

vlasky:
@sidorares I see that the code has changed a bit since I last looked at it.

To help us debug a recent issue, I changed the following line

https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L617

from

`throw new TypeError('Bind parameters must not contain undefined. To pass SQL NULL specify JS null');`

to 

`throw new TypeError('Bind parameters must not contain undefined. To pass SQL NULL specify JS null. SQL: ' + options.sql);`

vlasky:
Another question @sidorares - is it possible in the code to get a count of the number of question mark placeholders contained within a prepared statement?

I want to put in a safety check to generate an exception if the number of question marks placeholders in the prepared statement does not equal the number of elements in the parameter value array passed to execute().

sidorares:
> is it possible in the code to get a count of the number of question mark placeholders contained within a prepared statement?

yes, result of `prepare` command has `parameterDefinitions` array - https://github.com/sidorares/node-mysql2/blob/0c422dd9d88a211015ddf4fcc8531048de9b1340/lib/commands/prepare.js#L119

see how it's passed to execute here: https://github.com/sidorares/node-mysql2/blob/0c422dd9d88a211015ddf4fcc8531048de9b1340/lib/connection.js#L643

PR #547, opened 2017-04-13
`packet.skip(23);` means 23 unsafe bytes leak over wire

igorklopov:
you may discard if not that important :)


sidorares:
thanks!

Issue #540, opened 2017-04-05
doubt about end connection


lagden:
I have this system...

**connection.js**
```js
const mysql = require('mysql2/promise')
async function connection(database) {
    return mysql.createConnection({
        host:'127.0.0.1',
        user: 'root',
        password: '',
        database,
        rowsAsArray: false
    })
}

module.exports = connection
```

**create.js**
```js
const connection = require('./connection')
async function create(data) {
	const q = `INSERT INTO example SET ?`
	const conn = await connection('mydb')
	const results = await conn.query(q, data)
	conn.end()
	return results
}

module.exports = create
```

**app.js**
```js
const create = require('./create')
async function crash() {
    try {
        // Force MySQL error
        const users = await Promise.all[create({id: 1, name: 'Lagden'}), create({id: 1, name: 'Lagden'})]
    } catch (err) {
        console.log(err.message) // Error: Duplicate entry '1' for key 'PRIMARY'
    }
}

crash()
```

In the system above, we will have a MySQL error...

My doubt are:
- Will my connection remain open?
- Do I have to handle it on `create` to avoid leaving opened connections?

sidorares:
it'll remain open, you need to wrap the code in create with try/catch:

```js
async function create(data) {
  try {
	const q = `INSERT INTO example SET ?`
	const conn = await connection('mydb')
	const results = await conn.query(q, data)
   } finally {
	conn.end()
	return results
   }
}
```
or just use connection pool (this is always better default option, you should have a good reason not to use it)

lagden:
@sidorares good point!!

Let me understand better the `connection pool`

if I do something like that:

**pool.js**
```js
const mysql = require('mysql2/promise')
const pool = mysql.createPool({
	connectionLimit : 3,
	host:'127.0.0.1',
	user: 'root',
	password: '',
	rowsAsArray: false,
	multipleStatements: true
})

module.exports = pool
```

**create.js**
```js
const pool = require('./pool')
async function create(data) {
	let r
	try {
		const [results] = await pool.query(`USE mydb; INSERT INTO example SET ?`, data)
		const [rUse, rIns] = results
		r = rIns
	} catch (err) {
		console.log(err.message)
	} finally {
		const {connection} = await pool.getConnection()
		connection.release()
	}
	return r || false
}

module.exports = create
```

**app.js**
```js
const create = require('./create')
async function multiple() {
	try {
		const users = await Promise.all([
			create({id: 1, name: 'A'}),
[...]

PR #551, opened 2017-04-18
Fix pool leak

tangxinfa:
See issue: https://github.com/sidorares/node-mysql2/issues/540
See pull request: https://github.com/sidorares/node-mysql2/pull/508

Pull request https://github.com/sidorares/node-mysql2/pull/508 not work, so i create a new PR.


sidorares:
hi @tangxinfa I fixed it that in #579 and then started looking at your pr and realised that what you doing also fixes part of the problem. Thanks for your contribution anyway

PR #555, opened 2017-04-25
Promise Wrapper extension

Hookyns:
Third parameter (callback) added to PromiseConnection.prototype.query and PromiseConnection.prototype.execute which makes original query and execute return values available.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -35,32 +35,36 @@ function makeDoneCb (resolve, reject) {
   };
 }
 
-PromiseConnection.prototype.query = function (query, params, queryCb) {
+function makeDoneCbWithExtras (resolve, reject, extras) {
+  return function (err, rows, fields) {
+    if (err) {
+      reject(err);
+    } else {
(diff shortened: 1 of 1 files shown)


sidorares:
feels a bit strange to me - why promise wrapper in the first place if callback required? Can you describe your use case?

Hookyns:
I'm awaiting promise as usual but I want to log executed query which is stored in original return object in sql property. This is quite clear way (for me) how to access that returned object like.. "hey btw,.. what was original return value?"

Or is there another way how to access executed SQL when using promise?

Hookyns:
I've pushed new commit which change the first one solution with callback to the last one I've mentioned in comment above as EDIT4. Feel free to reject this PR if you don't like it. I can rewrite it just in my project, but it would be nice to have access to executed SQL.

sidorares:
I prefer this to be done by other means - calling `format(sql, parameters)` upfront or maybe introducing some generic hooks mechanism for logging and extension

PR #562, opened 2017-05-18
Clarifying `if` statement formatting

moberemk:
Currently the `if` block looks like it's part of the larger `if/else` block above it, but it's a separate declaration. Adding a newline makes it immediately clear that these are separate logical blocks to readers of the code.


sidorares:
yes, thanks. Not part of `if/else`
I'v very keen to integrate prettier, did with my day job projects and works perfectly

PR #567, opened 2017-05-26
Make Promise Pool wrapper extend EventEmitter

### LINKED ISSUE #468: The promise version of the API does not support events
The promise version of the API does not support events. 

Because of this, the following code throws an error that 'on' is not a function.
```js
pool = mysql.createPoolPromise(config);
pool.on('enqueue', function () {
  console.log('Waiting for available connection slot');
});
```

This is unnecessary because it is an easy fix. You can easily fix it by hooking it to the core pool:
```js
function createPool (opts) {
  var corePool = core.createPool(opts);
  var Promise = opts.Promise || global.Promise || require('es6-promise');

  var promisePool = {
    getConnection: function () {
      ...
    },

    query: function (sql, args) {
      ...
    },

    execute: function (sql, args) {
      ...
    },

    end: function () {
      ...
    },

    // This adds support for events
    on: corePool.on.bind(corePool)
  };

  
  return promisePool;
}
```

Now it works just as normal.

If people disagree with my solution, I think that it might be a good idea to make it possible to use the core pool from the promise version. This does not break anything, but allows developers to use the same pool in places where promises are not preferred.
Something like this:
```js
pool = mysql.createPoolPromise(config);
pool._internalPool.on('enqueue', function () {
  console.log('Waiting for available connection slot');
});
```

Myhlamaeus:
Fix #468

Introduce a new PromisePool constructor into promise.js. PromisePool
inherits from EventEmitter, and the following events are re-emitted
from the underlying Pool instance to it:
 - acquire
 - connection
 - enqueue
 - release

Methods like `on` can't just be bound to the Pool and added to the object as `this` would be
wrong in the callback: e.g. `this.query` wouldn't return a Promise.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -2,6 +2,18 @@ var core = require('./index.js');
 var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 
+function inheritEvents(source, target, events) {
+  events
+    .forEach(function (eventName) {
+      source.on(eventName, function () {
+        var args = [].slice.call(arguments);
+        args.unshift(eventName);
+
+        target.emit.apply(target, args);
+      });
+    });
+}
+
 function createConnection (opts) {
   var coreConnection = core.createConnection(opts);
   var Promise = opts.Promise || global.Promise;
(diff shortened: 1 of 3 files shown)


ben-z:
@Myhlamaeus While we are at this, can we also [make PromiseConnection inherit EventEmitter](https://github.com/sidorares/node-mysql2/issues/468#issuecomment-260626705)?

Myhlamaeus:
Sure, just give me a moment.

Issue #574, opened 2017-05-31
Promise wrapper issues for `prepare`


faazshift:
The recently released promise wrapper for `prepare` doesn't accept any arguments, but instead passes a callback to the underlying `prepare` function. Additionally, the returned `statement` object doesn't get wrapped, preventing a workflow like `statement.execute(...).then()`.

See: https://github.com/sidorares/node-mysql2/blob/master/promise.js#L132-L143

sidorares:
thanks for report @faazshift 
do you have time to provide a fix? seems fairly trivial. Also would be good to add tests

If not I'll try to do this over weekend

faazshift:
@sidorares Done. I went ahead and fixed both of the issues I mentioned in PR #575.

sidorares:
published as v1.3.2

PR #575, opened 2017-05-31
Fixing promisified prepare implementation and promisifying returned statement

faazshift:
This fixes both issues mentioned in #574.


sidorares:
Legend!

PR #583, opened 2017-06-08
Fix node encodings lookup in string parser

### LINKED ISSUE #582: string parser looks up in NODE_ENCODING incorrectly
Is this a bug?

https://github.com/sidorares/node-mysql2/blob/master/lib/parsers/string.js#L15
https://github.com/sidorares/node-mysql2/blob/master/lib/parsers/string.js#L28

`NODE_ENCODING` here is an array, `encoding` function arguments receives string values like `binary`, `cesu8`, `utf8`, etc (at least, in `decode` function). They will never be found using `encoding` variable as array key. Looks like we should scan array using `.indexOf` for this to work properly.

Does triggering `Iconv` decoder for every charset could give significant performance penalty on parsing rows?

dolphin278:
This fixes #582

### DIFF
--- a/lib/parsers/string.js
+++ b/lib/parsers/string.js
@@ -1,15 +1,18 @@
 var Iconv = require('iconv-lite');
 
-var NODE_ENCODING = {
-  'ascii': true,
-  'utf8': true,
-  'utf16le': true,
-  'ucs2': true,
-  'base64': true,
-  'latin1': true,
-  'binary': true,
-  'hex': true
-};
+var NODE_ENCODING = [
+  'ascii',
+  'utf8',
+  'utf16le',
+  'ucs2',
+  'base64',
+  'latin1',
+  'binary',
+  'hex'
+].reduce(function (map, item) {
+  map[item] = Buffer.isEncoding(item);
+  return map;
+}, {});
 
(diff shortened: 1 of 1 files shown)


dolphin278:
`latin1` is not supported in Node v4.x (this behavior was hidden due to bug we trying to fix here):

* https://nodejs.org/dist/latest-v4.x/docs/api/buffer.html#buffer_buffers_and_character_encodings
* https://nodejs.org/dist/latest-v7.x/docs/api/buffer.html#buffer_buffers_and_character_encodings

What should we do? Treat is as non node-supported one for compat, or check for Node version. @sidorares @sushantdhiman ?

dolphin278:
Figured it out - we could use `Buffer.isEncoding` at startup time.

dolphin278:
All done, test passed.

sidorares:
Thanks @dolphin278 !

dolphin278:
👍

dolphin278:
@sidorares when are you planning next npm release?

sidorares:
@dolphin278 done, v1.3.3 ( changelog: d5defad183fdc9d17a0457e16b9e567bf6be1ca3  )

dolphin278:
@sidorares Thank you!

PR #603, opened 2017-06-28
fixed method name in promise with co documentation

dotnil:
att


sidorares:
it's correct there - see https://github.com/sidorares/node-mysql2/blob/03683f433f621e206814b837e46bc7f46ef45e0c/index.js#L37

there are two separate apis, available with *Promise suffix or without suffix if you `require('mysql2/promise')`

dotnil:
ah I see. I jumped to pr too hasty I suppose. 😭

PR #618, opened 2017-07-26
Read ms in Packet.prototype.readDateTimeString

### LINKED ISSUE #617: dateStrings: true issue with TIMESTAMP(3) column
Hi,

We use `dateStrings: true` in our config so we can parse the date ourselves and found an issue with a `TIMESTAMP(3)` column.
Mysql version: 5.7. I've created a simple test script below to reproduce this.

mysql setup:
```
CREATE TABLE tst (id INT, d TIMESTAMP(3), name VARCHAR(10));
INSERT INTO tst VALUES(1, '2017-07-26 09:36:42.000', 'John');
INSERT INTO tst VALUES(2, '2017-07-26 09:36:42.123', 'Jane');
```

script:
```
'use strict';

const mysql = require('mysql2/promise');
  
const query = `SELECT id, d, name FROM tst WHERE id = ?;`;
const query2 = `SELECT id, name, d FROM tst WHERE id = ?;`;

mysql.createConnection({
  "host": "mysql",
  "user": "test",
  "password": "hello",
  "database": "test",
  "dateStrings": true
})
.then((connection) => {
  connection.execute(query, [1])
    .then((res) => {
      console.log(res[0]);
      return connection.execute(query, [2]);
    })
    .then((res) => {
      console.log(res[0]);
      return connection.execute(query2, [1]);
    })
    .then((res) => {
      console.log(res[0]);
      return connection.execute(query2, [2]);
    })
    .then((res) => {
      console.log(res[0]);
      process.exit(1);
    })
})
.catch(e => {
  console.log(e);
  process.exit(1);
});
```

output:
```
[ BinaryRow { id: 1, d: '2017-07-26 09:36:42', name: 'John' } ]
[ BinaryRow {
    id: 2,
    d: '2017-07-26 09:36:42',
    name: '�\u0001\u0000\u0004Jane\u0005\u0000\u0000\u0007�\u0000\u0000"\u0000' } ]
[ BinaryRow { id: 1, name: 'John', d: '2017-07-26 09:36:42' } ]
[ BinaryRow { id: 2, name: 'Jane', d: '2017-07-26 09:36:42' } ]
```

If the date is at the end of the query everything is ok (3rd and 4th BinaryRow in the output), however all fields after the date are incorrect (only happens with fields which ms precision is not `000`, 2nd BinaryRow in the output). After a quick search I've found out that it has to do with the buffer read incorrectly in (packet.js)[https://github.com/sidorares/node-mysql2/blob/master/lib/packets/packet.js]
If I uncomment the code in (Packet.prototype.readDateTimeString)[https://github.com/sidorares/node-mysql2/blob/master/lib/packets/packet.js#L295] all values are correct.

output after uncommenting:
```
[ BinaryRow { id: 1, d: '2017-07-26 09:36:42', name: 'John' } ]
[ BinaryRow { id: 2, d: '2017-07-26 09:36:42', name: 'Jane' } ]
[ BinaryRow { id: 1, name: 'John', d: '2017-07-26 09:36:42' } ]
[ BinaryRow { id: 2, name: 'Jane', d: '2017-07-26 09:36:42' } ]
```

Should I open a PR to uncomment that code seeing as in new version of mysql it is not working as expected or do you know of a case (hence the comment in the code) where it could cause and we need to take a different approach? I am happy to help either way.

bencebalogh:
Fixes #617

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -117,7 +117,8 @@ Packet.prototype.readInt64 = function() {
   var word0 = this.readInt32();
   var word1 = this.readInt32();
   var res = new Long(word0, word1, true);
-  var resNumber = res.toNumber(), resString = res.toString();
+  var resNumber = res.toNumber(),
+    resString = res.toString();
 
   res = resNumber.toString() === resString ? resNumber : resString;
 
@@ -128,7 +129,8 @@ Packet.prototype.readSInt64 = function() {
   var word0 = this.readInt32();
   var word1 = this.readInt32();
   var res = new Long(word0, word1, false);
-  var resNumber = res.toNumber(), resString = res.toString();
+  var resNumber = res.toNumber(),
+    resString = res.toString();
 
   res = resNumber.toString() === resString ? resNumber : resString;
 
@@ -194,7 +196,8 @@ Packet.prototype.readLengthCodedNumberExt = function(
 
     res = new Long(word0, word1, !signed); // Long need unsigned
 
-    var resNumber = res.toNumber(), resString = res.toString();
+    var resNumber = res.toNumber(),
+      resString = res.toString();
 
     res = resNumber.toString() === resString ? resNumber : resString;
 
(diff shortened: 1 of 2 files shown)


sidorares:
Thanks!

Issue #577, opened 2017-06-02
 MaxListenersExceededWarning: Possible EventEmitter memory leak detected


aankur:
Hi ,

i am using mysql2 with bluebird on nodejs 8.

as 

```javascript
const mysql = require('mysql2/promise');
const Promise = require("bluebird");
config.Promise = Promise;//plus database connection configuration
pool = mysql.createPool(config);

function  GetSqlConnection() {
    return  pool.getConnection().disposer(function(connection) {
        connection.release();
    });
};
```
after sometime i get hit by this warning,
i get , 11 end listeners added. 11 drain  listeners added etc ...

```javascript

at _addListener (events.js:281:19)
    at PoolConnection.addListener (events.js:298:10)
    at /home/test/servers/v6/project/node_modules/mysql2/promise.js:8:14
    at Array.forEach (native)
    at inheritEvents (/home/test/servers/v6/project/node_modules/mysql2/promise.js:7:6)
    at new PromiseConnection (/home/test/servers/v6/project/node_modules/mysql2/promise.js:37:3)
    at /home/test/servers/v6/project/node_modules/mysql2/promise.js:228:17
    at /home/test/servers/v6/project/node_modules/mysql2/lib/pool.js:36:14
    at _combinedTickCallback (internal/process/next_tick.js:95:7)
    at process._tickCallback (internal/process/next_tick.js:161:9)
---------------------------------------------
    at process.emitWarning (internal/process/warning.js:154:18)
    at _addListener (events.js:289:17)
    at PoolConnection.addListener (events.js:298:10)
    at /home/test/servers/v6/project/node_modules/mysql2/promise.js:8:14
    at Array.forEach (native)
    at inheritEvents (/home/test/servers/v6/project/node_modules/mysql2/promise.js:7:6)
    at new PromiseConnection (/home/test/servers/v6/project/node_modules/mysql2/promise.js:37:3)
    at /home/test/servers/v6/project/node_modules/mysql2/promise.js:228:17
    at /home/test/servers/v6/project/node_modules/mysql2/lib/pool.js:36:14
    at _combinedTickCallback (internal/process/next_tick.js:95:7)
    at process._tickCallback (internal/process/next_tick.js:161:9)
---------------------------------------------
    at Pool.getConnection (/home/test/servers/v6/project/node_modules/mysql2/lib/pool.js:35:20)
    at /home/test/servers/v6/project/node_modules/mysql2/promise.js:224:14
[...]

PR #620, opened 2017-07-28
redo event delegation in promise wrappers to be lazy, self-cleaning

rhendric:
fix sidorares/node-mysql2#577

---

I saw that #577 already had a fix in #578, but you seemed to have some concerns about that solution that I think I share, and I'm eager to see a fix for this issue actually land. So here's what I think the fix should be: it's fully lazy, so no event listeners get added to the wrapped object until one is added to the wrapper, and it removes those listeners as early as possible.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -3,13 +3,21 @@ var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 
 function inheritEvents(source, target, events) {
-  events.forEach(function(eventName) {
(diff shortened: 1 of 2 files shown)


sidorares:
Thanks, looking good, this is close to what I wanted to do (important part is that no listeners added at all by default - most users don't add any listeners on pool directly )

I'll have a closer look later tonight and try to merge!

PR #626, opened 2017-08-10
Added connection pool documentation

ulikoehler:
No code changes, only README changes.

I provided a quick guide on how to use connection pools, as it's currently not documented at all and (although it's quite simple) you have to read the source code to understand how to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ __Table of contents__
   - [Installation](#installation)
   - [First Query](#first-query)
   - [Using Prepared Statements](#using-prepared-statements)
+  - [Using connection pools](#using-connection-pools)
   - [Using Promise Wrapper](#using-promise-wrapper)
   - [API and Configuration](#api-and-configuration)
   - [Documentation](#documentation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,6 +115,40 @@ connection.execute(
   }
 );
 ```
+
+## Using connection pools
+
+Connection pools provide automatic ways of managing multiple connections. A pool maintains a queue of *free* connections, allowing the use of multiple connections in parallel.
```
aikar: The wording is not correct here.

A better way of describing them would be:
"Connection pools help reduce the time spent connecting to the MySQL server by reusing a previous connection, leaving them open instead of closing when you are done with them.

This improves the latency of queries as you avoid all of the overhead that comes with establishing a new connection."

sidorares:
@ulikoehler can you also update Query/query casing I commented?

ulikoehler:
@sidorares I can't find the comment where you mentioned that, but I found the issue by looking at `pool.js`.
Besides Query => query I also updated Execute => execute. Hopefully that fixes all casing issues.

sidorares:
looks good, thanks!

ulikoehler:
@sidorares  Thanks for merging!

PR #629, opened 2017-08-15
correctly parse microsecond in datetime type

Dieken:
For example, MySQL binary protocol data in little endian for
2017-08-14 23:55:29.327:

```
hex:    0b      e107   08   0e     17   37  1d   58fd0400
dec:    11      2017   8    14     23   55  29   327000
unit:   bytes   year   mon  day    hour min sec  microseconds
```

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -253,7 +253,7 @@ Packet.prototype.readDateTime = function() {
     S = this.readInt8();
   }
   if (length > 10) {
-    ms = this.readInt32();
+    ms = this.readInt32() / 1000;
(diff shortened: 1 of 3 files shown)


sidorares:
can you add integration test for this? I guess related is #618 so cc @bencebalogh

Dieken:
@sidorares  sorry, I just updated the broken integration test and added a new one for #629.

about #618, I'm surprised MySQL uses millisecond for text protocol and microsecond for binary protocol,  no idea why....

bencebalogh:
lgtm

PR #635, opened 2017-08-21
Cleanup ES7 example

aikar:
no reason for this to be commented out.


sidorares:
thanks @aikar 
eslint fails on this example ( I think that was original reason why I left it commented out ).
If you can fix this that would be great

sidorares:
Thanks!

PR #652, opened 2017-10-14
test(typecast): first null value, corrupts next field

sushantdhiman:
If first field  is null then it will corrupt next field value to become null as well.

I dont know reason behind this but this only happens with `typeCast`, even when only `next()` is called. I guess somehow depend upon access order to packet buffer.

https://github.com/sequelize/sequelize/issues/8475

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     env: LINT=0
   - node_js: "6.11"
     env: LINT=0
-  - node_js: "8.7"
+  - node_js: "8.0"
     env: LINT=1
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,9 @@ matrix:
   include:
   - node_js: "4.8"
     env: LINT=0
-  - node_js: "6.10"
+  - node_js: "6.11"
     env: LINT=0
-  - node_js: "7.10"
-    env: LINT=0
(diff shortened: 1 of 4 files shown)


sidorares:
the problem is in the code for ['read NULL'](https://github.com/sidorares/node-mysql2/blob/cad78200f67849119eeebb50a3490ebbb160b72f/lib/compile_text_parser.js#L156) - `null; packet.skip(1)`. When this code is used as right hand side part of assignment it works as expected (assign null and skip one byte):

```js
    this["test"] = null; packet.skip(1);
    this["value"] = packet.parseLengthCodedInt(false);
```

but when used as part of 'next()' body `return null;` hides the rest and read position is stuck at null field

sidorares:
probably easy solution would be to add `packet.readNull` ( skip one byte and return null, maybe in debug mode assert that byte is `0xfb`) 

( or maybe just `packet.readLengthCodedNumber();` which should return null there, NULL in text protocol is one byte `0xfb` )

sushantdhiman:
Thanks @sidorares I was mostly offline to celebrate Diwali this week. Fix looks good to me.

Dont know why tests fails on Node 8, will try rebuild

sushantdhiman:
Hmm https://github.com/npm/npm/issues/17858, will degrade node version for now

PR #660, opened 2017-10-27
Milliseconds for setting date values

vasyas:


sidorares:
Thanks! Would you be able to also add test for this?

vasyas:
Ok, will do.

vasyas:
Add tests in test-datetime.js. Also, updated how fractional seconds are read for tests to be working.
For this I also changed behavior in test-#617.js, now dateStrings is reflecting precision set on column type. Please review.

sidorares:
Looks good!

Note that in binary protocol ( aka 'prepared statements' ) we currently send everything as a string while for some types it might have more sense to use better types ( for example js dates serialise back using DATETIME )

PR #662, opened 2017-10-28
Correct await calls documentation to use Promise based versions

mr6502:
Without the *Promise postfix added Node will throw the following error:

```
(node:23673) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: (intermediate value) is not iterable
(node:23673) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### DIFF
--- a/documentation/Promise-Wrapper.md
+++ b/documentation/Promise-Wrapper.md
@@ -29,13 +29,13 @@ In addition to errback interface there is thin wrapper to expose Promise-based a
 ```js
 async function example1 () {
   const mysql = require('mysql2/promise');
-  const conn = await mysql.createConnection({ database: test });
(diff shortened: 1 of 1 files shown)


sidorares:
hi @mr6502 , I think documentation is correct.

main export from `mysql2` is `createConnection`, `createPool`, `createConnectionPromise`, `createPoolPromise`. `require('mysql2').createConnecttionPromise` same as doing `require('mysql2/promise').createConnecttion`

mr6502:
@sidorares Yes, you're right. I hadn't noticed the require() was pulling from mysql2/promise and not mysql2 direct.

PR #669, opened 2017-11-06
Expose escape & format methods on connection pool

### LINKED ISSUE #663: PromisePool doesn't proxy synchronous functions
For example:
```javascript
const {createPool} = require('mysql2/promise');
const pool = await createPool(options);
console.log(typeof pool.escape) // => 'undefined'
```
Presumably it could do something similar to [this](https://github.com/sidorares/node-mysql2/blob/8dc27c516472ad5068f136c28e70308f1e97640e/promise.js#L235-L273)?

c24w:
Escape methods already exist on `Pool`, so are only new to `PromisePool`.

(The pre-commit hook went to town prettifying the code.)

Fixes #663.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,25 +3,28 @@ var SqlString = require('sqlstring');
 var Connection = require('./lib/connection.js');
 var ConnectionConfig = require('./lib/connection_config.js');
 
-module.exports.createConnection = function (opts) {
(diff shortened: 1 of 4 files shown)


sidorares:
Thanks!

PR #671, opened 2017-11-09
Remove console log in promise wrapper

c24w:
Just realised I left this in previously :man_facepalming:


sidorares:
maybe worth adding to eslint rules

c24w:
Yes, although there are quite a few `console.log`s surrounded by `if (config.debug)` conditions :/

PR #684, opened 2017-11-20
writeString and readString default encoding

shtse8:
There is a bug in connection.createBinlogStream() as there is a missing encoding.
we should set the default encoding, once the encoding is empty.

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -388,6 +388,7 @@ Packet.prototype.readString = function(len, encoding) {
     len = this.end - this.offset;
   }
   this.offset += len;
+  encoding = encoding || 'utf8'
   return StringParser.decode(
     this.buffer.slice(this.offset - len, this.offset),
     encoding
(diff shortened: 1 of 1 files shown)


sidorares:
I'd probably prefer to treat `encoding` as required parameter for read/writeString. Can you reproduce issue again ant trace who called them without setting encoding? It's better to fix caller

shtse8:
https://github.com/sidorares/node-mysql2/blob/5f0fb8f1f5035e2c0207490aa2f0b838dc82fdc2/lib/packets/binlog_dump.js#L26

shtse8:
I have tried to fix the caller, but due to some `readString()` is reading the whole length. It's hard to pass the length to the first argument in order to setting the encoding at the second argument. Any ideas?

sidorares:
you can do `.readString(undefined, encoding)`. And encoding is probably sent somewhere at the beginning - there is no way it can be undefined or assumed utf8 bu default due to mysql history ( binlog protocol is older than utf )

Issue #493, opened 2017-01-17
execute.js not testing for undefined


GotJimmy:
I ran into a problem passing an undefined value to a query which caused the node server to crash. The problem was found in execute.js for all 3 occurrences of
`this.parameters[i] !== null` 
Changing these 3 occurrences to
`this.parameters[i] !== undefined`
eliminated the problem. Checking for null and undefined would probably be ideal.

sidorares:
How do you think `undefined` should be handled? Javascript `null` is mapped to mysql `NULL` type, but there is no "undefined" type in mysql. I feel like the best behaviour should be to return error early (with message like 'Parameter cannot be undefined, did you want to use NULL value?')

At the moment all non-null (and non-Buffer) values a serialised as string result of `value.toString()` (which obviously crashes for `undefined`)

JimmyGot:
I didn't want to send null nor undefined -- it just happened because I JSON decode an url-encoded request body. That's how the undefined value got executed and caused the crash on line 47 of execute.js because of .toString() from an undefined parameter.
`var str = this.parameters[i].toString();`
However, I changed line 35   `if (this.parameters[i] !== null) {`
to  `      if (this.parameters[i] !== undefined) {`
and therefore never got to line 47. I also had to change lines 67 and 89 and no error occurred anymore.

sidorares:
so I guess in your case early error would work best

JimmyGot:
Well, in this case it would avoid the error. Of course, I'll add additional checking of my parameters before sending them to execute. But if sending an undefined parameter happened to me it may also happen to someone else by accident.

dougwilson:
@sidorares FWIW, `sqlstring` treats both `undefined` and `null` as SQL `NULL`, and I do agree that at least the way in which the packet execute is throwing is hard to understand. Perhaps either `this.parameters[i] !== null` -> `this.parameters[i] != null` to behave similar to `sqlstring` or at least `this.parameters[i].toString()` -> `String(this.parameters[i])` which will call `.toString()` if it exists, otherwise `Object.prototype.toString`.

sidorares:
I think I'll do `String(this.parameters[i])` and "error if this.parameters[i] is undefined" together. I don't think that sending `'undefined'` as string if you misspelled key name is good behaviour

sushantdhiman:
[...]

Issue #187, opened 2015-08-09
named placeholders can cause exception


abellina:
Hello

I am getting an issue where named placeholder resolution can cause exceptions, say if you use a column that doesn't exist in the table (misspelled). However, regular query calls with ? placeholders usually return errors as a parameter in the .query callback.

Any reason to favor exceptions in the named placeholder case?

Thanks

Alessandro

sidorares:
Hi! No reason, I'd prefer to have similar error flow. Placeholder resolution needs to be wrapped in try/catch. ( Feel free to submit PR, this is relatively easy to fix )

cheshirecatalyst:
This issue would be related to this error I get on prepared statements with named placeholders correct:
```
TypeError: Cannot read property 'toString' of undefined
    at Execute.toPacket (/Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/packets/execute.js:44:39)
    at Execute.start (/Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/commands/execute.js:44:40)
    at Execute.Command.execute (/Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/commands/command.js:34:20)
    at PoolConnection.Connection.handlePacket (/Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/connection.js:310:28)
    at PoolConnection.Connection.addCommand (/Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/connection.js:326:10)
    at Prepare.executeStatement [as onResult] (/Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/connection.js:492:18)
    at /Users/emris/hq/AvailabilityToOffer/development/calculate-offer/node_modules/hqmysql/node_modules/mysql2/lib/commands/prepare.js:98:12
    at doNTCallback0 (node.js:408:9)
    at process._tickCallback (node.js:337:13)
```

Which is a very unhelpful error ^_^ as there is no way to trace it back to which query is causing it

PR #687, opened 2017-11-23
Add configurable logging capability to connection

swftvsn:
Adds logging (configurable) to connection to diagnose missing parameters issues.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -577,9 +577,15 @@ Connection.prototype._resolveNamedPlaceholders = function(options) {
     if (convertNamedPlaceholders === null) {
       convertNamedPlaceholders = require('named-placeholders')();
     }
+    if (this.config.logNamedPlaceholdersConversion || options.logNamedPlaceholdersConversion) {
+      console.log('Converting named parameters. Original SQL: ', options.sql, 'Original parameters; ', options.values)
+    }
     unnamed = convertNamedPlaceholders(options.sql, options.values);
     options.sql = unnamed[0];
     options.values = unnamed[1];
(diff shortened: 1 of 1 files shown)


sidorares:
maybe this should go to debug log ( when `debug` flag is true ) , perhaps using https://www.npmjs.com/package/debug ? I don't want to pass all possible logging options via main api

swftvsn:
Actually the best solution would be just to have a failsafe catch, that would catch any errors and exceptions and log those + the current SQL + values. I'm just unsure of the correct location to add this. First I thought it would be connection.execute, but due to the async nature I'm not sure if catch there would actually catch anything. What do you think @sidorares

sidorares:
try/catch used to be performance killer, maybe we should add more of those even to guard against internal code problems. To direct that particular error callback you'd need try/catch around https://github.com/sidorares/node-mysql2/blob/5f65614a0d61e4ea23d6b1ce0f75548205ba49f7/lib/commands/execute.js#L48 and call `this.onResult(error)` in catch block

swftvsn:
I see what you mean. I've added a try-catch wrap to PR #689, but if try is performance hazard, maybe we should pass the callback down to all *.toPacket methods so errors could be dealt with there. This, however is a big change, so I'll not prepare a PR for that. I'm happy with the fix in #689, so if it's ok to add that, please go ahead and merge and release.

It would fix these things: (all symptoms of the same problem, which actually is bad input, but this makes the library more developer friendly by informing the dev better what the problem is)
1. #493
2. #187
3. #480

sidorares:
we have early error reporting on missing parameters now, closing this

PR #689, opened 2017-11-24
Wrap executePacket.toPacket(1) call with try/catch

swftvsn:
This is to provide more context to error messages when a parameter is undefined.


sidorares:
now it's not about undefined parameter, but "if there is some crash in sync path help reporting it in the right place"

Issue #642, opened 2017-09-08
timezone pain


magnusjt:
The timezone option does not appear to do anything. If this package is supposed to be a dropin replacement for node-mysql, why isn't this at least mentioned somewhere? I've seen a few attempts at using the typeCast function instead, but this function doesn't work with pool (at least when I tried it). Are you planning on adding any of these soon?

The reason I need these is because my node application and the mysql server are running on different timezones, and I don't know what they are beforehand.

ssypi:
typeCast does work in pool. Are you using .execute? typeCast seems only to work currently with .query as mentioned in #649

magnusjt:
You're probably right about that.

In the meantime I've switched back to https://github.com/mysqljs/mysql

I'm using the timezone settings shown below:

````js
let config = {
	// ...
	timezone: 'UTC', // Interpret all received timestamps as UTC. Otherwise local timezone is assumed.
	dateStrings: [
		'DATE', // DATE's are returned as strings (otherwise they would be interpreted as YYYY-MM-DD 00:00:00+00:00)
		'DATETIME' // DATETIME's return as strings (otherwise they would interpreted as YYYY-MM-DD HH:mm:ss+00:00)
	]
};

let pool = mysql.createPool(config);
pool.on('connection', conn => {
	conn.query("SET time_zone='+00:00';", error => {
		if(error){
			throw error;
		}
	});
});
````

These settings work no matter what the client/server timezones are.

magnusjt:
I didn't notice this was fixed until now. Just wanted to say, thanks for fixing it!

Issue #633, opened 2017-08-18
Timezone seems not work


LichFaker:
I create a mysql pool and set the timezone option to `Asia/Shanghai`,but it still differ for 8 hours when I use `DATE(exp)` in sql.
```
const pool = mysql.createPool({ 
    host: dbCfg.host, 
    user: dbCfg.userName, 
    password: dbCfg.password,
    database: dbCfg.databaseName,
    connectionLimit: dbCfg.limit,
    timezone: "+08:00", // also try: "UTC+8","Asia/Shanghai", "local" but nothing different
})
```
Is there something wrong in my code?

sidorares:
looking at the code I don't see timezone applied at all anywhere

Note that this meant to be client side transformation, mysql DATE type does not have any tz info in it. You can emulate `timezone` config option with custom `typeCast` https://github.com/mysqljs/mysql#type-casting :

(haven't tested this)
```js
  typeCast: function (field, next) {
    if (field.type == 'DATETIME') {
      return new Date(field.string() + '+08:00')
    }
    return next();
  }
```

looks like timezone option from config need to be applied here - https://github.com/sidorares/node-mysql2/blob/5f0fb8f1f5035e2c0207490aa2f0b838dc82fdc2/lib/packets/packet.js#L606 called from function generated at https://github.com/sidorares/node-mysql2/blob/5f0fb8f1f5035e2c0207490aa2f0b838dc82fdc2/lib/compile_text_parser.js#L173

LichFaker:
thx, `typeCast` not worked for me. 
However, I just find a simple way to [solve this issue:](https://stackoverflow.com/questions/23571110/nodejs-responded-mysql-timezone-is-different-when-i-fetch-directly-from-mysql) config `dateStrings : true`

darksabrefr:
Why not adding the timezone option to known incompatibilities with node-mysql (https://github.com/sidorares/node-mysql2/tree/master/documentation#known-incompatibilities-with-node-mysql) ?

sidorares:
good idea @darksabrefr . Do you want to submit PR with docs update?

darksabrefr:
PR request submited here : https://github.com/sidorares/node-mysql2/pull/752

wisetc:
Why not set `timezone: "+00:00"`, cause your mysql server default timezone set to `UTC` maybe @LichFaker intersting

Issue #262, opened 2016-02-02
Timezone is wrong on client when reading records from SQL server running in UTC time


jasonhillier:
Using mysql2 library on my machine running in Pacific time, the records, when deserialized from SQL response, seem to be applying local timezone offset but still keeping date object marked at UTC. This means the timezone offset is actually applied backwards:

Time in db record (SQL server set to UTC): **2016-02-01 21:39:46**
Time returned by mysql2 (running on a machine in Pacific TZ): **2016-02-02T05:39:46.000Z** (this is a UTC date in future, which is wrong)
Time returned by mysql2 when running on a machine with TZ set to UTC: **2016-02-01T21:39:46.000Z** (this is correct time)

Expected result: I would expect toISOString() to return the same UTC date value regardless of client TZ, when reading the same record from the same server.

I believe this code is responsible for inserting SQL's UTC date into a JS object which is instantiated in local time:

`Packet.prototype.readDateTime:
...
  return new Date(y, m-1, d, H, M, S, ms);`

sidorares:
Server does not store TZ value in the data row, the only information we have at the time we receive row is "2016-02-01 21:39:46"

Also the client does not now which TZ server is running.

> Expected result: I would expect toISOString() to return the same UTC date value regardless of client TZ, when reading the same record from the same server.

I need to double check if this is supported or not, but the only solution I know is to pass TZ to a client, I think this is what node-mysql is doing. This way you do "this client would treat dates as TZ, because I know that server has it's timezone set TZ"

sidorares:
see 'timezone' option in https://github.com/felixge/node-mysql#connection-options

jasonhillier:
Hi sidorares,

I tried setting the 'timezone' option on the SQL connection string, it didn't make a difference.

MySQL does have a way of querying it to find out the SQL connection session's timezone:

More info here:
http://stackoverflow.com/questions/930900/how-to-set-time-zone-of-mysql

sidorares:
yes, you can get offset by doing `SELECT TIMEDIFF(NOW(), UTC_TIMESTAMP)` but I don't want all clients to perform this query at connection time just in case this information is needed later.

> I tried setting the 'timezone' option on the SQL connection string, it didn't make a difference.
@jasonhillier - looks like this functionality is missing, I'll try to add it

[...]

PR #705, opened 2018-01-09
Use source parameter types

gazoakley:
This PR is a follow on from https://github.com/sidorares/node-mysql2/pull/353 - it sets the type for bind parameters based on the source parameter type.

- Numbers are typed as `DOUBLE`
- JSON is typed as `JSON`
- Booleans are typed as `TINY`
- Dates are converted to string

There's a PR for sequelize that relies on these changes: https://github.com/sequelize/sequelize/pull/8861

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     env: LINT=0
   - node_js: "6.11"
     env: LINT=0
-  - node_js: "8.0"
+  - node_js: "8.9"
     env: LINT=1
 
 cache:
--- a/lib/packets/execute.js
+++ b/lib/packets/execute.js
@@ -21,7 +21,73 @@ function leftPad(num, value) {
     return (pad + s).slice(-num);
 }
 
+function toMysqlDateTime(date) {
+  return [date.getFullYear(), date.getMonth() + 1, date.getDate()].join('-') +  ' ' +
+    [date.getHours(), date.getMinutes(), date.getSeconds()].join(':') + '.' +
+    leftPad(3, date.getMilliseconds());
+}
+
+function isJSON(value) {
+  return Array.isArray(value) ||
+    value.constructor === Object ||
+    (typeof value.toJSON === 'function' && !Buffer.isBuffer(value));
+}
+
+/**
+ * Converts a value to an object describing type, String/Buffer representation and length
+ * @param {*} value
+ */
+function toParameter(value, encoding) {
+  var type = Types.VAR_STRING;
+  var length;
+  var fixed = false;
+  if (value !== null) {
+    switch (typeof value) {
+      case 'number':
+        type = Types.DOUBLE;
+        fixed = true;
+        var bufferValue = Buffer.allocUnsafe(8);
+        bufferValue.writeDoubleLE(value, 0);
+        value = bufferValue;
+        break;
+
+      case 'boolean':
+        type = Types.TINY;
+        fixed = true;
+        var bufferValue = Buffer.allocUnsafe(1);
+        bufferValue.writeInt8(value | 0, 0);
+        value = bufferValue;
+        break;
+
+      case 'object':
+        if (Object.prototype.toString.call(value) == '[object Date]') {
+          value = toMysqlDateTime(value);
+        } else if (isJSON(value)) {
+          type = Types.JSON;
+          value = JSON.stringify(value);
+        }
+        break;
+    }
+  } else {
+    type = Types.NULL;
+    value = '';
+  }
+  if (fixed) {
+    length = value.length;
+  } else {
+    if (Buffer.isBuffer(value)) {
+      length = Packet.lengthCodedNumberLength(value.length) + value.length;
+    } else {
+      value = value.toString();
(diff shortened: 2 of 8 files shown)


gazoakley:
Node 8 build on Travis failed (but passed on AppVeyor). Looks like a random failure with npm. Closing (will reopen to retrigger Travis build).

gazoakley:
Reopening to retrigger Travis build.

sidorares:
thanks @gazoakley ! I'll try to fix build issue

btw date has it's own serialization format, not sure though if there is any benefit over sending tham as a string ( you can see how it's serialised in `packet.readDateTime` - https://github.com/sidorares/node-mysql2/blob/513451fe632aaffe886618db1face0776485add8/lib/packets/packet.js#L230
Haven't started reviewing carefully yet, but you set type for Date to be string, right?

gazoakley:
Hi @sidorares - thanks for a quick response.

I didn't spot the date serialization that you suggested although to me I think it'd be preferable for Date objects to round trip back to Date if possible (I don't know if that would be a breaking change?). I'll give it a try.

sidorares:
>  it'd be preferable for Date objects to round trip back to Date if possible

not sure I understand. This is client -> mysql communication, not casting results mysql -> js . Priorto your pr all parameters were sent as strings ( and marked as having string type ). On mysql side server would do casting like "prepared statement needs number here, we received "1111" so it's probably a number 111).

gazoakley:
Sequelize has its own tests that run against drivers, and it expects numbers to round trip with the same type. For example:

SELECT ? AS foo (where foo is a number parameter)

Without this PR the tests will fail, since Sequelize sends a number as the input parameter and gets back a string in the result. Granted it's a bit unusual to do this, but the test passes for other drivers (it doesn't currently support bind params for MySQL - https://github.com/sequelize/sequelize/pull/8861 hopes to resolve that)

sidorares:
oh ok I see now.

sidorares:
what do you think is ideal behaviour when parameter is undefined? Return error immediately, convert to string "undefined", treat as NULL ?

gazoakley:
Hi @sidorares 

In response to your previous post, I'd argue for throwing an error since it doesn't map to a MySQL data type - if you want to provide a parameter as MySQL NULL then use JS null. One thing I struggled with is catching the error properly since I think it's being thrown inside a callback - I had to use `process.on('uncaughtException', ...)` to write a test to check it works.

I've implemented sending Date objects as DATETIME - other tests are still passing. 

In response to your follow up, would you like to see an option like `timezone`? I'm guessing in these changes it would convert to UTC/specified offset (but elsewhere in the code it would need to convert from that back to local time). Should it go in a separate PR?

gazoakley:
It looks like the previous behaviour didn't do anything with the timezone either? https://github.com/sidorares/node-mysql2/pull/705/files#diff-3b8f681f01aefd0299dedcf941cd7a3dL46

sidorares:
> In response to your previous post, I'd argue for throwing an error since it doesn't map to a MySQL data type - if you want to provide a parameter as MySQL NULL then use JS null. One thing I struggled with is catching the error properly since I think it's being thrown inside a callback - I had to use process.on('uncaughtException', ...) to write a test to check it works.

yes, this is how I think about this as well - most `undefined`s here in my experience are typos or similar silly errors, and it's better to throw early ( preferably before loosing sync stack ). This is however different from `.query()` casting where `undefined` is treated as NULL. I'd vote for "throw early"

>  would you like to see an option like timezone? I'm guessing in these changes it would convert to UTC/specified offset (but elsewhere in the code it would need to convert from that back to local time). Should it go in a separate PR?

yes, definitely separate pr, just wanted to flag that because it's often cause of a problem for anything related with mysql DATETIME type

sidorares:
sorry, I'm extremely busy this week, I'll try to find time to help to land this PR

sidorares:
> It looks like the previous behaviour didn't do anything with the timezone either? 

yes, that's what I'm saying by " does not behave well in communicating user intentions about timezone" - #642 #15 #262 #633

gazoakley:
> sorry, I'm extremely busy this week, I'll try to find time to help to land this PR

No worries - thanks for your help so far! 👍

sidorares:
Hey @gazoakley sorry for silence.
I'll put few notes here as low priority and maybe tackle on them later, but want to merge this as is anyway, don't want this PR to sit any longer

- move `toMysqlDateBuffer` to packet.js - this is where all low level serialising/deserializing lives. Commands/packets are for higher level protocol logic
- JSON.stringify can throw if there are circular references. Maybe we need to handle this (maybe not)
- add handling of `undefined` (maybe earlier when command is added to make stack trace better)

vlasky:
> add handling of undefined (maybe earlier when command is added to make stack trace better)

Yes please do this post haste. This has been an outstanding issue for a long time (see https://github.com/sidorares/node-mysql2/pull/516 and https://github.com/sidorares/node-mysql2/pull/480).

gazoakley:
Hi @sidorares - thanks for getting this merged. I've tackled some of the issues identified in #718

PR #706, opened 2018-01-12
Use takis again

sushantdhiman:
I noticed some builds are hanging forever in [this pr](https://github.com/sidorares/node-mysql2/pull/705). Reverted back to using takis again, which I think is more reliable.

If there is any reason for not using takis, please let me know

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     env: LINT=0
   - node_js: "6.11"
     env: LINT=0
-  - node_js: "8.0"
+  - node_js: "8.9"
     env: LINT=1
 
 cache:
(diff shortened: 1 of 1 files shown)


sidorares:
takis is at tcp level, while wait-up script allows to wait until you can actually connect to database. Without that some test randomly fail if they start after docker is up but before server finished initialisation

sidorares:
not sure why it was hanging, I did re-run once and it worked

dougwilson:
I have been having Travis CI jobs hanging in various projects randomly for the past week. It may have been just something wrong with the Travis job itself (maybe some kind of infra problem?). I have to keep restarting jobs every so often because they'll hang forever.

sushantdhiman:
Good point, we should use both `takis` and `wait-up`

sidorares:
why both then?

sushantdhiman:
Because build keep pinging mysql server even if its not listening

PR #716, opened 2018-01-26
Treat undefined as nulls for execute - Fixes #715

### LINKED ISSUE #715: Prepared Statement Executes with undefined values
If you call 
```javascript
const result = await connection.execute("SELECT foo FROM bar where baz = ?", [undefined]);
```

then the Statement Execute will never finish, and forever hang.

The fix is pretty simple, in lib/packets/execute.js needs to use double equals instead of triple equals checks.

aikar:
See #715

### DIFF
--- a/lib/packets/execute.js
+++ b/lib/packets/execute.js
@@ -37,7 +37,7 @@ Execute.prototype.toPacket = function() {
     length += 1; // new-params-bound-flag
     length += 2 * this.parameters.length; // type byte for each parameter if new-params-bound-flag is set
     for (i = 0; i < this.parameters.length; i++) {
-      if (this.parameters[i] !== null) {
+      if (this.parameters[i] != null) {
         if (
           Object.prototype.toString.call(this.parameters[i]) == '[object Date]'
         ) {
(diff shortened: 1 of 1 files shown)


sidorares:
I'm trying to balance "correctness" "least surprise" and "if it's a programmer error, give feedback as early as possible".

See https://github.com/sidorares/node-mysql2/pull/705#issuecomment-356505493

To me better treat `undefined` as "you probably have a typo or forget parameter?" error. ( Depends of course on how often there is legit use case to set value to NULL when you actually send undefined ) There is no mysql `undefined` type, just NULL

What's your personal opinion @aikar ? (not saying that current behaviour is good, we need to fix it. Just do not want to replace crash with hiding issue)

aikar:
@sidorares I battled with that decision when working on it tonight. At first I was going to just add an error if undefined is found before I call this API, but figured if null works, should just treat it as null.

But, I can also get behind the idea that undefined is more likely than not an application code bug, and it could help developers discover bugs that are otherwise able to be more sneaky (especially for inserts!)

So, I can get behind making it an error instead, but if that other PR is nearing landing, we can just close this then, as it def looks good!

aikar:
Personally, Id say we could pull this for now, then iterate when a decision is made on treating undefines in that other PR

vlasky:
We need a solution that allows the programmer to identify the query (and arguments) responsible for passing the undefined parameters to execute, as this is almost certainly due to a bug.

See previous attempts https://github.com/sidorares/node-mysql2/pull/516 and https://github.com/sidorares/node-mysql2/pull/480.

aikar:
@sidorares see https://github.com/sidorares/node-mysql2/pull/689/files - may be a good basis for this.

Going to close this PR since #705 is merged, and i think most people do favor alerting you to an error vs treating as null.

PR #718, opened 2018-01-30
feat(execute): Tidy up source param types

gazoakley:
This tidies up #705 : 

* Moves date serialization to packet.js
* Uses functions in packet.js instead of creating buffers
* Catches use of undefined params earlier (test updated)

Tests are passing locally

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -715,6 +715,18 @@ Connection.prototype.execute = function execute(sql, values, cb) {
   }
   this._resolveNamedPlaceholders(options);
 
+  // check for values containing undefined
+  if (options.values) {
+    options.values.forEach(function(val) {
+      if (val === undefined) {
+        throw new TypeError('Bind parameters must not contain undefined. To pass SQL NULL specify JS null');
+      }
+      if (typeof val === 'function') {
+        throw new TypeError('Bind parameters must not contain function(s). To pass the body of a function as a string call .toString() first')
+      }
+    });
+  }
+
   var executeCommand = new Commands.Execute(options, cb);
   var prepareCommand = new Commands.Prepare(options, function(err, stmt) {
     if (err) {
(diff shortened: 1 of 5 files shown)


lib/connection.js
```diff
@@ -715,6 +715,15 @@ Connection.prototype.execute = function execute(sql, values, cb) {
   }
   this._resolveNamedPlaceholders(options);
 
+  // check for values containing undefined
+  if (options.values) {
+    options.values.forEach(function(val) {
+      if (val === undefined) {
+        throw new TypeError('Bind parameters must not contain undefined');
```
sidorares: Maybe expand a bit this message? Add something like " If you wanted to pass NULL value you have to use Javascript null explicitly"

lib/packets/execute.js
```diff
@@ -17,74 +17,59 @@ function isJSON(value) {
...
         } else if (isJSON(value)) {
-          type = Types.JSON;
           value = JSON.stringify(value);
+          type = Types.JSON;
+        } else if (Buffer.isBuffer(value)) {
+          length = Packet.lengthCodedNumberLength(value.length) + value.length;
+          writer = Packet.prototype.writeLengthCodedBuffer
         }
         break;
+
```
sidorares: maybe add checks for other types to help prevent silly errors? for example, if I accidentally put callback as a parameter
I guess it's better to throw when function is passed and require explicit fn.toString() if that was actually your intention

gazoakley: I've added a specific check for `function` params here: https://github.com/sidorares/node-mysql2/pull/718/files#diff-259dec4414a400a6e895f16ff1d0ca3bR724 - should preserve the stack trace better (I've added a test for it too)

lib/connection.js
gazoakley: Updated to read 'Bind parameters must not contain undefined. To pass SQL NULL specify JS null'

sidorares:
nice! Thanks @gazoakley !

gazoakley:
Thanks @sidorares 👍 - any idea when the next release might be?

sidorares:
@gazoakley in ~5 hours

sidorares:
@gazoakley published v1.5.2

gazoakley:
@sidorares Thanks :+1:

PR #720, opened 2018-01-30
Improve performance of removing connections from pools

aikar:
Use the built in splice method to avoid reordering the queue and improve performance.

Originally was removed due to https://github.com/sidorares/node-mysql2/issues/226 - but that was when it used 'double-ended-queue', and we now use denque, which does include a splice method.


sidorares:
hm, it errors because there is no `indexOf` in denque

aikar:
Ah yeah just saw that myself, forgot it wasn't an array. Will fix

sidorares:
LGTM, thanks!

PR #723, opened 2018-02-02
Store Compiled Packet Parsers in a global cache - Resolves #722

### LINKED ISSUE #722: Memory leak due to packet parsers
I've noticed a memory leak in our processes, and after days of debugging, I've finally narrowed it down to the packet parsers.

The issue stems with the nature that the parser cache is stored on the connection object.

Most people are ideally using connection pools, and due to the nature that connections aren't closed (until #651 at least) usually, this problem doesn't really make itself apparent.

But once we start cycling connections, the leak starts growing... and slowly.

Here's the fun part, it doesn't show up in the heap dump fully!
A process can show a much larger memory footprint than the Heap Dump says its around.

The V8 engine is creating Compiled Code for these functions, and I assume that V8 does not clear this compiled code cache. 
Some of the memory is represented in the heap dump, but a large chunk isn't.

I orchestrated my environment with my #651 PR to auto time out connections every minute, opening 20 connections and closing them rapidly.

My process leaked upwards to 1GB of memory in a few hours. Comparing Heap Snapshots, all of the memory went mostly to strings, copies of the same BinaryRow parsers over and over, and compiled code.
But the size reported in Strings being 200MB~, the real RSS size was 1GB.

I am about to submit a PR to resolve this issue.

aikar:
See #722 for details

When a connection is closed, any compiled parser created for that will
be lost, as the connection is removed.

However, Node/V8 Internals appear to be holding onto any created function.

This can result in memory leaking based on connection cycling and new
parsers being created again.

This change creates a global cache store since everything around the generated
code is not connection specific.

This also completes a TODO to use an LRU cache, although I set a high value
as to avoid running into the leak if the LRU is too small.

An API is provided for anyone who does need to control the max value, as
well as an API to clear the parser cache.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@ var SqlString = require('sqlstring');
 
 var Connection = require('./lib/connection.js');
 var ConnectionConfig = require('./lib/connection_config.js');
+var parserCache = require("./lib/parsers/parser_cache");
 
 module.exports.createConnection = function(opts) {
   return new Connection({ config: new ConnectionConfig(opts) });
@@ -61,3 +62,11 @@ exports.__defineGetter__('Charsets', function() {
 exports.__defineGetter__('CharsetToEncoding', function() {
   return require('./lib/constants/charset_encodings.js');
 });
+
+exports.setMaxParserCache = function (max) {
+  parserCache.setMaxCache(max);
+};
+
+exports.clearParserCache = function () {
+  parserCache.clearCache();
+};
--- a/lib/commands/execute.js
+++ b/lib/commands/execute.js
@@ -6,7 +6,7 @@ var Packets = require('../packets/index.js');
 
 var objectAssign = require('object-assign');
 
-var compileParser = require('../compile_binary_parser.js');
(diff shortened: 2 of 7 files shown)


sidorares:
Looking good! I'll try to review carefully later (and I'm away next 3 days) as it's quite core part

Can you think of some tests that can cover cache?

Also would be good what was the cause of original leak

aikar:
It would be too difficult to test the original case, the speed of the leak varied and could take hours to be noticed. Nor can you easily identify it in code since its leaking more natively.

As for testing in general of the change, this was mainly moving code around / changing the where its stored.
the behavior really is sort of the same as it was before when 'caching' is considered. Is it in cache? yes use it, else make and cache it. Except its now 1 global bucket instead of 2 buckets per connection.

I don't think we really need to bother implementing tests to check LRU behavior heh, the LRU library tests that.

Nor do I have the time to even try right now, as spending days on this issue was already undesired to my boss as it was... He's going to be happy to hear it's resolved.

Just to add an update, process has been running for an hour now with 0 signs of leaking.

PR #732, opened 2018-02-16
Allow setting connection timezone and honor UTC For Date/DateTime

### LINKED ISSUE #262: Timezone is wrong on client when reading records from SQL server running in UTC time

aikar:
This resolves #262

if timezone is set, we change the timezone of the connection upon opening.

If timezone is set to utc and not using dateStrings, we create Date objects
using UTC mode instead.

Also discovered missing config keys that affect packet parsers not
being included in the cache key for packet parser cache, so added them.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -117,7 +117,18 @@ function Connection(opts) {
       }
       connection._handshakePacket = handshakeCommand.handshake;
       connection.threadId = handshakeCommand.handshake.connectionId;
-      connection.emit('connect', handshakeCommand.handshake);
+      if (connection.config.timezone && connection.config.timezone !== 'local') {
+        connection.query("SET time_zone = ?", [connection.config.timezone], function (err) {
+          if (err) {
+            connection.end();
+            connection.emit("error", err);
+          } else {
+            connection.emit('connect', handshakeCommand.handshake);
+          }
+        });
+      } else {
+        connection.emit('connect', handshakeCommand.handshake);
+      }
     });
     handshakeCommand.on('error', function(err) {
       connection._closing = true;
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -31,7 +31,7 @@ function ConnectionConfig(options) {
   this.debug = options.debug;
   this.trace = options.trace !== false;
   this.stringifyObjects = options.stringifyObjects || false;
-  this.timezone = options.timezone || 'local';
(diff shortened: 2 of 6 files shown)


sidorares:
looks good! Can you add tests for time changes? Not sure what's better way to test, real mysql server or mock server, but we need to test at least one happy path

aikar:
I'll see if i can over weekend, hopefully won't forget.... was a late night in the office.

jasonlewicki:
Would love to see this pushed along.

I am trying to SELECT TIMESTAMP fields from a MySQL server set to UTC and they are converted into the future by my local machine's TZ offset.

aikar:
@sidorares Would you be able to add some test for it so we can get this merged? I've been way too busy to find the time to setup testing for this.

sidorares:
@aikar yes, I can work on tests myself. Hope to clean up high priority issues over weekend

aikar:
@jasonlewicki make sure you set timezone: 'utc' in your pool config. This PR will not auto detect timezone settings.

We should follow up with another PR to add (configurable) automatic timezone detection.

kibertoad:
@sidorares Need any help with this PR?

sidorares:
@kibertoad @aikar this is fairly big change, we need at least add some integration tests for this

sidorares:
also a bit on the fence about "local" timezone and automatic "SET time_zone = ?" query. If we end up adding this it need at least good documentation, otherwise it will end up being dead code no one uses

sidorares:
hey @abhidp sorry I was not helpful enough to get this work into master 
Closing in favour of #996

PR #736, opened 2018-02-28
feat(pool): Add Minimum connections, and auto close extra connections

aikar:
This adds a new configuration for desired minimum connections, to
complement the connection limit.

If the autoOpenConnections is enabled (default true), the pool will
immediately open the desired minimum number of connections.

Any connection opened above the minimum, who sits unused for a time
specified in idleTimeout, will be automatically closed.

This allows you to set up a minimum expectation of standard load for
connections to your database, but also allow you to configure a
higher maximum to handle unexpected burst traffic, and be able to
close them automatically when the burst subsides.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -90,6 +90,7 @@ function Connection(opts) {
   });
 
   this.stream.on('end', function() {
+    this._closed = true;
     // we need to set this flag everywhere where we want connection to close
     if (connection._closing) {
       return;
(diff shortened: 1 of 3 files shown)


aikar:
Github not letting me reopen the previous branch struck yet again (no github you will not change my workflow!)

Given that the memory leak bug is not an issue with node-mysql2, we should be good here now.

Since the last PR, I did discover (in relation to #732) that errors triggered during the initial connection such as setting the timezone, did not bubble back up correctly.

The logic of handling errors upon connection in the auto opened connections has been resolved, by bubbling the error to a pending connection.

I'm not sure what else we could do if theres an error with no pending connections, given that there is no error logging system officially.

I think my next PR will be to configure a general logger interface like I have here: https://github.com/aikar/node-mysql2/commit/b7d624c4fac781e32cdf4b577c98d7040af126fa

But let there be debug/info/error levels.

aikar:
@sidorares confirming that the memory leak fear is completely gone and was related to my own use of URL.

I've been running this for many days now with stable memory.

PR #741, opened 2018-03-01
Fixes execute example

exalted:
`execute` function ought to be called on the `connection` rather than `mysql`.


sidorares:
Thanks! Looks like `unprepare` also must be called on connection, can you fix that as well?

exalted:
@sidorares done!

PR #748, opened 2018-03-10
fix(packet): remove `noAssert`

BridgeAR:
The `noAssert` argument is no longer supported by Node.js and all
input will be validated even if it is passed through. This removes
the argument therefore.

Refs: https://github.com/nodejs/node/pull/18395

Off-topic: the git hooks in use seem to interfere a lot with the contributors will.
It is for example not possible to commit something with `git commit -p` because even if you only apply a few patches, all code is going to be commited anyway. That is not very user friendly :-/


sidorares:
Thanks!

Yeah, it's a bit annoying, I'll probably remove them and only check everything on CI

Issue #761, opened 2018-04-02
Passing Empty String as Value Directly Throws


anliting:
Code:

```js
import mysql2 from 'mysql2'
mysql2.createPool({
    host:'localhost',
    user:'anliting',
    password:'',
}).query('select ?','',(err,res)=>{
    console.log(err)
})
```

Expected: No error.

Actual:
```
{ Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '?' at line 1
    at Packet.asError (/home/anliting/node_modules/mysql2/lib/packets/packet.js:716:13)
    at Query.Command.execute (/home/anliting/node_modules/mysql2/lib/commands/command.js:28:22)
    at PoolConnection.Connection.handlePacket (/home/anliting/node_modules/mysql2/lib/connection.js:502:28)
    at PacketParser.onPacket (/home/anliting/node_modules/mysql2/lib/connection.js:81:16)
    at PacketParser.executeStart (/home/anliting/node_modules/mysql2/lib/packet_parser.js:77:14)
    at Socket.<anonymous> (/home/anliting/node_modules/mysql2/lib/connection.js:89:29)
    at Socket.emit (events.js:180:13)
    at addChunk (_stream_readable.js:274:12)
    at readableAddChunk (_stream_readable.js:261:11)
    at Socket.Readable.push (_stream_readable.js:218:10)
  code: 'ER_PARSE_ERROR',
  errno: 1064,
  sqlState: '42000',
  sqlMessage: 'You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near \'?\' at line 1' }
```

SpencerTuft:
The second parameter in the query function should be an array of values. Try: `import mysql2 from 'mysql2'
mysql2.createPool({
    host:'localhost',
    user:'anliting',
    password:'',
}).query('select ?',[''],(err,res)=>{
    console.log(err)
})`

PR #762, opened 2018-04-02
Fix for #761

anliting:


sidorares:
Thanks! Would you mind adding test for this?

anliting:
Okay, I'll try.

anliting:
Maybe we'll wait for this (https://github.com/mysqljs/mysql/issues/1978) if it's closing in a few days.

anliting:
Tried; it seems that I have to investigate and setup too many things to code and run the test.

Maybe you can do it for me? The test should be "trying perform `pool.query` (`pool.format` has been good) with directly passed falsy values except for `undefined` and `null`". It should be like this:

```js
import assert from 'assert'
import mysql2 from 'mysql2'
let pool=mysql2.createPool({
    host:'localhost',
    user:'anliting',
    password:'',
})
for(let e of [false,0,'',NaN])
    pool.query('select ?',e,err=>assert(!err))
```

Besides, is this (https://github.com/sidorares/node-mysql2/blob/master/test/unit/test-Pool.js#L29) a bug? L33 and L39 also.

sidorares:
no problems, I can add test to your pr

what do you think is wrong with L39?

```js
 const params = ['table name', 'thing'];
 assert.equal(
      pool.format('SELECT a FROM ?? WHERE b = ?', params),
      "SELECT a FROM `table name` WHERE b = 'thing'"
    );
```

anliting:
We should test `promisePool` there, not `pool`, isn't it?

sidorares:
I don't think so, this is unit test for non-promise version of pool

sidorares:
this is promise pool test: https://github.com/sidorares/node-mysql2/blob/5c5aa3aed7e772e599dbefdafb04405672901808/test/integration/promise-wrappers/test-promise-wrappers.js#L250-L268

anliting:
But L27-L43 is almost the same as L8-L24, and promisePool created on L26 is not being referenced anywhere.

sidorares:
oh, you are right. It's probably a bug

kibertoad:
@sidorares What is left for this PR to be finished? Do you need any help?

sidorares:
@kibertoad I think it can be merged. Can you have a look and review it as well @kibertoad ? Any concerns from you?

kibertoad:
@sidorares 
1) I'm not too comfortable with having no tests for the change;
2) Formatting seems to be off, no spaces;
3) Does specific "=== undefined" (as opposed "not truthy") make sense here?

anliting:
@kibertoad 
For the "3. Does ...": I think it make sense here. `e0===undefined?e1:e0` have different semantic comparing to `e0||e1`.

Issue #766, opened 2018-04-11
Throw an error when a connection ends with a RST,ACK packet?


elemount:
Hi,

I'm using Azure MySQL database.

According to MySQL docs, COM_QUIT command should end with "either a connection close or a OK_Packet".

It can be reproduced when you run this example code against an Azure MySQL instance.
```js
var mysql      = require('mysql2');
var connection = mysql.createConnection({ config });

connection.connect();

connection.query('SELECT 1 + 1 AS solution', function (error, results, fields) {
  if (error) throw error;
  console.log('The solution is: ', results[0].solution);
});

connection.end();
```

Throws `ECONNRESET` error.

Should this library throw an error when a connection end with a RST with the ACK flag set?
Can we detect this?

I see mysqljs already fixed it at commit https://github.com/mysqljs/mysql/commit/9168ced54148721c8872ddab583116f00b4bc4f8 .
But we prefer using mysql2 now and it block me. Could mysql2 fix it either?

sidorares:
probably fix should be here - https://github.com/sidorares/node-mysql2/blob/5c5aa3aed7e772e599dbefdafb04405672901808/lib/connection.js#L123

if already `_closing`, ignore error ( same way `.end()` is expected when this flag set on line 94 )

sidorares:
would you be able to apply this fix and check @elemount ?

elemount:
@sidorares , I'm not sure if it is already in the newest release. 
I can repo it by following steps.
1. Create a folder and cd it.
2. 
```
npm install mysql2
npm install knex
``` 
3. Save this scripts as file `test.js`
```js
const knex = require('knex');

async function run() {
    const db = knex({
        client: 'mysql2',
        connection: {
            user: "elemount@elemount-test",
            password: "Testmysql2",
            host: "elemount-test.mysql.database.azure.com",
            port: 3306,
            database: "test",
        }
    });

    await db.schema.dropTableIfExists('test');
    await db.schema.createTable('test', table => {
        table.increments('id').primary();
        table.integer('key1');
    });
    db.destroy();
}

run();
```
4. `node test.js`

And then I will got 
```
events.js:165
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at TCP.onread (net.js:602:25)
Emitted 'error' event at:
    at Connection._notifyError (D:\Play\repo\node_modules\mysql2\lib\connection.js:229:16)
[...]

PR #768, opened 2018-04-12
Do not throw an error when a connection ends with a RST,ACK packet

elemount:
Refer to issue #766


sidorares:
Thanks! I'll see what else is in the backlog and try publish soon

PR #772, opened 2018-04-15
Supports a Japanese charset `ujis`

cognitom:
Hello. This PR makes it possible to connect to DBs in a Japansese charset `ujis`.
I've already tested that it works in my server manually, but I'm still thinking about how to make a test for it... Pls, let me know, if anyone knows how to. Thank you!

| current | this PR |
|:--|:--|
| ujis --> ujis | ujis --> **eucjp** |
| eucjpms --> eucjp | eucjpms --> eucjp |

### Points and rationales

- `ujis` in MySQL is equivalent to `eucjp` in Linux, so `ujis` should be naturally mapped to `eucjp`. (note: `ujis` is an old-fashioned way to say `eucjp`)
- `eucjpms` is another charset, and it's a slightly extended vesion of `eucjp`.
- [iconv-lite extends eucjp](https://github.com/ashtuchkin/iconv-lite/blob/master/test/dbcs-test.js) from the standard one, and it looks close to `eucjpms`.
- Historically, `eucjpms` has been mapped to `eucjp` in `node-mysql2`, and this was practically almost correct because of the reason above.
- So keep this mapping from `eucjpms` to `eucjp`, too.


sidorares:
> Historically, eucjpms has been mapped to eucjp in node-mysql2, and this was practically almost correct because of the reason above.

Do you know what are the exceptions?

cognitom:
Thanks for your quick response!
Yes, there's some exceptions. For example:
- [0xF9 - 0xFC](https://github.com/ashtuchkin/iconv-lite/blob/master/encodings/tables/eucjp.json#L93-L96) is in iconv-lite's `eucjp` but not in `eucjpms`
- `eucjpms` has a different code table for the characters above

`eucjp` has a few varieties like `eucjpms` and `CP51932`. The 90% of characters are the same in each charset, but there're differences on some Kanji letters which are hardly used.

- `eucjp`
- `eucjpms`: supported by MySQL (almost only ?)
- `CP51932`: supported by Internet Explorer and other Windows applications

And, `ivonv-lite` looks covering `CP51932`. So, correctly speaking, it's better to do like below:

| current | strict |
|:--|:--|
| ujis --> ujis | ujis --> **eucjp** |
| eucjpms --> eucjp | eucjpms --> **eucjpms** |

**Note**: `iconv-lite` has no support for `eucjpms`, so it will be an error when the MySQL server's charset is `eucjpms`.

cognitom:
I've found the iconv-lite auther's comment [here](https://github.com/ashtuchkin/iconv-lite/issues/145#issuecomment-275902481). @ashtuchkin says:
> iconv-lite tries its best to mirror the [WHATWG Encoding Standard](https://encoding.spec.whatwg.org/)

That's another variety of `eucjp`. Hmm, t's complicated...
I may recommend `strict` pattern above, but it needs a breaking change. How do you think?

sidorares:
> I may recommend strict pattern above, but it needs a breaking change. How do you think?

you mean doing eucjpms --> eucjpms part as well? I'm ok for this, we'll just bump major version to communicate potential incompatibility but I doubt anyone would be affected

sidorares:
While you at it, if you have a chance maybe you can check how good TextDecoder / Encoder is in last stable node https://nodejs.org/api/util.html#util_class_util_textdecoder ?

cognitom:
Sure. Let me see....
`TextDecoder` needs some build options here:
https://nodejs.org/api/intl.html#intl_options_for_building_node_js

They are not enabled in default environments. We can check it:

```javascript
const hasICU = typeof Intl === 'object'
```

In case the user environment is supported, it seems good idea to use it:

- Build with ICU: `TextDecoder`
- Build without ICU: `iconv-lite`

sidorares:
yes, I still assume `iconv-lite` as a fallback but if there is currently benefit in using TextDecoder when it's available ( more prformant, better reflect standard ?- need to actually test ) maybe we should use it

sidorares:
I have default node install on my work windows laptop and TextDecoder is there

cognitom:
> I have default node install on my work windows laptop and TextDecoder is there

@sidorares that may only support `utf8`.

sidorares:
yeah, you are right:

```
> const d = new util.TextDecoder('shift_jis')
Error [ERR_ENCODING_NOT_SUPPORTED]: The "shift_jis" encoding is not supported
    at new TextDecoder (internal/encoding.js:375:15)
    at repl:1:11
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:440:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:279:10)
> (node:15676) ExperimentalWarning: The WHATWG Encoding Standard implementation is an experimental API. It should not yet be used in production applications.
```

in that case maybe not worth efforts yet, iconv-lite is good enough

sidorares:
I'm happy to merge this as is, will bump major version before releasing some time soon
Anything else you might forgot or it's all done from your side?

cognitom:
Nope. Done. Go ahead pls 😄

sidorares:
Thanks for your work @cognitom !

PR #783, opened 2018-05-21
Emit pool connection acquire and release events

dvejmz:
At the moment mysql2 does not emit any events when a connection is acquired or released back into the pool, which is something mysql-node does support. Adding these helps instrument database usage patterns more comprehensively,  allowing users to better understand how their applications make use of the library, whether they're leaking connections or just using more than they need and help them diagnose connection limit errors and suchlike.


sidorares:
thanks @dvejmz !

PR #790, opened 2018-05-30
Add Node 10 to CI builds. Bump versions for older Node

kibertoad:


kibertoad:
@sidorares This uses latest compatible versions - appveyor doesn't have newest builds for everything altogether, and npm 6.x (included starting with Node 10.3.0) seems to break the pipeline with "ERR! Invalid Version: https://registry.npmjs.org/cardinal/-/cardinal-1.0.0.tgz" error.

sidorares:
thanks! Let's bump to 10.3 a bit later when appveyor catches up

I'm OK to drop node 4 altogether. Maybe also remove readable-stream and safe-buffer shims

PR #791, opened 2018-05-31
Drop support for Node 4, remove shims

kibertoad:


kibertoad:
@sidorares As requested in #790

PR #803, opened 2018-07-10
Update pool-test.js

ISNIT0:
`.end()` is now deprecated, converted to `.release()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,8 +135,16 @@ const pool = mysql.createPool({
 ```
 The pool does not create all connections upfront but creates them on demand until the connection limit is reached.
 
-You can use the pool in the same way as connections (using `pool.query()` and `pool.execute()`, but there is also the possibility of manually acquiring a connection from the pool and returning it later:
+You can use the pool in the same way as connections (using `pool.query()` and `pool.execute()`):
+```js
+// For pool initialization, see above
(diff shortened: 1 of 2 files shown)


sidorares:
looks like `.release()` should not be there as well
Maybe original intention in example was to show how to create pool, use it with `.query()` helper and then close _pool_

`.query()` calls release for you immediately after query is finished execution - see code here https://github.com/sidorares/node-mysql2/blob/224c74de8184f78c540f343c432c0f37986c8aea/lib/pool.js#L162

ISNIT0:
Ahh, didn't see that in any documentation. I'll change this PR to reflect this - definitely worth having *at least* valid documentation or examples :)

sidorares:
I'm really sorry, I thought it's `pool.query()` and in that example it's actually pool.getConnection() + connection.query() - in that case `.release()` is required and your initial change was in fact correct!

ISNIT0:
Okay, so in this case should there be an example of `pool.query()`? I had no idea that existed!

sidorares:
yes, won't hurt adding example here

Since I'm trying to mirror mysqljs/mysql API where possible best source of documentation is there, for example pool.query part is https://github.com/mysqljs/mysql/#pooling-connections

sidorares:
thanks!

Issue #809, opened 2018-07-23
A pool should be able to hand out both promise/non-promise wrapped connections


mdierolf:
In several of my applications built using node-mysql2, I maintain a pool of connections in one module:

### /db.js

```
import mysql from 'mysql2';
export const pool = mysql.createPool(mysqlURI);
```

In another module, I then use the pool to retrieve a connection, for operating on the connection in streaming mode, so I do this:

### /module1.js
```
let con = pool.getConnection()
con.query("SELECT ...")
.on('result', ...)
```
In another module, I wish to use the pool to retrieve a connection, for operating on the connection in promise mode, so i'd like to do this:

### /module2.js
```
let con = pool.getConnectionPromise()
con.query("SELECT ...")
.then( .... )
```

Unfortunately, because the pool only allows you to retrieve either a promise or non-promise wrapped connection, I need to create 2 different pools:

### /db.js
```
import mysql from 'mysql2';
export const pool = mysql.createPool(mysqlURI);
export const poolPromise = mysql.createPoolPromise(mysqlURI);
```

It would be helpful if the non-promise pool allowed you to call getConnectionPromise, or queryPromise. Not all code is suited for use with promises, and maintaining two pools isn't ideal.

mdierolf:
It occurred to me after posting this, that perhaps the pool should be left alone, and any connection could be altered to allow "upgrading" it to a promise. For example:

```
// Get a normal connection
let con = pool.getConnection();
// Do some non-promisy stuff with the connection
streamAndMangleData(con);

// Now we need a promise wrapped connection for some reason
let pcon = con.promise();
await pcon.query("SELECT stuff");

// Now release con back to pool
con.release();
```

The model of calling .promise() on a non-promise API is used by aws-sdk.

sidorares:
this is actually easy to implement - see https://github.com/sidorares/node-mysql2/blob/18619673910c911a8977a9ac02d898a94c27d156/promise.js#L54

once PromiseConnection is exported the code in connection.js would look like

```js
Connection.prototype.promise = function() {
   const PromiseConnection = require('../promise').PromiseConnection;
   return new PromiseConnection(this);
}
```

do you want to volunteer to implement this @mdierolf ? Ideally with tests covering new functionality and few new lines in documentation

Also same can be applied to pool

```js
const pool = mysql.createPool(opts);
// ...
const promisifiedPool = pool.promise();
// ...
await promisifiedPool.getConnection();
```

[...]

PR #810, opened 2018-07-24
Added a .promise() function to Pool, Connection, PoolConnection

mdierolf:
This adds a .promise() function to Pool, Connection, and PoolConnection, which allows you to wrap an existing non-promise instance of any of these objects with a promise wrapper.

See issue https://github.com/sidorares/node-mysql2/issues/809

Code is not well tested yet. I wrote a unit test for Pool which passes but not sure how to test a Connection.


sidorares:
Looks good, thanks!

PR #811, opened 2018-07-25
Added .transaction(options ?,asyncFunction) to PromisePool & PromiseConnection

mdierolf:
This PR adds a .withConnection(asyncFunction) function to PromisePool, which allows the user to borrow a connection from the connection pool, run as many queries as they like, and then it will automatically release the connection back to the pool when the asyncFunction(con) they pass in either resolves or rejects.

Solves a garbage disposal problem when using promise pools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,19 +220,22 @@ async function main() {
   );
 ```
 
-When using a promise pool, MySQL2 also exposes a .withConnection(asyncFunction) function for automatically disposing of the connection after you are finished working with it:
+### Using managed transaction
+
+When using a promise pool, MySQL2 offers a .transaction() function to allow you to execute code within a transaction, with pre-defined behavior to roll back the transaction on error, and which automatically releases the connection after you are finished working with it:
 ```js
 async function main() {
   // get the client
   const mysql = require('mysql2/promise');
   // create the pool
   const pool = mysql.createPool({host:'localhost', user: 'root', database: 'test'});
-  // using the pool, execute a query
-  pool.withConnection(async function(con) {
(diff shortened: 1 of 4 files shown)


sidorares:
let me think about this :) I'm trying to resist adding more functionality to pool but it actually looks useful

Couple of notes:
result of the function should be passed outside when inner promise resolved:

```js
   const xyz = await promise.withConnection(async conn => {
     await conn.beginTransaction();
     const [rows] = await conn.query('insert .,,,');
     await conn.commit();
     return rows[0].someAutoIncrementField;
   });
```

also - maybe "withTransaction" that does .rollback() + .release in catch block?

Definitely needs unit tests to move forward

mdierolf:
Perhaps this would be better to model after the .transaction() call in the sequelize module.

http://docs.sequelizejs.com/manual/tutorial/transactions.html

A .transaction() with no isolation, in auto commit mode, with no rollback, would essentially have the same behavior as .withConnection(), except the developer would have to choose that mode explicitly. 

It’s probably a good thing to force the developer to choose an isolation level.

mdierolf:
I changed .withConnection() to .transaction(), and added behavior to run queries under different types of transactions.

I haven't tried this code yet, it is completely untested.

The previous behavior of .withConnection() is now available by calling .transaction({ autoCommit: true })

sidorares:
Could you extract ping() and `Query.then()` changes into separate pr? I'll merge those changes straight away. 

When new api methods I'd like PRs to stay open a bit longer with more eyes, once it's in it's really difficult to go back without potentially breaking someone's code

PR #813, opened 2018-07-25
Fix .ping(), error checking for Query.then(), const correctness

mdierolf:
This PR includes 3 changes:
- A fix for PromiseConnection.ping(), as it was not failing on errors
- An error is now thrown if .then() or .catch() is called on a non-promise Query
- promise.js was updated to consistently use const instead of var

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -33,7 +33,7 @@ function Query(options, callback) {
 util.inherits(Query, Command);
 
 Query.prototype.then = Query.prototype.catch = function() {
- var err = "Error: you have tried to call .then(), .catch(), or invoked await on the result of query that is not a promise, which is a programming error. Throwing an error.";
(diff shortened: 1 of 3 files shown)


lib/commands/query.js
```diff
@@ -32,6 +32,12 @@ function Query(options, callback) {
 }
 util.inherits(Query, Command);
 
+Query.prototype.then = Query.prototype.catch = function() {
+ var err = "Error: you have tried to call .then(), .catch(), or invoked await on the result of query that is not a promise, which is a programming error. Try calling con.promise().query(), or require('mysql2/promise') instead of 'mysql2' for a promise-compatible version of the query interface. To learn how to use async/await or Promises check out documentation at https://www.npmjs.com/package/mysql2#using-promise-wrapper, or the mysql2 documentation at https://github.com/sidorares/node-mysql2/tree/master/documentation/Promise-Wrapper.md";
+ console.log(err);
+ throw err;
```
sidorares: please use `new Error(msg)` instance instead of string, also no need for 'Error: ' prefix. 

For example if I type `new Error('test xyz')` in the repl here the result look like this:

```
> new Error('test xyz')
Error: test xyz
    at repl:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:468:10)
    at emitOne (events.js:121:20)
    at REPLServer.emit (events.js:211:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8)
>
```

sidorares:
Tests timing out, can you check locally what's happening?

mdierolf:
This should be ready to merge

PR #815, opened 2018-07-26
Added a uri: parameter to the connection config

mdierolf:
Frequently, I find myself using the following code:

```js
mysql.createPool(process.env.MYSQL_URI);
```

Unfortunately, when passing in a URI from an environment variable, it means that my application is unable to override any of the URI parameters, as createPool and createConnection only accept either URI String or Object parameters, not both at once.

This PR allows you to use both formats, by passing a uri: key into the options:
```js
mysql.createPool({
  uri: process.env.MYSQL_URI,
  multipleStatements: true
});
```

In the case of overlapping parameters, the Object parameters are currently given preference over the URI String parameters. I'm not sure if that is the correct preference.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -3,6 +3,49 @@ var ClientConstants = require('./constants/client');
 var Charsets = require('./constants/charsets');
 var SSLProfiles = null;
 
+const validOptions = {
+ authSwitchHandler: 1,
+ bigNumberStrings: 1,
+ charset: 1,
+ charsetNumber: 1,
+ compress: 1,
+ connectAttributes: 1,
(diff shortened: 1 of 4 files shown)


sidorares:
thanks, happy to add this.

one another thing that I wanted to add for long time re config is to disallow any keys we don't understand so any typos are detected early. Maybe should be different pr, just saying it would be good value for relatively small change

mdierolf:
Ok, i've added a check to ConnectionConfig which will print a warning to the console if you pass in any options that are unsupported.

I am wary about throwing an Error - if any users are currently passing in unsupported options, we would start crashing their applications, which I think is going a bit overboard.

Issue #820, opened 2018-07-30
Release planning information?


framp:
I'm a user of the library and I'd love to use some of the features that landed in master.
Is there a plan publish the changes from the last months on npm?

It would be great to know if there is a fixed release schedule or if it just slipped by.

Thanks for the library

sidorares:
Hi!

I'd love to publish everything already merged, maybe you can help? Need to prepare changelog and should be ready to bump

In terms of future plans:
- 8.0 caching_sha2_password ( similar to https://github.com/mysqljs/mysql/pull/1962 but hopefully smaller ) is in works and high priority for me
- new API for auth plugins ( current one is broken and going to be deprecated. This is also requirements for caching_sha2_password implementation ) - will be major version bump
- more stricter checks for number and type of arguments for `.execute()` ( prepare gives us number and type info which we can use to detect errors earlier - you only want to convert object or array to string if it's JSON parameter, if it's a number worth erroring earlier )
- https://github.com/sidorares/node-mysql2/pull/814 if there is no objection to new api, and maybe soft deprecate `.beginTransaction()` as suggested there

sidorares:
there is heaps of probably more valuable work that is not new features - review and close outdated prs/issues, improve docs that is not covered by mysqljs/mysql ( `createServer()` api, binlog etc )

framp:
Absolutely, PR incoming!

Thanks for the info, really appreciated
I had no idea it was a stopper, so thanks a lot also for the quick reply.

PR #824, opened 2018-08-01
Added changelog for next release

framp:
As per title and following issue https://github.com/sidorares/node-mysql2/issues/820.

Regarding the version number:
The new features in this release are backward compatible, therefore I bumped the minor only (following semver).

Given we're also dropping support for node 4, we may want to bump the major.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,17 @@
+1.6.0 (01/08/2018)
+ - Fixed `PromiseConnection.ping()` ignoring errors   #813
+ - Added a uri parameter to the connection config     #815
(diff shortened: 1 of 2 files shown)


sidorares:
Thanks! I'd probably still call it minor version change - package version is not fully automated magical number, it's a way to communicate expectations to humans. In that regard node v4 going past end of life is already communicated well enough

sidorares:
I'll publish from home computer in couple of hours. Great job putting summary together!

sidorares:
1.6.0 in npm

Issue #825, opened 2018-08-01
Invalid config option "Promise"


derN3rd:
Hey there,

Looks like #815 and its checks doesn't like the "Promise" prop in ("mysql2/promise").createPool config.
Its used, but shows an ugly warning.

My (typescript) code:
```ts
import * as BluebirdPromise from 'bluebird'
import { db as dbCredentials } from './config'
import * as mysql from 'mysql2/promise'

const pool = mysql.createPool({
  Promise: BluebirdPromise,
  host: dbCredentials.host,
  user: dbCredentials.user,
  password: dbCredentials.password,
  database: dbCredentials.database,
  charset: 'utf8mb4',
  timezone: process.env.timezone || 'local',
  connectionLimit: dbCredentials.connectionLimit,
})
```

Console output when using `pool.query`:
`Ignoring invalid configuration option passed to Connection: Promise. This is currently a warning, but infuture versions of MySQL2, an error will be thrown if you pass an invalid configuration options to a Connection`

MichaelPeter-Shockoe:
For added information: this does not just apply to pools, but also creating a single connection

sidorares:
@DayBr3ak is your issue related to original issue?

If you try to `await` non-promise connection you should see warning now because of #813

PR #826, opened 2018-08-01
Fix missing Promise option in checks for Pool

derN3rd:
See #825


sidorares:
Thanks! published as v1.6.1

PR #847, opened 2018-09-02
chore(*): reduce dependencies

SimonSchick:
1) Not sure about the package-lock.json, please advise.
2) Since we are on node v6 I took the liberty to do some change some _surrounding_ code:
  Make imports use `const`
  Remove `object-assign` polyfill.


sidorares:
Thanks!

PR #850, opened 2018-09-09
Update README.md

kwdiwt:
MDN uses "ES2017" - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Specifications


kwdiwt:
@sidorares okey

PR #856, opened 2018-10-04
document query() method with at least a string

danielecr:
pool.query() method need requires a string first parameter, optional array as second, and callback (and execute() also)
examples/pool-test.js has to be fixed as well (TODO)


sidorares:
Thanks! Looks like this was copy-paste mistake from copying `pool.getConnection()`

Issue #711, opened 2018-01-19
Got premature close error when using pump


Congelli501:
If you don't know about it, pump is a module to better handle stream pipe (better error handling and better close handling to prevent resource leak) https://www.npmjs.com/package/pump.
It is considered by many as safer than pipe and may be part of node core in the future (https://github.com/nodejs/node/pull/13506).

mysql2 streams are not working well with pump: a "premature close" error is thrown at the end of the stream.

Here is an example code to reproduce the error:
```js
const mysql = require('mysql2/promise');
const pump = require('pump-promise');
const {Writable} = require('stream');

/**
 * Main test function
 */
async function main () {
	const connection = await mysql.createConnection({
		/* ... */
	});

	const sqlReadable = connection.connection.query('select 1')
		.stream();

	const testWritable = new Writable({
		objectMode: true,
		write (chunk, encoding, callback) {
			console.log(chunk);
			callback();
		}
	});

	// Not working
	await pump([
		sqlReadable,
		testWritable
	]);


	// Working
	//sqlReadable.pipe(testWritable);
	//await new Promise(resolve => testWritable.on('finish', resolve));

	connection.close();
}

main()
	.then(() => console.log('ok'))
	.catch((e) => {
		process.exitCode = 1;
		console.error(e);
	});
```

You can uncomment the "working" code (and comment the "not working" code) to use the working pipe version.

sidorares:
thanks for report @Congelli501 ! Have you tried to investigate what could cause that?

In promise version we'll likely prefer async iterators to stream, with api similar to something like this:

```js
for await (const row of conn.queryRows(sql)) {
  console.log(row);
}
```

Congelli501:
The problem is caused by the 'close' event being emitted before the 'end' event.
Node's `pipeline` (https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback) also breaks with this behaviour.

vicary:
@Congelli501 @sidorares Sorry to bring this up again, but I think I hit the same issue.

Using `for await` still produces a premature close from the stream, I think it's caused by the row data still being transmitted.

Wrapping `stream.emit('close');` inside a `setImmediate` would allow the data event to be consumed by the async generator before ending the stream, preventing the error.

Shall I open a new issue instead?

nofarham:
@sidorares any news on this issue? We're using node 18 and getting the `[ERR_STREAM_PREMATURE_CLOSE]: Premature close` error.
[...]

PR #859, opened 2018-10-13
Fix issue #711: close emitted before end

Congelli501:
The 'end' event (emitted by `stream.push(null)`) should always be emitted before the close event.
This fixes issue https://github.com/sidorares/node-mysql2/issues/711


sidorares:
Thanks @Congelli501 !

PR #879, opened 2018-11-13
Update query.js

godforall:
Add namedPlaceholder option in Query object


godforall:
I have created a test and also found additional problem in execute command. You will see the commit.

PR #882, opened 2018-11-16
fix(debug): remove usage of callee

### LINKED ISSUE #881: The debug option stopped working in 1.6.2
It seems like version 1.6.2 broke the `debug: true` option from working. I was able to dissect it to the following change: https://github.com/sidorares/node-mysql2/commit/e0077c0892fa27fe4bd39334be4a29dfe1592d18

The commit adds `'use strict'` to [lib/connection.js](https://github.com/sidorares/node-mysql2/blob/e0077c0892fa27fe4bd39334be4a29dfe1592d18/lib/connection.js) but there is a debug log in there that uses [arguments.callee](https://github.com/sidorares/node-mysql2/blob/e0077c0892fa27fe4bd39334be4a29dfe1592d18/lib/connection.js#L471) which is an error under strict mode.

I'm not sure what the appropriate fix would be. A quick fix would be just to remove `'use strict'`, but not sure if there should be additional refactoring to just make it work under strict mode.

SimonSchick:
Fixes #881
I kinda want to add a test but maybe it would be a good idea to run one of the test instances with `DEBUG=1` instead, that would instantly ensure that they at least don't cause errors :)

In any case it's 3am and I really don't want to deep dive into this right now.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -467,9 +467,10 @@ class Connection extends EventEmitter {
     // this.compressedSequenceId = 0;
     // this.sequenceId = 0;
     if (this.config.debug) {
+      const commandName = cmd.constructor.name;
       // eslint-disable-next-line no-console
-      console.log('Add command: ' + arguments.callee.caller.name);
-      cmd._commandName = arguments.callee.caller.name;
+      console.log('Add command: ' + commandName);
+      cmd._commandName = commandName;
     }
     if (!this._command) {
       this._command = cmd;


dougwilson:
LGTM. I don't think it was pressing to stay up if you didn't want to :) but sometimes for me I'm just in the moment and do so even though I shouldn't lol.

kschat:
Sorry to nercopost an old PR, but was this ever released to `npm`? I have the latest version (AFAIK, 1.6.4) and `debug` mode is still broken due to the following error:

`caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them`.

SimonSchick:
I don't think there has been a release.

sidorares:
@kschat @SimonSchick I just published latest master as v1.6.5

kschat:
Thank you!

PR #895, opened 2018-12-05
Fixed a bug when ssl config is a boolean and a property is added to it

bl4ckm0r3:
Sequelizejs passes ssl configuration as a boolean.
This will assign `true` to this.ssl at line 99
```js
    this.ssl =
      typeof options.ssl === 'string'
        ? ConnectionConfig.getSSLProfile(options.ssl)
        : options.ssl || false;
```
and eventually pass the check at line 115 and assign a property to a boolean that will throw a TypeError in strict mode.
This doesn't attempt to fix the bug, but it gives a better error message.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -113,6 +113,11 @@ class ConnectionConfig {
       this.timezone = '+' + this.timezone.substr(1);
     }
     if (this.ssl) {
+      if (typeof this.ssl !== 'object') {
+        throw new TypeError(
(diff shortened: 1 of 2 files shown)


sidorares:
@bl4ckm0r3 could you add a little unit test so this fix is not removed accidentally?

bl4ckm0r3:
@sidorares added, let me know if it's ok for you :)

sidorares:
Looks good, thanks!

PR #896, opened 2018-12-06
Fix when ssl is boolean value, it can not be set rejectUnauthorized issue

kunyan:
If `ssl` is a boolean value, and it is `true`.
the code will set a property for a boolean value and will throw error : Cannot create property 'rejectUnauthorized' on boolean 'true'

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -112,7 +112,7 @@ class ConnectionConfig {
       // connection string..
       this.timezone = '+' + this.timezone.substr(1);
     }
-    if (this.ssl) {
+    if (this.ssl && typeof this.ssl === 'object') {
       // Default rejectUnauthorized to true
       this.ssl.rejectUnauthorized = this.ssl.rejectUnauthorized !== false;
     }


sidorares:
oh wow, 2 PRs fixing same issue in less than 12 hrs

@kunyan thanks for doing this, but I think #895 covers it

josephgrossberg:
@kunyan and @bl4ckm0r3 Thank you both so much -- I need to resolve this issue at work tomorrow, and I'm really happy that you two saved me hours of time! :)

sidorares:
@kunyan @josephgrossberg @bl4ckm0r3 @sushantdhiman Was that reported to sequelize? Maybe worth fixing on their side as well

kunyan:
@sidorares sequelize extends the options from mysql2.
https://github.com/sequelize/sequelize/blob/master/lib/dialects/mysql/connection-manager.js#L76

dougwilson:
> if ssl: true is valid option mysql2 should not throw for it

The "ssl" option has never accepted a boolean afaik

sidorares:
closing in favour of similar #895

PR #915, opened 2019-02-12
(fix/query): Row parsing could lead to uncaught errors

miroslavLalev:
Data type parsers could throw an exception, e.g. `JSON.parse` for the JSON type. Currently, this exception is unhandled and crashes the application that receives it.

This change does the catching and passes it to either the callback, or the error event subscribers in order to support graceful recover if needed.

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -226,12 +226,18 @@ class Query extends Command {
       }
       return this.done();
     }
-    const row = new this._rowParser(
-      packet,
-      this._fields[this._resultIndex],
-      this.options,
-      CharsetToEncoding
-    );
+    let row;
+    try {
+      row = new this._rowParser(
+        packet,
+        this._fields[this._resultIndex],
+        this.options,
+        CharsetToEncoding
+      );
+    } catch (err) {
+      this._localStreamError = err;
+      return this.doneInsert(null);
+    }
     if (this.onResult) {
       this._rows[this._resultIndex].push(row);
     } else {
--- a/test/unit/commands/test-query.js
+++ b/test/unit/commands/test-query.js
@@ -1,0 +1,20 @@
+'use strict';
+
+const assert = require('assert');
+const Query = require('../../../lib/commands/query');
+
+const testError = new Error('something happened');
+const testQuery = new Query({}, (err, res) => {
+  assert.equal(err, testError);
+  assert.equal(res, null);
+});
+
+testQuery._rowParser = class FailingRowParser {
+  constructor() {
+    throw testError;
+  }
+};
+
+testQuery.row({
+  isEOF: () => false
+});


sidorares:
Do you have any particular example in mind when this can happen @miroslavLalev ? Maybe we should add some fuzz testing tool to make sure client never crashes even when server responds with broken data

miroslavLalev:
Hello @sidorares, this is an example how I encountered this issue.

I was working on file parser that extracts metadata from a file and sends it to service, where they are persisted in the database (mysql v5.7.14). There used to be a bug in which this parser would accept any type of files, so we had an incoming request with broken settings, that were parsed from archive file, and the persisted json would look like this (note that it's still a valid json, just the encoding conversion to utf8 doesn't make sense): 

screen shot 2019-02-13 at 10 47 03

Then, after trying to query the row that contained this corrupted data from my node application, it was crashing with the following unhandled exception:

```
    SyntaxError: Unexpected token  in JSON at position 616
        at JSON.parse (<anonymous>)
        at eval (eval at line.toFunction (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/generate-function/index.js:55:21), <anonymous>:29:142)
        at ConnectionManager._typecast (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/sequelize/lib/dialects/mysql/connection-manager.js:61:12)
        at new TextRow (eval at line.toFunction (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/generate-function/index.js:55:21), <anonymous>:29:32)
        at Query.row (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/mysql2/lib/commands/query.js:232:13)
        at Query.Command.execute (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/mysql2/lib/commands/command.js:40:20)
        at Connection.handlePacket (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/mysql2/lib/connection.js:516:28)
        at PacketParser.onPacket (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/mysql2/lib/connection.js:94:16)
        at PacketParser.executePayload (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/mysql2/lib/packet_parser.js:139:12)
        at Socket.<anonymous> (/go/src/git.chaosgroup.com/vcloud/saas/node_modules/mysql2/lib/connection.js:102:29)
        at emitOne (events.js:116:13)
        at Socket.emit (events.js:211:7)
        at addChunk (_stream_readable.js:263:12)
        at readableAddChunk (_stream_readable.js:250:11)
        at Socket.Readable.push (_stream_readable.js:208:10)
        at TCP.onread (net.js:601:20)
```

I suspect that this is a problem with the mysql server returning bad (possibly not properly escaped) data, but couldn't find any issue related to it.

Anyway, I think the fix makes sense even without this context because any data type parser could be prone to throwing exception and the mysql client should return it. About the fuzz testing, I really like the idea but it might be appropriate for another pull request in case other issues are found and gets out of scope. Do you want me to open issue for it, linking this pull request as example where things could go wrong?

sidorares:
I'm pretty sure JSON field payload  (as we see it after receiving on the client side) was not correct for whatever reason ( would be good to track why) - incorrect decoding or json content broken on server.

Would be good to add some unit test for this, maybe around similar example you had using mocked server ( server returns json field type in field description but invalid json as row payload). Let e know if you need help with that

miroslavLalev:
I was quite busy the last couple of days but found some time to start doing the test today. Currently, I am struggling to create the mocked server duplex stream that returns everything in correct order when being read. I couldn't find anything similar in the existing unit tests as well. Do you have any ideas how to do it properly?

miroslavLalev:
@sidorares I couldn't get the custom mock server to work, so I wrote test with throwing `rowParser` to assert that the error is correctly passed to callback.

sidorares:
thanks @miroslavLalev , sorry it took so long, I finally found some time to review old PRs

PR #931, opened 2019-03-06
Connection.execute() must honor callback

jafl:
If there is a callback provided to Connection.execute(), do not throw any exceptions.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -600,29 +600,41 @@ class Connection extends EventEmitter {
     // check for values containing undefined
     if (options.values) {
       //If namedPlaceholder is not enabled and object is passed as bind parameters
+      let err = null;
       if (!Array.isArray(options.values)) {
-        throw new TypeError(
+        err = new TypeError(
           'Bind parameters must be array if namedPlaceholders parameter is not enabled'
         );
       }
-      options.values.forEach(val => {
+      err || options.values.every(val => {
         //If namedPlaceholder is not enabled and object is passed as bind parameters
         if (!Array.isArray(options.values)) {
-          throw new TypeError(
+          err = new TypeError(
             'Bind parameters must be array if namedPlaceholders parameter is not enabled'
           );
+          return false;
         }
         if (val === undefined) {
-          throw new TypeError(
+          err = new TypeError(
             'Bind parameters must not contain undefined. To pass SQL NULL specify JS null'
           );
+          return false;
         }
         if (typeof val === 'function') {
-          throw new TypeError(
+          err = new TypeError(
             'Bind parameters must not contain function(s). To pass the body of a function as a string call .toString() first'
           );
+          return false;
         }
+        return true;
       });
+
+      if (err && cb) {
+        cb(err);
+        return null;
+      } else if (err) {
+        throw err;
+      }
     }
     const executeCommand = new Commands.Execute(options, cb);
     const prepareCommand = new Commands.Prepare(options, (err, stmt) => {
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -161,9 +161,11 @@ class Pool extends EventEmitter {
         return cb(err);
       }
       const executeCmd = conn.execute(sql, values, cb);
-      executeCmd.once('end', () => {
-        conn.release();
-      });
+      if (executeCmd) {
+        executeCmd.once('end', () => {
+          conn.release();
+        });
+      }
     });
   }
 


sidorares:
can you give a bit more context about the issue you solving? Also need a unit test

aikar:
:-1: from me. this is validation logic of the call. The callback should only be called with an error from the server itself.

It should throw an error on the call and break the caller so they can't ignore that error.

This would mix scope of what the error in the callback is in relation to.

jafl:
My reason for suggesting this patch was that, when using the promise api, there doesn't seem to be any way to catch the exceptions.  If you know of a way to do that, please let me know!

sidorares:
> If you know of a way to do that, please let me know!

with async/await

```js
try {
  const result = await connection.execute(sql, parameters);
}
catch(error) {
   handle(error);
}
```

jafl:
OK, I see my problem.  I was expecting all exceptions to go through the Promise, but the validation exceptions don't.

jafl:
There is still a problem.  When I use a pool instead of a connection:

```
try {
  const result = await pool.execute(sql, parameters);
}
catch(error) {
   handle(error);
}
```

the error is not caught.  Instead, I get the following stack trace:

```
.../node_modules/mysql2/lib/connection.js:616
          throw new TypeError(
          ^

TypeError: Bind parameters must not contain undefined. To pass SQL NULL specify JS null
    at options.values.forEach.val (.../node_modules/mysql2/lib/connection.js:616:17)
    at Array.forEach (<anonymous>)
    at PoolConnection.execute (.../node_modules/mysql2/lib/connection.js:608:22)
    at getConnection (.../node_modules/mysql2/lib/pool.js:163:31)
    at process.nextTick (.../node_modules/mysql2/lib/pool.js:44:37)
    at processTicksAndRejections (internal/process/next_tick.js:74:9)
```

aikar:
this error is by design not supposed to be 'caught'. It SHOULD crash your app, as this is a developer error informing you of a bug in your code to fix.

This is not the type of error you should be trying to catch.

sidorares:
try / catch with await should be able to catch both sync and async errors

jafl:
@aikar - During development, yes, I agree. However, in production, it is NEVER acceptable to crash under any circumstance. The app should log it and keep running.

@sidorares - try/catch does normally catch all exceptions, but in this particular case, using a promise pool, it somehow doesn't. Please try it and let me know if you can reproduce it.

sidorares:
@jafl you are right, promise wrapper is not able to catch and reject that error, this needs to be fixed

sidorares:
so what your PR does is routing this error to a standard "error first" callback way instead of sync throw
I'm partially agree with @aikar in that we need to assist in locating this types of errors early in the code but at the same time having an error that is absolutely not possible to intercept is not good as well

jafl:
Yes, it's a very interesting edge case. I will try to figure out the underlying cause. Hopefully, that will show us a way to fix it without using the callback method.

jafl:
As an aside, can you explain why `undefined` is not treated the same as `null`?  It seems that the original mysql module does this.  I feel that `undefined` and `nulll` both mean the same thing, "no value," and the database should be making the decision on whether or not an empty value is allowed.

sidorares:
> an you explain why undefined is not treated the same as null

for me personally two reasons:
- mysql itself does not have "undefined" type
- I want to prevent users from accidentally sending NULL when it's in fact missing or uninitialised parameter

I can see convenience of treating undefined same as null but imo reasons above outweigh benefits. Maybe it's better to judge if we have few simple real life scenarios

aikar:
I believe I was involved with the undefined as an error change as I know that hit me in that I had a bug in my application that shouldn't of been sending undefined, that the error would have caught.

jafl:
@aikar The database didn't complain about the "undefined" value, generating an error that the value was not allowed to be null?

jafl:
I think I figured it out. The summary is that the "extra" layer of callback introduced by the pool breaks the connection between the promise and the code in `connection.execute` that throws the exception.

My understanding is that the promise creates an execution stack, but that is finished once the pool registers the callback for a connection and returns.  When the connection is available, a separate execution stack is created, and this is what gets unwound when the exception is thrown inside `connection.execute`.  (Normally, `await` simply pauses the execution stack, so exceptions can unwind it.)

I will investigate if using `PromiseConnecton` inside `PromisePool` can work around this.

jafl:
#938 is a better fix

PR #933, opened 2019-03-08
fix #932

### LINKED ISSUE #932: Error in Promise Wrapper section in README, with a question that how to check whether success or not after connecting to databse
In README,there is one code which has problem,see below:
```js
async function main() {
  // get the client
  // create the connection
  mysql.createConnection(
   {host:'localhost', user: 'root', database: 'test'},
   (err,con) => {
    con.promise().query("SELECT 1")
    .then( ([rows,fields]) => {
     console.log(rows);
    })
    .catch(console.log)
    .then( () => con.end());
   }
  );
```
I tried this, but can't go to callback.Then I check out the source code, find there is no second arguments in `createConnection`.If so, how can I check whether success or not after connecting to databse?

meloseven:
fix #932

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,19 +202,18 @@ async function main() {
 
 MySQL2 exposes a .promise() function on Connections, to "upgrade" an existing non-promise connection to use promise
 ```js
-async function main() {
-  // get the client
-  const mysql = require('mysql2');
-  // create the connection
-  mysql.createConnection(
-   {host:'localhost', user: 'root', database: 'test'}
-  );
-  con.promise().query("SELECT 1")
-    .then( ([rows,fields]) => {
-     console.log(rows);
-    })
-    .catch(console.log)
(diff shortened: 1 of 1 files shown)


sidorares:
I think you lost closing `}` matching 'async function main() {`. Maby better to remove `async function main() {` from example altogether as it's not using async/await?

meloseven:
Sorry, let me modify it.

sidorares:
sorry for nits but I'd really like for this change to result in example that can me almost copy pasted, and in your version `con` is undefined, should be something like `const connection = mysql.createConnection({host:'localhost', user: 'root', database: 'test'})`; con.promise().query("SELECT 1").then(...)`

meloseven:
I'm terribly sorry.

sidorares:
Thanks @meloseven !

PR #938, opened 2019-03-28
Remove callback layer to catch sync connection exceptions

jafl:
This PR restructures PromisePool's `query` and `execute` so they catch synchronous exceptions.

I don't think `Connection.query` actually throws synchronous exceptions, so we could remove the changes to support `PromisePool.query`, but I felt that would be going only half way.  Why doesn't `Connection.query` throw exceptions like `Connection.execute`?

Alternative to #931


jafl:
I tried merging the latest from master, but the precommit hook does a major reformatting job.  This happens even when I pull a clean copy without any pre-existing node_modules directory.  How can I get around this?

jafl:
> The `return process.nextTick(() => cb(null, connection));` needs to instead be `return process.nextTick(() => { try { cb(null, connection); } catch (e) { cb(e, null); } });`

How does this help with catching exceptions related to `undefined` values?

jafl:
> @jafl sorry I missed this pr. One thing I don't like is that you want to duplicate pool connection release logic in promise wrapper while currently it's delegated to base implementation

I wasn't happy with it, either, but I couldn't see any other way to make the exceptions catchable.  Duplicating the logic removes the extra layer, so exceptions propagate correctly.

sidorares:
@jafl yeah, I see your point

can you try to check where error stack points with your changes when you run this

```js

(async function() {
  try {
    await pool.execute('select 1+?', [ undefined ]);
  } catch (e) {
    console.log('It would be great if I could catch the ball!');
  } finally {
    await pool.end();
  }
})();
```

also would be good to add unit test

jafl:
Sorry for taking so long.  Life got in the way...

I added your code as a test in `test-promise-wrappers.js`, and it prints:

```
  It would be great if I could catch the ball!
  error: TypeError: Bind parameters must not contain undefined. To pass SQL NULL specify JS null
      at /Users/jolindal/tools/node-mysql2/lib/connection.js:602:17
      at Array.forEach (<anonymous>)
      at PoolConnection.execute (/Users/jolindal/tools/node-mysql2/lib/connection.js:594:22)
      at /Users/jolindal/tools/node-mysql2/promise.js:110:11
      at new Promise (<anonymous>)
      at PromisePoolConnection.execute (/Users/jolindal/tools/node-mysql2/promise.js:107:12)
      at /Users/jolindal/tools/node-mysql2/promise.js:350:19
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at async testCatchSyncUndefinedError (/Users/jolindal/tools/nodemysql2/test/integration/promise-wrappers/test-promise-wrappers.js:452:5)```

jafl:
Have you had a chance to review the diffs?

sidorares:
Thanks @jafl! could you fix link errors here?
Also, is there any unit test that touches the issue that this PR fixes?

jafl:
Both builds are complaining about indentation - but that is caused by the pre-commit hook.

jafl:
I added the unit test in commit 70b75a7

jafl:
How do I force the pre-commit hook to indent the way lint wants?

jafl:
Build is fixed.

jafl:
What else needs to be done before this PR can be merged?

sidorares:
hey @jafl , the PR probably looks good, I just need to find time to review it as it is quite big

SebastianZaha:
I wanted to use @jafl 's fix branch but it seems if call getConnection() on a pool, the query method on that connection expects a pool parameter as first argument instead of the query string. This, for example crashes: `let conn = await pool.getConnection(); conn.query('BEGIN');`

jafl:
Thank you for catching this!  I have adjusted it.

leibale:
Is this PR going to be merged into master?

sidorares:
@leibale yes

Mikapsien:
My naive sugestion is to return cb(e) in the catch clause of execute in pool.js. It will make our code simpler unlike now manually checking for all values.

jafl:
> My naive sugestion is to return cb(e) in the catch clause of execute in pool.js. It will make our code simpler unlike now manually checking for all values.

Can you provide more details?  What values are you checking for? Can you provide example code?

jafl:
Is this PR still useful, or is it obsolete?

jafl:
@sidorares I cleaned up the PR for you to review again.

jafl:
This appears to be working in the latest release.

PR #950, opened 2019-04-18
Always cancel connect timeout

evanskinner:


sidorares:
Looks good to me. Could you add a test for this? Prob use fake server and cut connection

evanskinner:
@sidorares Test now added. Without the fix in connection.js the test registers the connection closed error and then hangs for 10 seconds before throwing ETIMEDOUT. With the fix in connection.js the test completes quickly and just registers the connection closed error.

SpenserJ:
@sidorares Do you have any update on this? I'm running into this issue when my code attempts to connect to the database while it is starting up (server closes the connection because it isn't ready), and the timeout still fires even though I've received the exception and handled it.

The exception that I'm receiving and handling is being sent from [`this.stream.on('close', ...)`](https://github.com/sidorares/node-mysql2/blob/v2.1.0/lib/connection.js#L84-L98), and after applying this patch with Yarn, the issue is no longer occurring.

sidorares:
@SpenserJ @evanskinner somehow missed update here. I'll prepare release to npm later this week

vlasky:
@sidorares eagerly awaiting the release of the NPM with this uncatchable timeout exception fix! :-)

Issue #960, opened 2019-05-07
Allow to pass secureProtocol or minVersion to ssl options


xrnoz:
Node 12.1.0 defaults to tls version 1.2 and mysql 5.7 on ubuntu 18.04 uses tls 1.1, so to be able to connect I need to run node with --tls-min-v1.1 option.

I would like to be able to set the tls version with the secureProtocol or minVersion options that are received by Tls.createSecureContext so I don't have to run node with --tls-min-v1.1.

sidorares:
We need to add extra option here - https://github.com/sidorares/node-mysql2/blob/58c1056edd21ad42e2e755959b75f76b502e3f72/lib/connection.js#L299-L305

Do you want to volunteer to do that @xrnoz ?

stephen-dahl:
how to set this while setting ssl to 'Amazon RDS'? right now i am doing 
```
const certs = require('mysql2/lib/constants/ssl_profiles');
{ssl: { ...certs['Amazon RDS'], minVersion: 'TLSv1.1' }}
```

sidorares:
@stephen-dahl maybe we should make it public. There is no magic behind ssl profiles, it's just name => "ssl config object" hash, should be accessible for others

PR #961, opened 2019-05-08
Allow to set minVersion ssl option

xrnoz:
This option allows Node v12.1.0 to connect to a server using 'TLSv1.1' because Node v12.1.0 minVersion default is 'TLSv1.2'.

Otherwise Node v12.1.0 has to be run with --tls-min-v1.1 CLI option.


sidorares:
this should fix #960

PR #965, opened 2019-05-13
Fix a MaxListenersExceededWarning with stream local infile

darksabrefr:
The error event on a connection is never removed with stream local infile, resulting in a MaxListenersExceededWarning from node at the 11th attempt to stream on this connection. This change fixes this problem.


sidorares:
Thanks, looks good!

PR #968, opened 2019-05-13
Fix #967 - Return a string for '00:00:00' time

### LINKED ISSUE #967: execute and query, serialisation of '00:00:00' time
Run the following scripts

### Database

```sql
CREATE TABLE `abc` (
  `time` time NOT NULL COMMENT 'The time'
) ENGINE=InnoDB DEFAULT CHARSET=latin1;

INSERT INTO `abc` (`time`) VALUES
('00:00:00'),
('18:00:00');
```

### Node

```node
// get the client

// create the connection to database
  host: 'localhost',
  user: 'root',
  database: 'aa'
});

// simple query
connection.query(
  'SELECT * FROM `abc`',
  function(err, results, fields) {
    console.log(results); // results contains rows returned by server
    //console.log(fields); // fields contains extra meta data about results, if available
  }
);

connection.execute(
  'SELECT * FROM `abc`',
  function(err, results, fields) {
    console.log(results); // results contains rows returned by server
    //console.log(fields); // fields contains extra meta data about results, if available
  }
);
```

### Result
```
[ TextRow { time: '00:00:00' }, TextRow { time: '18:00:00' } ]
[ BinaryRow { time: 0 }, BinaryRow { time: '18:00:00' } ]
```

williamdes:
The method `readTimeString` returns a string in all other cases, so it should return `00:00:00`

Fixes: #967

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -319,7 +319,7 @@ class Packet {
   readTimeString(convertTtoMs) {
     const length = this.readInt8();
     if (length === 0) {
-      return 0;
+      return '00:00:00';
     }
     const sign = this.readInt8() ? -1 : 1; // 'isNegative' flag byte
     let d = 0;


williamdes:
@sidorares Can you please review this very small PR ? ;)

sidorares:
Hey @williamdes  sorry I forgot about this. Can you add a small unit test? Otherwise looks good to me

williamdes:
Hey @sidorares
I had a look and I do not feel like I could make a good test
I would prefer that you add one

sidorares:
No problems, I'll try to work on this over weekend

williamdes:
@sidorares?

williamdes:
@sidorares I think you can merge the fix without adding a test?

PR #983, opened 2019-07-17
Ping connection from the pool to check health

zebullax:
Looks like the connection from the pool are not checked, after long inactivity this may lead to issue with the other side closing the connection and the client not knowing (https://stackoverflow.com/questions/56919129/econnreset-etimedout-error-with-mysql2-nodejs-pooled-connection)

Since mysql2 is based off mysql, i found out that pinging the connection is an ok first solution (https://github.com/mysqljs/mysql/blob/master/lib/Pool.js#L119), this very rough PR is to start addressing that.
Keep in mind that i am not that familiar with this lib/js so this may be lacking...

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -311,7 +311,17 @@ class PromisePool extends EventEmitter {
         if (err) {
           reject(err);
         } else {
-          resolve(new PromisePoolConnection(coreConnection, this.Promise));
+          function onPingOk() {
+            // Ping went through, connection good to be handed to client
+            resolve(new PromisePoolConnection(coreConnection, self.Promise));
+          }
+          function onPingError(err) {
+            // Ping went dead, release the connection for it is dead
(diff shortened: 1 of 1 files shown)


sidorares:
hm, I'd rather allow users what to do when connection acquired. Ping requires full roundtrip connection  time, that might be a big price for some. Also when you make query you still have no guarantee that connection is still alive, what if it died after ping? You still need to know how to handle errors and how to recover ( retry / report up etc )

Overall the feature might be practical for most users, but I'd like to discuss it more

sidorares:
did you change your mind @zebullax or want to continue this discussion in separate issue?

zebullax:
I just noticed the number of other related PR / Issues on this topic, I dont think it is worth having another one of those, I'll wait and hope one of them land on release branch.

sidorares:
feel free to link them here

PR #993, opened 2019-08-04
replace deprecated inspect with symbol inspect

aditya1906:
inspect function is no longer supported as mention in https://nodejs.org/api/deprecations.html#deprecations_dep0079_custom_inspection_function_on_objects_via_inspect and according to https://nodejs.org/api/util.html#util_util_inspect_custom it is defined as a shared symbol.


sidorares:
could you rebase from master @aditya1906 ? Build matrix now only includes newer version of node where `Symbol` can be used

PR #994, opened 2019-08-06
Add threadId and config getters to PromiseConnection

trevorr:
These properties are documented in `mysqljs/mysql` [[`config`](https://github.com/mysqljs/mysql#custom-format)] [[`threadId`](https://github.com/mysqljs/mysql#getting-the-connection-id)] and typed in [`types/mysql2`](https://github.com/types/mysql2/blob/master/promise.d.ts#L8) but not directly exposed in `PromiseConnection`.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -217,6 +217,10 @@ class PromiseConnection extends EventEmitter {
     });
   }
 
+  get config() {
+    return this.connection.config;
+  }
+
   get threadId() {
     return this.connection.threadId;
   }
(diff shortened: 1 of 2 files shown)


sidorares:
Thanks @trevorr ! can you add simple unit test for this?

trevorr:
I feared you might ask that, as I'd have to downgrade from MySQL 8 to run the tests. 😉 However, your request is fair, so here you go.

sidorares:
I have previous previous experience of accidentally removing / breaking the code that is not used often and not realising that for many month :)

sidorares:
>  I'd have to downgrade from MySQL 8 to run the tests.

I use mysql in local docker to test different versions

Issue #888, opened 2018-11-23
createConnection timezone not working


ktechadmin:
Hi guys,

in below code timezone setting does not seem to have any effort , i think its the issue with the driver could anyone help ?

my sql DB timezone : "-08:00"
application time : "+05:30"

yes this needs to be handled at app level. DB cant change.

var mysql = require('mysql');

//connect to db
var dbCon = mysql.createConnection({
    host: "xxxx",
    user: "xxxxx",
    password: "xxxxx",
    database:"xxxx",
    **timezone:"+05:30"**
  });

  //db connection 
  dbCon.connect(function(err) {
    if (err) throw err;
    console.log("DB Connected!");
  });

var dbCon = db.dbCon;
//query functions 
function get_total_sales(startDate, endDate, callback) {
    dbCon.query("SELECT sum(Paid) as total_sales, count(Paid) as item_soled FROM `orders`  WHERE DateCreated >= " + mysql.escape(startDate) + " AND DateCreated <= " + mysql.escape(endDate) + "  ",
        function (err, result, fields) {
            if (err) {
                throw err;
            } else {

                finalWriteDataObject["total_sales"] = numeral(result[0].total_sales).format();
                finalWriteDataObject["item_soled"] = numeral(result[0].total_saitem_soledles).format();
                callback();
            }

        });
}

get_total_sales(startDate, endDate, writeToFile);

swimmadude66:
Chiming in to add that timezone definitely has no effect. In `node-mysql`, the data-parsing section for date objects handles timezones here: https://github.com/mysqljs/mysql/blob/ad014c82b2cbaf47acae1cc39e5533d3cb6eb882/lib/protocol/packets/RowDataPacket.js#L78

However in `node-mysql2` the same functionality is accomplished here: https://github.com/sidorares/node-mysql2/blob/master/lib/packets/packet.js#L252

the latter does not read or check timezone from the connection config at all, and instead just formats from a parsed string to a `Date` object, which will default to local time.  I have used a workaround for this issue by adding `dateStrings: true,` to my db connection config, which returns the date fields as they are shown in mysql workbench. From there, I cast to a UTC date like so:

```typescript
        const dateString = row.CreateDate as string;
        const utcDate= new Date(dateString .replace(/\s/, 'T')+'Z');
```

Personally, I am going to switch back to using `node-mysql` at least until this is fixed, but I hope that workaround can help other people.

rob5408:
[...]

Issue #877, opened 2018-11-12
Timezone bug: clearly documented


sjb933:
There is a clear bug when specifying database timezones that is easily reproducible.

**Knex.js migration script, written in typescript**
```typescript
import * as Knex from 'knex';

const testTbl = 'TestTime';
const testCol = 'dt';

interface T { id: number; dt: Date };

module.exports = {
    up: async (knex: Knex) => {
        await knex.schema.createTableIfNotExists(testTbl, (t) => {
            t.increments('id').primary();
            t.dateTime(testCol);
        })

        console.log('\n\nDUMPING TABLE\n\n');
        await knex(testTbl).delete();

        const dt = new Date();
        console.log('Creating record with date:', dt.toISOString());
        await knex(testTbl).insert({ dt });

        const [record]: T[] = await knex(testTbl);
        console.log('Record returned with date', record.dt.toISOString());

        console.log('Updating record with same date');
        await knex(testTbl).update({ dt: record.dt }).where({ id: record.id });

        console.log('Record returned', (await knex(testTbl))[0].dt.toISOString());
        console.log('\n\nTEST END');
    },
    down: () => Promise.resolve()
}
```

**RESULTS FROM 4 EXECUTIONS, MYSQL VS MYSQL2, LOCAL VS UTC**
```yaml
Run 1: | # (PASS) DRIVER: MYSQL,  TIMEZONE: LOCAL (default)
  - 'Creating record with date: 2018-11-12T19:45:23.157Z'
  - 'MYSQL DRIVER: Parsing new date from DB "2018-11-12 13:45:23"' # Date stored in DB as local time...
  - 'Record returned with date 2018-11-12T19:45:23.000Z' # Date parsed correctly
  - 'Updating record with same date'
  - 'MYSQL DRIVER: Parsing new date from DB "2018-11-12 13:45:23"'
  - 'Record returned 2018-11-12T19:45:23.000Z' # Date is stable over reads/writes

Run 2: | # (PASS) DRIVER: MYSQL2,  TIMEZONE: LOCAL (default)
  - 'Creating record with date: 2018-11-12T19:47:39.630Z'
  - 'MYSQL DRIVER: Parsing new date from DB "2018-11-12 13:47:40"' # Date stored in DB as local time...
  - 'Record returned with date 2018-11-12T19:47:40.000Z' # Date parsed correctly
  - 'Updating record with same date'
[...]

PR #996, opened 2019-08-11
Add support for timezone connection option

### LINKED ISSUE #15: Use timezone from connection options
'SELECT FROM_UNIXTIME(631152000)' returns the timestamp in local tz instead of UTC (e.g. for PST, equivalent of 1990-01-31T22:00:00.000Z instead of expected 1990-01-01T00:00:00.000Z)

Simliar issue in mysql-native has been proposed a fix in sidorares/nodejs-mysql-native#78

trevorr:
Support for the `timezone` connection option as specified by `mysqljs/mysql`:

> `timezone`: The timezone configured on the MySQL server. This is used to type cast server date/time values to JavaScript `Date` object and vice versa. This can be `'local'`, `'Z'`, or an offset in the form `+HH:MM` or `-HH:MM`. (Default: `'local'`)

Inspired by https://github.com/sidorares/node-mysql2/pull/732 but with tests and against current master. Removes the need to run tests with `process.env.TZ = 'UTC'`. It doesn't set the session `time_zone` variable, but the user should do that if the server time zone is not the same as the requested time zone. However, since `time_zone` only applies to `timestamp` columns and date functions &mdash; and not stored `datetime` values &mdash; the client and server really need to agree on the time zone in advance anyway.

Fixes #15, #262, #642, #877, #888, and maybe others.

### DIFF
--- a/documentation/Readme.md
+++ b/documentation/Readme.md
@@ -36,19 +36,6 @@ You need to check corresponding field's zeroFill flag and convert to string manu
 ```
 **Note :** *This option could lose precision on the number as Javascript Number is a Float!*
 
-- `timezone` connection option is not supported by `Node-MySQL2`. You can emulate this by using `typeCast` option instead:
-```javascript
-const config = {
-  //...
-  typeCast: function (field, next) {
-    if (field.type === 'DATETIME') {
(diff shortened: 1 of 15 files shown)


sidorares:
Great work @trevorr !

PR #1003, opened 2019-08-19
Support ConnectionConfig.flags as an array

trevorr:
[mysqljs/mysql supports a string or an array for flags](https://github.com/mysqljs/mysql/blob/master/lib/ConnectionConfig.js#L146)
and the TypeScript bindings expect an array (until [this upstream fix](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37722) can be integrated)


trevorr:
@sidorares Thoughts on this?

sidorares:
Somehow missed that. LGTM

Issue #906, opened 2019-01-05
MySQL 8 SHA256 Support


normano:
Hey, I'm wondering if you still planned on adding mysql 8 support. I did find one branch on here that has your WIP changes for it and I wonder if you will continue with it? I might switch to mysqljs since they are close to merging support. I was trying to do it myself but I don't understand it enough! The first marker in the handsahake is 254 and then it comes back with 1 and data is 0x04 which I guess it wants the FullAuthentication to be performed.

sidorares:
yes, I'm planning to work on this some time but if anyone would like to contribute that would be great

One way I want to implement this differently from https://github.com/mysqljs/mysql/pull/1962 is to have it part of auth plugin system and not treat specially. Unfortunately I discovered that auth plugin need big rewrite as well

I'm happy to answer any questions you have re auth plugins in general and sha256 particularily here @normano !

normano:
Did you have any spec for the rewrite? I don't know enough about MySQL auth, but I'm curious to know where you are going with your client side plugin support.

I'm gonna try again today, maybe I can respond to the FullAuthentication request properly and make it work.

sidorares:
I'll try to document more later, but here are few starting points:

Auth switch packet documentation: https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest

Current api: https://github.com/sidorares/node-mysql2/blob/master/documentation/Authentication-Switch.md implemented in https://github.com/sidorares/node-mysql2/blob/efec9d5cfc3496bb254f72d5a8d7258d797c2e3e/lib/commands/client_handshake.js#L162-L173

Problem with current api is that it's difficult to make plugin that supports multiple responses from server and maintains state

Also see https://github.com/sidorares/node-mysql2/issues/560

normano:
Still trying to attempt this, but seems fullauthentication request (plugindata == 0x04) is requested then need to send clear text pass if ssl is not enabled.

```
if(authSwitchHandlerParams.pluginData.toString("hex") === "04") {
connection.writePacket(new Packets.ClearText(this.password).toPacket());
}
return null; // Not sure if this is correct
```

if fast path then I just return null currently.

Seems like the auth worked, but then get
Warning: got packets out of order. Expected 1 but received 5
Warning: got packets out of order. Expected 2 but received 1
Warning: got packets out of order. Expected 3 but received 2
Warning: got packets out of order. Expected 4 but received 3
[...]

Issue #991, opened 2019-07-31
Error: Client does not support authentication protocol requested by server; consider upgrading MySQL client


infusion:
Since MySQL 8, a new authentication algorithm is used, which introduces a breaking change. The mysql npm package received a PR for this. Maybe this helps improving this awesome packag: https://github.com/mysqljs/mysql/pull/2233

Robert

sidorares:
thanks @infusion , I'm aware of this and have put some initial efforts just didn't have time to finish ( I'd like to take slightly different approach to a mysqljs though ). See som related discussions in #975 #906 #820 . There is also some code in https://github.com/sidorares/node-mysql2/tree/mysql8-wip and https://github.com/sidorares/node-mysql2/tree/feature/fast-caching_sha2_password branches. Closing issue for now but feel free to add any relevant feedback here!

infusion:
That's awesome news, thanks for the effort! Can you give any release date information?

sidorares:
Regarding mysql 8 change - not going to promise anything, it was "very close" a year ago

unfortunately very little time right now, I'm the only active maintainer and I have day job and 6 kids :)

I'll try to keep doing small releases more often, maybe I'll cut next release this coming weekend

fabiofcferreira:
Experiencing this issue and just wanted to add my support on this issue.

I can definitely start reading the docs and code and hopefully help you maintaining this wonderful package, what do you say? @sidorares

sidorares:
Hey @fabiofcferreira , thanks for the offer! Happy to accept help, I'll try to write down priorities here in few minutes ( I think I already did that in previous issues, need to find and link to here )

sidorares:
Ok so initial support for plugin auth landed 3 years ago ( https://github.com/sidorares/node-mysql2/pull/331 ) and I think it's time to rewrite it. 

https://github.com/sidorares/node-mysql2/issues/906#issuecomment-451653763
https://github.com/sidorares/node-mysql2/issues/560

So priorities now are:
  - agree on new `authPlugin` api
  - implement new plugin api system, make it work for both initial plugin name and "auth swich" flows ( and also change user command )
  - move existing auth code into corresponding plugins ( mysql_native_password etc ) and call that code from main protocol in case non plugin
   - add `caching_sha2_password` implementation as individual plugin


Main change to authPlugin api - I want to convert handler from (name, data, cb) to probably something like (name) => (data) => Promise<Buffer>
[...]

Issue #560, opened 2017-05-15
allow to use arbitrary plugin as first auth method


sidorares:
currently only allowed connect method initially is `mysql_native_password`. Some servers can potentially prefer to start with custom auth immediately, instead of rejecting  `mysql_native_password` and doing AUTH_SWITH_HANDLER sequence afterwards. Also some servers can be configured to allow 'plugin based auth' but not 'auth switch request' - those are two different capabilities flags

1) respect handshake packet plugin name
2) deprecate `authSwitchHandler` and rename it to be `authPluginHandler`
3) provide default handler for `mysql_native_password`

also need to think of something to make it easy to chain handlers: 
```
const mysqlIamAuth = require('mysql-iam-auth'); // imaginary, does not exist
const mysqMyCustomAuth = require('@internal/customauth');

const pool = mysql2.createPool({
  authPluginHandler: combineAuthHandlers(mysqlIamAuth, mysqMyCustomAuth)
})
```

ref http://stackoverflow.com/questions/43448563/connecting-to-mariadb-with-nodejs-over-ssl-with-clear-text-password/43450396#43450396

sidorares:
thinking about merging plugin auth and auth switch. Maybe api like this: 

```js
const mysql2 = require('mysql');
const pool = mysql2.createPool({
  authPlugins: {
     sha256_password: (data, cb) => { /* ... probably going to be included by default. You shoold be able to disable it by specifying  authPlugins: { sha256_password: null } */ },
     mysql_clear_password: mysql2.authPlugins.mysql_clear_password(validateFunction), // not enabled by default but included with driver and documented
  }, 
  connectAuthMethod: 'mysql_native_password'  // not used if server does not support PLUGIN_AUTH
                                                                               // should it error if specified and no server support for PLUGIN_AUTH ? 
  // some way to handle dynamic plugin names? currently possible with authSwitchHandler but I want to deprecate it
}
```

sidorares:
> it's own kinds of packets though

I think it's ok to assume plugin is onlu able to send data or null, afaik it's never allowed to send any other packet than `AuthMoreData` or OK, but there can be possible multiple exchanges and just having function as plugin is not enough, we need instanse of object that updates it state as it progresses from AuthMoreData to final OK

normano:
I think function that returns the next handler is fine. You as the middle man don't care for state, only what comes next. If I write a plugin then I'll worry about state on my side and know only to return to you what should be called next. 

[...]

PR #1021, opened 2019-09-29
Mysql 8 fixes

sidorares:
still WIP, but tests pass and I think now it's only matter of adding docs

- "authSwitchHandler" is deprecated in favour of "authPlugins"
- mysql8 specific tests fixed ( mostly GRANT not creating users and standard geometry functions name update )
- Add SSL-enabled server to test matrix and test all combos of compression + ssl
- add default authPlugin for `caching_sha2_password` and `mysql_native_password` auth methods

this would not be possible without https://github.com/mysqljs/mysql/pull/1962

cc @ruiquelhas and @johannes

some related previous API discussion:
https://github.com/mysqljs/mysql/issues/1793
https://github.com/sidorares/node-mysql2/issues/560

should fix:
https://github.com/sidorares/node-mysql2/issues/906
https://github.com/sidorares/node-mysql2/issues/991


RedactedProfile:
I know you **just** merged this into master 2 days ago, but might I ask when NPM will be triggered with a release for this?  Thanks

sidorares:
@RedactedProfile the plan is to bump major version, and just to be on the safe side release -alpha1 and have it for few days. I'll try to publish alpha release today, would be good to have you among early testers. I'll post a comment here when published

ruiquelhas:
@sidorares I didn't have the chance to look at the code yet. But I'll gladly run some quick tests of that alpha release.

sidorares:
@ruiquelhas published as `mysql2@2.0.0-alpha1`

robstev0:
mysql2@2.0.0-alpha1 tested and working with 8.0.17 and latest authentication mode.

The timing of this release could not have come at a better time!

sidorares:
Thanks @robstev0 , I'll leave it for few more days in alpha1 and happy to cut 2.0 if no regressions

ruiquelhas:
Seems to be working fine for both `mysql_native_password` and `caching_sha2_password` with MySQL 8.0.18 and some other bleeding edge versions as well. :wink: 

Thanks @sidorares

ruiquelhas:
Works with MySQL 5.7 (`mysql_native_password` only), 8.0.3 (pre-`caching_sha2_password`), 8.0.11 (first **GA** with both authentication plugins) and 8.0.18 (current **GA** with both authentication plugins) using empty or non-empty passwords. 

LGTM

Thanks @sidorares

claytongulick:
Works like a charm for me!

sidorares:
Thanks everyone who helped with testing, I'm planning to publish 2.0 in a day or two (same code as alpha1, maybe documentation updates only)

treyhuffine:
Thanks for the hard work in making this happen @sidorares :)

mwalden2004:
Looks like I'm still having issues with 8.0, and sha256_password. Getting:

`Error: Server requests authentication using unknown plugin sha256_password. See TODO: add plugins doco here on how to configure or author authentication plugins.`

sidorares:
what's your server version @mwalden2004 ? Plugin name is different from what most 8.x server report ( notice it's `sha256_password` and not `caching_sha2_password` )

mwalden2004:
Didn't notice, sorry! The version is 8.0.18.

sidorares:
you could manually alias it like this:

```js
const caching_sha2_password = require('mysql2/lib/auth_plugins/caching_sha2_password.js');
  authPlugins: { 
      sha256_password: caching_sha2_password
  }
})
```

mwalden2004:
Still getting the same error with that code, I made sure that I am running the latest version of mysql2. Don't know what to do from here. Thank you for the assistance so far.

sidorares:
could you try to debug this lines ( with console.log or debugger ) ? https://github.com/sidorares/node-mysql2/blob/ebc2cb438d380d81d57ae6e2d227bfafc684eb2a/lib/commands/auth_switch.js#L42-L50

check spelling and what server expects vs what is actually in authPlugins

Note that I slightly updated the example, `require('mysql2/lib/auth_plugins/caching_sha2_password.js')` returns plugin factory function and you need to use plugin instance in the config

sidorares:
@ruiquelhas any idea what servers return `sha256_password` as plugin name? Is it the same as `caching_sha2_password `? Should I add alias?

mwalden2004:
Okay, looking into this now. So it looks like authSwitchRequest is called two times, the first it sends in sha256_password as the plugin name, and the second time standardAuthPlugins is undefined. After which it appears it errors out with the throw Error()

mwalden2004:
I've added an alias inside the auth switch, and now it's trying to force me to use root! I'll look into this in the morning. Good night!

ruiquelhas:
@sidorares `sha256_password` is a different plugin, but it's definitely not the default one. So, somehow, the `default_authentication_plugin` server option was overridden, either via config file or via the `--default-authentication-plugin`.

https://dev.mysql.com/doc/refman/8.0/en/sha256-pluggable-authentication.html

sidorares:
@mwalden2004 looks like using `caching_sha2_password` here is wrong approach, either you need to configure server to use different auth plugin or we need to add `sha256_password` support

johannes:
Adding support for `sha256_password` isn't rocket science, but I would advise against usage ... it might be worthwhile with 5.7 or older servers as password storage on the server is more secure than `mysql_native_password` (which is relevant for threat cases where an attacker gains access to crypted passwords i.e. from a leaked server backup or something) but on 8.0 servers (thus anything new) the caching variant is notable faster, while bringing the same security benefits over the old native password.

mwalden2004:
Well here is where this gets really interesting, I have two other projects running off this database on the server running mysql2, but this is the only one having any issues. I also am not using sha256_password, looks like it is being forced...

sidorares:
auth plugin can be configured per user / ip. Are those same for two other projects?

mwalden2004:
They are the same.

pokurrunku:
@mwalden2004  did you  find any solution for sha256_password

Pezhvak:
this happened to me as well, it was working fine, all of the sudden it stopped working

sidorares:
Hi @Pezhvak can you clarify exactly what stopped working?

Note that we recently added support for `sha256_password` auth plugin and it's still relatively experimental

Pezhvak:
@sidorares it started giving me that error message which i didn't get before, it turns out to be a problem with terminal which runs it, if i run it on iterm i get that error, if i run it inside phpstorm i don't get that (using run feature of the ide), i can experiment more and let you know if you will.

johannes:
@Pezhvak are you using the same node installation? I could imagine that you might have a global installation with a different version and one node installation picks it up, giving you old code. Running `console.log(require.main.paths)` and `console.log(require.resolve('mysql2'))` could help to verify this.

Pezhvak:
i will check it and get back to you @johannes

TheDirigible:
This just started happening to me out of nowhere, after Windows Terminal updated.  I don't have any plugin, I don't know what that message is talking about.

Pezhvak:
this happened to me, it seems to be something related to user or where from you call your file

Issue #1018, opened 2019-09-26
Promise Pool ignores namedPlaceholders on pool.query


derN3rd:
Hey there,

When I try to use the following (example) code below, the pool.query function ignores the passed namedPlaceholders option and uses the default from pool config.
```
const BluebirdPromise = require('bluebird')
const mysql = require('mysql2/promise')

const db = mysql.createPool({
  Promise: BluebirdPromise,
  host: dbCredentials.host,
  user: dbCredentials.user,
  password: dbCredentials.password,
  database: dbCredentials.database,
  charset: 'utf8mb4',
  namedPlaceholders: false,
})

db.query({
  sql: `SELECT id, name FROM user WHERE type NOT IN (:userTypes)`,
  values: {
    userTypes: [0, 1, 5, 6],
  },
  namedPlaceholders: true,
})
  .then(console.log)
  .catch(e => console.error(JSON.stringify(e)))
```
response:
```
{
  "message": "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':userTypes)' at line 1",
  "code": "ER_PARSE_ERROR",
  "errno": 1064,
  "sqlState": "42000",
  "sqlMessage": "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':userTypes)' at line 1"
}

```
I get the error response, where I can see that the placeholders are not replaced. If I change the pool config to `namedPlaceholders: true` it works. I get the same error when I switch to the non-promise version

Node Version: 8.15.1
Mysql2 Version: 1.6.5

sidorares:
same issue with non-promise pool.query()

this line overwrites local options ( which correctly has `namedPlaceholders` set to true )  https://github.com/sidorares/node-mysql2/blob/c460c0fe9c9b47282f573801f6e883913752916f/lib/pool.js#L135

Need to change it to something like 

```js
if (typeof cmdQuery.namedPlaceholders === 'undefined') {
   cmdQuery.namedPlaceholders = this.config.connectionConfig.namedPlaceholders;
}
```

so it only sets parameter to one from global config if there is no local override

Do you want to test this, create unit test and create pr with change @derN3rd ?

derN3rd:
Just fixed it with your proposed changes, but can't get the tests to work.
Should I just make a PR without the tests?

sidorares:
make a PR and I'll help with tests ( or just add them on top of your commits )

[...]

PR #1022, opened 2019-09-30
fix(pool): fixes pool ignoring namedPlaceholders config

derN3rd:
This fixes pool ignoring the namedPlaceholder setting on query function as mentioned in #1018


test/integration/test-namedPlaceholders.js
```diff
@@ -36,5 +37,13 @@ test('Test namedPlaceholder as command parameter', {
       assert.equal(rows[0].result, 1);
       c.end();
     });
+  },
+  'Disabled in pool config, enable query command': () => {
+    const c = createPool({ namedPlaceholders: false });
+    c.query({ sql: query, namedPlaceholders: false }, values, (err, rows) => {
```
sidorares: This tests namedPlaceholders disabled in both initial config and local override, is that intended?

derN3rd: I wanted to see, if the tests in CI pass as well, because they should fail and didn't on my system. So looks like they don't actually get verified, or am I wrong?

sidorares: oh, ok, I'll try to investigate

derN3rd: at first I thought this might be a problem with the pool, but it seems to broken one test above as well

sidorares: Looks like utest is finishing before any commands even start, and it says there is no planned support for async

I guess none of the tests in that file ever worked. One easy way to fix is to just cange to simple run + assert similar to other tests ( this is starting to become messy, I'd like to refactor all tests one day and migrate so some modern test runner like jest/mocha or ava )

derN3rd: So should I remove/fix the test again or do you want to merge this anyways?

sidorares: If you have time would be great if you can rewrite that test. If not let me know and I'll merge pr and fix the test myself

derN3rd: Don't have the time to fix the whole test setup right now, but I could help with jest in the future.
But would be nice if you could merge this pr until then ^^

derN3rd:
Do you have a fixed release cycle or is it possible to release this soon?

PR #1025, opened 2019-10-02
Fix promise methods

njh7799:
Filter out callback functions that went wrong as input argument in  methods of PromiseConnection and PromisePool.

When **query** & **execute** methods in PromiseConnection & PromisePool get **callback function** as the second argument(which is invalid), it is filtered so that **callback function** cannot be passed to query & execute  method in **connection** & **pool**

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -90,9 +90,14 @@ class PromiseConnection extends EventEmitter {
   query(query, params) {
     const c = this.connection;
     const localErr = new Error();
+    if (typeof params === 'function') {
+      throw new Error(
+        'Callback function is not available with promise clients.'
+      );
+    }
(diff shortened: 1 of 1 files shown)


sidorares:
I fell it might be better to error if function is passed as parameter with a message like "you are using callback function together with promise client, please don't"

Just silently ignoring them is not great ( but I agree that it's better than trying to interpret as a parameter )

I guess currently this code

```js
const [[result]] = await conn.query('select ?', (err, row) => { console.log(row) });
 ```

would assign source of callback function to `result` variable which might be a very unexpected behaviour

njh7799:
@sidorares Oh sure. I changed method to throw error when receiving a callback function :)

Issue #1029, opened 2019-10-10
[question RE: proxy example] unable to select database


warren-bank:
Hi. Great library!

The following output is produced by a [proxy server](https://github.com/warren-bank/node-mysql-proxy) which uses your wonderful library as both a server and a client (to a remote MariaDB database).

[This test](https://github.com/warren-bank/node-mysql-proxy/tree/master/.tests/3.%20misc/mysql2%20proxy%20example%20-%20mysql%20client%20issue%20-%20use%20db) fully confirms that the observed behavior isn't the result of anything that I added, since the same occurs in the [proxy server example](https://github.com/sidorares/node-mysql2/blob/master/examples/mysqlproxy.js) as taken directly from your repo.

Example from the `mysql` command-line client connected to the proxy:
```bash
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MySQL connection id is 1
Server version: mysql-proxy mariadb.org binary distribution

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MySQL [(none)]> show databases;
+--------------------+
| Database           |
+--------------------+
| information_schema |
| mysql              |
| mysql_proxy_test   |
| performance_schema |
| test               |
+--------------------+
5 rows in set (0.005 sec)

MySQL [(none)]> use mysql_proxy_test;
Database changed
MySQL [mysql_proxy_test]> select database();
+------------+
| database() |
+------------+
| NULL       |
+------------+
1 row in set (0.014 sec)

MySQL [mysql_proxy_test]> show tables;
ERROR 1064 (_____): No database selected
MySQL [mysql_proxy_test]>
```

The fundamental issue is how the `mysql` client executes "USE db_name;" statements. Apparently, this command isn't sent to the proxy and, consequently, is not forwarded to the remote db connection. Instead, this connection sees: "SELECT DATABASE()".

I can fully confirm that when the "USE db_name;" query is directly passed to the remote db connection.. it works. The problem is that the `mysql` client doesn't communicate this query.

Do you think this is a bug in the `mysql` client?

sidorares:
[...]

PR #1030, opened 2019-10-11
[#1029] fix/prevent calls to "packet.readString()" w/ wrong signature

warren-bank:
there were no other calls to `packet.readString(len, encoding)` that forgot to pass a `len` parameter.

matches

We didn't discuss my change to `packet.readString()` that detects this error and automagically fixes it. I'll remove it if you dislike.


sidorares:
Thanks @warren-bank !

PR #1036, opened 2019-10-16
Add dotenv and compose file

emauricio:
To make easier local testing, instead of exporting variables use a common `.env` file
if there are new environment variables that are possible to use we can list them on the `.env.example`

and in case the contributor doesn’t have a MySQL DB a docker-compose file is added to have
- MySQL
- [adminer](https://www.adminer.org/) (optional)

This PR should not affect or change anything on the code and can be used even without the .env file.


emauricio:
🤔 
image

@sidorares I think that is not related to the `-r dotenv/config` but I will try to confirm

PR #1052, opened 2019-11-06
Add missing authPlugins assignment in ConnectionConfig

hsjoberg:
Hi!
Using `authPlugins` as an option in `createConnection` does not seem to work.
In `connection_config.js`, you get a warning as it is not in the `validOptions` object and is also not assigned to the class further down in the constructor.

In this PR, this is fixed so that `auth_switch.js` can correctly utilize the plugin as it would otherwise not be able to because `connection.config.authPlugins[pluginName]` would be undefined.

I am not sure how `authPlugins` and the new API is supposed to work, but using this code at least works. Please tell me if it is supposed to used in another way.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -6,6 +6,7 @@ const Charsets = require('./constants/charsets');
 let SSLProfiles = null;
 
 const validOptions = {
+  authPlugins: 1,
   authSwitchHandler: 1,
   bigNumberStrings: 1,
   charset: 1,
@@ -138,6 +139,7 @@ class ConnectionConfig {
       ? ConnectionConfig.getCharsetNumber(options.charset)
       : options.charsetNumber || Charsets.UTF8MB4_UNICODE_CI;
     this.compress = options.compress || false;
+    this.authPlugins = options.authPlugins;
     this.authSwitchHandler = options.authSwitchHandler;
     this.clientFlags = ConnectionConfig.mergeFlags(
       ConnectionConfig.getDefaultFlags(options),


sidorares:
Main focus for authPlugins was to initially handle standard auth types in a way that is easy to extend, not sure I even tried examples with custom plugins

Your change looks good to me, I'll have a bit closer look tomorrow

How did you notice this issue? What's your use case and why are you trying to pass plugin? There is definitely a need for good documentation for plugin implementers

hsjoberg:
> Main focus for authPlugins was to initially handle standard auth types in a way that is easy to extend, not sure I even tried examples with custom plugins

Gotcha. Yes I saw the two standard plugins available already. The code suggested that it should be possible to pass in your own plugins when creating the connection, hence this PR.

> How did you notice this issue? What's your use case and why are you trying to pass plugin?

We are using "passwordless" AWS RDS login. This requires the `mysql_clear_password` client-side plugin to be available.

When upgrading to 2.0, we got an deprecation warning as we were still using the old `authSwitchHandler` to handle the `mysql_clear_password` plugin:

> WARNING! authSwitchHandler api is deprecated, please use new authPlugins api

To be clear, this is the code we used before to be able to handle RDS passwordless login:

```Typescript
authSwitchHandler(data: any, cb: any) {
  if (data.pluginName === "mysql_clear_password") {
    cb(null, Buffer.from(token + "\0"));
  }
},
```

and now with this PR, `authPlugins` can be used like this:

```Typescript
authPlugins: {
  mysql_clear_password: () => () => {
    return signer.getAuthToken({
      region,
      hostname,
      port,
      username
    });
  }
},
```

sidorares:
Gotcha

Yes this is probably the main scenario how now deprecated `authSwitchHandler` was used, and your `authPlugins` version is correct

sidorares:
Any help improving `authPlugins` docs would be happily accepted :)

PR #1065, opened 2019-11-20
Fixes #1063: PoolCluster#end now accepts a callback function

### LINKED ISSUE #1063: PoolCluster#end is not taking callback
[mysql2](https://github.com/sidorares/node-mysql2/blob/7dab7dfa9753631081f0f6d423ca6c6ec084bcec/lib/pool_cluster.js#L121) vs [mysqljs](https://github.com/mysqljs/mysql/blob/d62a7afc3c56350f5d7f95904cc4e6290d6513c6/lib/PoolCluster.js#L61)
PoolCluster.prototype.end doesn't accept callback

[TypeORM's disconnect](https://github.com/typeorm/typeorm/blob/c87e4e7ad92023a1914a6a1778d683ba8f23c5ab/src/driver/mysql/MysqlDriver.ts#L357) never resolves as a result of this

mintuhouse:


sidorares:
also related: #1170

PR #1074, opened 2019-12-03
Avoid setting numeric config options to NaN

### LINKED ISSUE #721: MYSQL2 silently dies on when a pool is initiated with incorrect connectionLimit values
I had the joy of debugging this for a while. But apparently the mysql2 library appears to have a logic error that will cause it to silently die.

I tracked this down to the use of `Number` to init the variable for the pool config (found at https://github.com/sidorares/node-mysql2/blob/5f0fb8f1f5035e2c0207490aa2f0b838dc82fdc2/lib/pool_config.js ). If for example the variable is a string, such as `"FILL-ME-IN"` this will initialise the value to `NaN`, which will then lead to a logic error further down in the applications pool logic.

I was considering making a patch changing the usage to `parseInt` but you might have some maintainers might have other comments or preferences. This would by default throw on an "unparsable" integer.

Alternatively further checks could be defined to verify that the connectionLimit doesn't end up as `NaN`, or is something satisfying `connectionLimit > -1` or similar.

Example code that will stop silently, this was tested using node 9.3.0 and mysql2 version 1.5.1:

```
// @ts-check
const mysql = require('mysql2/promise')

function getSourceDatabaseConfig() {
    const {
        S_MYSQL_HOST = '127.0.0.1',
        S_MYSQL_PORT = '3306',
        S_MYSQL_USER = 'root',
        S_MYSQL_PASSWORD = 'sometestpw',
        S_MYSQL_DATABASE = 'somedb',
        S_MYSQL_CONNECTIONS = 'FILL-ME-IN'
    } = process.env

    return {
        host: S_MYSQL_HOST,
        port: S_MYSQL_PORT,
        user: S_MYSQL_USER,
        password: S_MYSQL_PASSWORD,
        database: S_MYSQL_DATABASE,
        connectionLimit: S_MYSQL_CONNECTIONS
    }
}

let sourcePool;

function getSourcePool() {
    if (sourcePool === undefined) {
        sourcePool = mysql.createPool(getSourceDatabaseConfig());
    }

    return sourcePool;
}


async function run() {
    console.log('Executing Data-Mirror')

    const sourcePool = getSourcePool()

    console.log('Checking connections - Source pool')
    const connSource = await sourcePool.getConnection()
    console.log('Connection verified')
}


process.on('uncaughtException', err => {
    console.error(err)
    process.exit(1)
})
process.on('unhandledRejection', reason => {
    console.error(reason)
    process.exit(1)
})

process.on('SIGINT', function () {
    console.error('Got SIGINT')
    process.exit(0)
})
process.on('SIGTERM', function () {
    console.error('Got SIGTERM')
    process.exit(0)
})

run()
    .then(() => {
        console.log('Exiting')
    })
    .catch((e) => {
        console.log(`Got error from run`, e)
    })


```

jleedev:
Instead, treat NaN (as well as undefined) as an absent option and use
the default value.

Fixes #721 where a connectionLimit of NaN would render the pool silently
unable to acquire connections.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -82,8 +82,9 @@ class ConnectionConfig {
     this.password = options.password || undefined;
     this.passwordSha1 = options.passwordSha1 || undefined;
     this.database = options.database;
-    this.connectTimeout =
-      options.connectTimeout === undefined ? 10 * 1000 : options.connectTimeout;
+    this.connectTimeout = isNaN(options.connectTimeout)
+      ? 10 * 1000
+      : options.connectTimeout;
     this.insecureAuth = options.insecureAuth || false;
     this.supportBigNumbers = options.supportBigNumbers || false;
     this.bigNumberStrings = options.bigNumberStrings || false;
@@ -153,7 +154,9 @@ class ConnectionConfig {
     let flags = 0x0,
       i;
     if (!Array.isArray(user_flags)) {
-      user_flags = String(user_flags || '').toUpperCase().split(/\s*,+\s*/);
+      user_flags = String(user_flags || '')
+        .toUpperCase()
+        .split(/\s*,+\s*/);
     }
     // add default flags unless "blacklisted"
     for (i in default_flags) {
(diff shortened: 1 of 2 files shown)


sidorares:
Thanks @jleedev !

PR #1076, opened 2019-12-05
Pooling connections destroy

saravjotsingh:
Created a config property `idleTime` to destroy the pool connection.If not set then it will work as default.


perry-mitchell:
I personally think this logic would be better handled by _passively_ checking for idle time. Track the last timestamp a query completed, and when the pool is checking for a connection, have it check the idle time.

sidorares:
@perry-mitchell that means if a node had a spike of load and opened 10 db connections but then there is no traffic it'll keep 10 connections open. For mysql a connection usually means full os thread with stack memory allocation and other resources that are unavailable until connection is closed

Issue #907, opened 2019-01-14
Mysql2 Pool takes a 10-minute break after inquiring for information. The second inquiry will be reported to read ECONNRESET.


wadecha:
![image](https://user-images.githubusercontent.com/10524259/51119964-0d87d700-184f-11e9-9dcb-53fa1bda36ba.png)

**Test**
![image](https://user-images.githubusercontent.com/10524259/51119990-1aa4c600-184f-11e9-83f3-d86242b89c68.png)
![image](https://user-images.githubusercontent.com/10524259/51120015-285a4b80-184f-11e9-8656-b0886b00a549.png)

**results of execution**

first query, Pool value:
![image](https://user-images.githubusercontent.com/10524259/51120096-53449f80-184f-11e9-992d-e0b19e3a3c7b.png)

Execute the second time without rest:
![image](https://user-images.githubusercontent.com/10524259/51120133-65264280-184f-11e9-836b-4ed3905a7ac6.png)

Take a 10-minute break and execute again:
![image](https://user-images.githubusercontent.com/10524259/51120165-7707e580-184f-11e9-8ca9-10a3b0eacc09.png)

**Other:**
const mysqlConfig = {
host: ‘localhost’,
user: ‘root’,
password: ‘root’,
database: ‘my_db’,
port: 3306,
};

mysql version：5.6.31                 
use mysql + co running the same code is OK，No database settings have been made, wait_timeout is 28800

anthonywebb:
I am having the same issue with mysql 8.0.13 @wadecha

wadecha:
@sidorares  Can you help us ?

sidorares:
It feels like for whatever reason idle connections are not removed from pool when server closes them after `wait_timeout`. Can you check this line is called after 10 minutes? https://github.com/sidorares/node-mysql2/blob/821b2bcdc254c483e6879b5d4afc9757c55c7ed7/lib/connection.js#L94

Maybe not with pool but with one single persistent connection:

```js
const conn = mysql.createConnection(config);
conn.on('error', err => {
   console.log('Connection error:', err'); // this line is expected to be called when server terminates connection
});
conn.query('select 1+1');
```

Issue #897, opened 2018-12-11
Random ECONNRESET when talking to docker mysql


TomYeoman:
Hi all,

I have a big issue where I randomly get ECONNRESET every 1-5 minutes when using mysql / mysql2 npm package. ( Works fine up until this point )

**Environment**
 
- `mysql server` running in docker, mapped to host port 3306
- `node server` running within docker, talking to mysql via host_ip:3306

**Errot**

```
data-server-container |   Error: read ECONNRESET
data-server-container |
data-server-container |   - net.js:622 TCP.onread
data-server-container |     net.js:622:25
```

**Fix attempts**

I've tried all sorts to try and get the connection to remain - including

### Using pools

```

const mysql = require('mysql2');
const connection = await mysql.createPool({
  connectionLimit: 10,
  host: process.env.DB_HOST_DOCKER,
  user: process.env.DB_USER_DOCKER,
  password: process.env.DB_PASSWORD_DOCKER,
  database: process.env.DB_DATABASE,
});

query = (sql) => {
  return new Promise(async (resolve, reject) => {
    connection.getConnection(function (err, poolConnection) {
      if (err) throw err; // not connected!

      // Use the connection
      poolConnection.query(sql, function (error, results, fields) {

        if (error) reject(error)

        resolve(results);

        // When done with the connection, release it.
        poolConnection.release();

      })

    })
  })
}

```

### Creating a new connection on every request! ( Somehow this still gets ECONNRESET! )

```
query = (sql) => {

  const mysql = require('mysql2');

  const connection = await mysql.createConnection({
    host: process.env.DB_HOST_DOCKER,
    user: process.env.DB_USER_DOCKER,
    password: process.env.DB_PASSWORD_DOCKER,
    database: process.env.DB_DATABASE,
  });
  
  return new Promise(async (resolve, reject) => {
    connection.query(sql, (error, results) => {
      if (error) {
        reject(error);
      }
      resolve(results);
    });
  }

}
```

* Sending a ping to the connection every 10 seconds
* Checking my wait_timeout is high enough on mysql server ( 28800 as expected )

I'm running out of ideas here it's stopping me being able to use this in production :(

My best assumption after trying all the above is there an issue with the TCP Keep-Alive on the socket perhaps that is only triggered when running  node / mysql within docker ( I stumbled on this MR - could this be a solution? https://github.com/mysqljs/mysql/pull/2110/  )

eagle7410:
I have the same problems.
[...]

Issue #683, opened 2017-11-20
Pool cannot detect idle connection disconnected by MySQL and will trigger `read ECONNRESET` on newly acquired connection


t83714:
MySQL will disconnect idle connections after certain time frame. 

This time frame is determined by MySQL variable `wait_timeout`.  

It seems mysql2's pool implementation cannot detect this type of disconnection (ping is not sufficient).

And when it happens, any queries sent through the acquired connection will trigger a `read ECONNRESET` error.

At this moment, I have to send a `SELECT 1` query to test the acquired connection before use.

And previous code that calls pool.query directly will have to be re-written.

Can Mysql2 provides two extra pool creation options:
* verifyConnectionInterval: number of seconds before pool should send `SELECT 1` query to test connections in the pool.
* verifyConnection:  boolean. If set `true`, the pool will test a connection in the pool every `verifyConnectionInterval` seconds.

sidorares:
I saw this problem as well, and I think there is no way to automatically detect "server closed connection" automatically other than doing some kind of heartbeat messages

Are you sure doing `.ping(callback)` is not enough? Probably not much difference compared to `SELECT 1`

sidorares:
I think first step would be adding all events supported by mysqljs/mysql pool - https://github.com/mysqljs/mysql/#release

That way you can start/stop heartbeat in your code manually in reponse to `release`/`acquire` events

And later we could simplify that and make available as pool option

t83714:
Thanks for your reply.
Will run more tests on using `.ping(callback)` for heartbeat test and get it back to you.

t83714:
Hi @sidorares ,
Had a few more tests on this. 

It seems I can't get `.ping(callback)` work for heartbeat test purpose but `SELECT 1` works well.

Here are the code examples (I used [generic-pool](https://www.npmjs.com/package/generic-pool) for pool management):

Working Example (use `SELECT 1`):
```
import mysql2 from "mysql2/promise";
import genericPool from "generic-pool";
import dbConfigInfo from "../config/db.local";

const opts = {
    host: dbConfigInfo.host,
    user: dbConfigInfo.username,
    password: dbConfigInfo.password,
    database: dbConfigInfo.database,
};

const PoolDB = genericPool.createPool({
    create : () => mysql2.createConnection(opts),
    destroy : (connection) => connection.end(),
    validate : (connection) => connection.query(`SELECT 1`).then(()=>true,()=>false),
}, {
    max : 5,
[...]

Issue #128, opened 2014-11-06
ECONNRESET


kevinmartin:
I get these errors sometimes. Randomly.

```
    Error: read ECONNRESET 
        at exports._errnoException (util.js:742:11) 
        at TCP.onread (net.js:535:26)
```

and 

```
    Error: write ECONNRESET 
        at exports._errnoException (util.js:742:11) 
        at Object.afterWrite (net.js:749:14) 
```

I read somewhere that this also happens on `node-mysql`, but that it could be fixed by using pool. I began using pool, but it still happens.

Suggestions?

jsbinette:
I get this error too but only when I use koa-proxy which depends on co-request.  When I hit the post directly on the (VPN) server, I don't have the problem.
```
Error: read ECONNRESET
      at exports._errnoException (util.js:742:11)
      at TCP.onread (net.js:535:26)
```
The code I'm using is very simple

```
router.post('/api/Produit/:method',  proxy({
  host: "http://192.168.XXX.X"
}))
```

sidorares:
@jsbinette - is that new connection can't be established to mysql server or existing drops? Do you have local mysql server localhost:3306 ? Which version of node-mysql2 are you using?

jsbinette:
@sidorares Sorry turns out it doesn't apply.   In my case, koa-better-body interfered somehow with the post when it went thru the proxy.

farzher:
getting this sometimes randomly. simple test server, nothing complex happening on it.

```
{ Error: read ECONNRESET
    at PromisePool.query (./node_modules/mysql2/promise.js:323:22)
    at Object.select (./lib/db.js:36:44)
    at <anonymous>
  message: 'read ECONNRESET',
  code: 'ECONNRESET',
  errno: 'ECONNRESET',
  sqlState: undefined,
  sqlMessage: undefined }

sidorares:
@farzher could be anything - server idle timeout, network issues etc

farzher:
@sidorares you don't seem surprised that this causes random errors on production servers.
shouldn't the library do some kind of error catching retries before throwing an error?
or is everyone supposed to write custom wrapper code to do that themselves? (i missed the memo)
or is computer software just supposed to have some errors?

sidorares:
@farzher this is a low level driver and retries out of the scope. It's not clear from your report if you can catch and process error in your code or it's something that your code can't handle ( if latter, it definitely need to be fixed )

[...]

PR #1081, opened 2019-12-13
Keep alive

benbotto:
This PR allows the user to enable keep-alive on the underlying socket with a configurable `initialDelay` parameter, in milliseconds.  Keep-alive is disabled by default.  Example usage:

```js
  host: 'db',
  user: process.env.MYSQL_USER,
  password: process.env.MYSQL_PASSWORD,
  database: process.env.MYSQL_DATABASE,
  connectionLimit: 2,
  waitForConnections: true,
  queueLimit: 0,
  keepAliveInitialDelay: 10000, // 0 by default.
  enableKeepAlive: true, // false by default.
});

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -43,6 +43,10 @@ class Connection extends EventEmitter {
     } else {
       this.stream = opts.config.stream;
     }
+
+    // Enable keep-alive on the socket.
+    this.stream.setKeepAlive(true, 10000);
+
     this._internalId = _connectionId++;
     this._commands = new Queue();
     this._command = null;
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -44,8 +44,10 @@ class Connection extends EventEmitter {
       this.stream = opts.config.stream;
     }
 
-    // Enable keep-alive on the socket.
-    this.stream.setKeepAlive(true, 10000);
+    // Enable keep-alive on the socket.  It's disabled by default, but the
+    // user can enable it and supply an initial delay.
+    if (typeof this.stream.setKeepAlive === 'function' && this.config.enableKeepAlive)
+      this.stream.setKeepAlive(true, this.config.keepAliveInitialDelay);
 
     this._internalId = _connectionId++;
     this._commands = new Queue();
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -18,10 +18,12 @@ const validOptions = {
   dateStrings: 1,
   debug: 1,
   decimalNumbers: 1,
+  enableKeepAlive: 1,
   flags: 1,
   host: 1,
   insecureAuth: 1,
   isServer: 1,
+  keepAliveInitialDelay: 1,
   localAddress: 1,
   maxPreparedStatements: 1,
   multipleStatements: 1,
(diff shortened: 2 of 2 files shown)


sidorares:
I wonder if it worth making keepAlive on by default. If it won't add perf degradation and regression overall it'll potentially make life easier for many
At the same time default off is safer and we can change this behaviour later

sidorares:
Also if it's default off and not mentioned in read me there is 0 chance it'll be used anywhere in prod

benbotto:
It's up to you.  I can update the PR if you want, but I wanted to err on the side of caution.

benbotto:
There are some other issues that talk about ECONNRESET.

https://github.com/sidorares/node-mysql2/issues/897
https://github.com/sidorares/node-mysql2/issues/907
https://github.com/sidorares/node-mysql2/issues/128

In the first one the reporter is using Docker Swarm, so I'm pretty sure it's the same issue I was running up against.

At any rate, I wonder if there are some other members of the community that could test this change by enabling keep-alive manually using the new options in this PR.  Then the option could be enabled by default if it's viewed as generally beneficial.

sidorares:
Yeah, let's soft launch it first and I'll just advise all similar reports to try keepAlive option, later we can document and enable on by default

midnightcodr:
@sidorares can you up the npm version so the `enableKeepAlive` and `keepAliveInitialDelay` options can be used? Thanks.

sidorares:
after some thought - maybe `.setKeepAlive()` should have been added only when we create tcp/pipe connection instead of detecting if method is available on stream (after https://github.com/sidorares/node-mysql2/blob/f9b9f96c2cf1dace5570d60a89caacec2ecb4b48/lib/connection.js#L39 line).

benbotto:
I mentioned this as an aside over in #683, but I'm not up to speed on the four branches in the code referenced above ([lines 31-45](https://github.com/sidorares/node-mysql2/blob/f9b9f96c2cf1dace5570d60a89caacec2ecb4b48/lib/connection.js#L31-L45)).  There are 4 ways that `this.stream` gets set:

1. connecting over a unix socket;
2. connection via TCP;
3. a user-supplied stream;
4. or a factory function to produce a stream.

Is my understanding of (3) and (4) accurate?  And in those cases, since the user is creating the stream, is it essentially up to them to enable keep-alive if they want to use it?

sidorares:
@benbotto yes, all correct

sidorares:
option 4 was added later when I realised that pool needs a way to know how to create new underlying connection and it was too late to remove option 3

benbotto:
Cool, thanks for the clarification.  I made the changed and added a new PR: https://github.com/sidorares/node-mysql2/pull/1086

jsumners:
FYI: documentation should have been updated in this PR.

DigitalSolomon:
> Yeah, let's soft launch it first and I'll just advise all similar reports to try keepAlive option, later we can document and enable on by default

This deserves to be on the front page of the documentation. Took a lot of traveling down various rabbit holes to come upon this relatively simple solution. Would be great if this showed up on https://www.npmjs.com/package/mysql2 as well.

Kennytian:
`SET GLOBAL wait_timeout = 60; `

run this command in mysql client command terminal, the timeout unit is seconds

vlasky:
@benbotto couldn't we just have the single configuration option **keepAliveInitialDelay** which enables keepalive when it is set to a non-zero value? What additional benefit is there for having a separate **enableKeepAlive** configuration option?

Issue #908, opened 2019-01-14
IN() statement seems broken in execute()


Ngob:
Hello,

I am currently working with your package (version 1.6.1) and it is amazing (really love it). I am not sure if this intended or not:

I am doing the following:

mysql.js
```
export const mExecute = (function constructExecute() {
  let pool = null;
  let retry = 0;
  return async function tryExecute(...args) {
    if (!pool) {
      pool = constructPool();
    }
      const res = await pool.execute(...args);
      return res;
  };
}());
```

And :

App.js

```
mExecute(`SELECT tableA.name, 
            FROM tableA INNER JOIN tableB ON tableA.id_tableB=tableB.id AND tableA.active=1
            WHERE tableA.id IN (?)`, [[92, 93]]);
```

I would expect this code to work but it does not. If I change the code of mExecute with 

```
      const query = pool.format(...args);
      const res = await pool.execute(query);
```

instead of ` const res = await pool.execute(...args);` it is working as expected.

Is this intended or is it a bug?

Thanks,

PR #1105, opened 2020-01-18
fixed #476

### LINKED ISSUE #476: Problem using execute and array parameters
Hi,

mysql2 is a good alternative to mysql, but there is a little bug I guess.

I try to prepare a statement with an array parameter, that should be turned into a list. It works with `query` but not with `execute`.

```javascript

const ids = [1, 2, 3];
connection.query('SELECT id FROM mytable WHERE id IN (?)', [ids]);
// that's okay, I get 3 rows
[
  TextRow { id: 1 },
  TextRow { id: 2 },
  TextRow { id: 3 }
]

connection.execute('SELECT id FROM mytable WHERE id IN (?)', [ids]);
// I get only 1 row
[
  TextRow { id: 1 }
]

It seems that only the first element of the array is taken into account.

```
I'm using mysql2 v1.1.2.

Leask:
Fixed issue #476: https://github.com/sidorares/node-mysql2/issues/476#issuecomment-264342349

This is a confusing issue, and I know that the two-step execution can greatly improve performance. But in cases where the statement contains an `IN (?)` this will cause some unexpected code failures. Many known issues are related to this.

In my opinion, the correctness of code is more important than performance. I suggest that when the SQL command contains `IN (?)`, the formatter should be used first to format the command. Although this will lead to additional preparation processes, this simple modification can ensure that the code execution meets the expectations of most engineers in most cases.

Thanks for your excellent work!

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -564,6 +564,10 @@ class Connection extends EventEmitter {
       options.sql = sql;
       options.values = values;
     }
+    if (/in.*\(.*\)/i.test(options.sql)) {
+      options.sql = this.format(options.sql, options.values);
+      options.values = undefined;
+    }
     this._resolveNamedPlaceholders(options);
     // check for values containing undefined
     if (options.values) {


melroy89:
It's now 2024, is there any update on this PR?

wellwelwel:
It's been a long time, but thanks @Leask 🤝

Regardless the performance, this _PR_ needs a rebase to trigger all current tests. Also, it would be good to have additional tests covering these changes 🙋🏻‍♂️

> In my opinion, the correctness of code is more important than performance.

About performance, I believe that @sidorares can handle this discussion better.

Leask:
Hi all! It's cool to be mentioned. It's been a long time. But if you guys can fix this, it will be excellent.

wellwelwel:
Based in https://github.com/sidorares/node-mysql2/issues/908#issuecomment-454168035, I suggest waiting for a conclusion from @sidorares before working on it 🙋🏻‍♂️

> Note:
> 
> Screenshot 2024-04-07 at 17 21 24

Leask:
👌

melroy89:
Also consider updating the PR title, I believe the title is not helping either :P

Leask:
> Also consider updating the PR title, I believe the title is not helping either :P

Done 🤝

sidorares:
I'm open to discuss better DX here in general ( especially more helpful errors suggesting the fix ). However there is so many things that can potentially go wrong if we try to extract information from sql using a simple regex. Comments, strings, and many more scenarios can make the test incorrect.

Happy to continue discussion, but my vision in general:
 - if logic requires understanding of sql structure, this logic needs full sql parser and should probably live somewhere in a higher level library using this driver ( ORM / routers etc )
 - it's ok to have some code that tries to guess intention at an error handling level "did you mean field X" / "If the intention is to interpolate IN parameters, use this approach"

jacobbogers:
Question:

to bind vars is it enough for the mysql server to parse he SQL and return bind variables and their type, is it possible to explicitly specify the time during the bind process?

PR #1108, opened 2020-01-23
Release connection on exception

jafl:
If a synchronous exception is thrown (primarily by `execute()`), then the connection must be released.  I also made the same changes in #938


sidorares:
thanks @jafl !

PR #1118, opened 2020-02-27
Process dots feature

f8k8:
This is to enabled the `processDots` config option, which allows you to do things like:
`` SELECT field1, field2 AS `obj1.fieldA`, field3 AS `obj1.fieldB` ``
and get the results:
```
{
  field1,
  obj1: {
    fieldA,
    fieldB
  }
}
```


f8k8:
As mentioned in #1118

sidorares:
Sorry, but I don't feel functionality like this should belong to the driver. Happy to see it in some other library - orm or query builder. I'll keep PR open for some time to see if anyone else has feedback

In short, I'd rather see a function in separate module "expand dots" and apply it like `const [rawRow] = conn.query(sql); console.log(expandDots(row))` in user code rather having support in the driver with all the extra complexity (more code, documentation, support) in exchange for maybe a little bit better ergonomics for small number of users

Thanks for the effort though! ( and happy to listen for other opinions and maybe change mine )

Issue #1129, opened 2020-03-14
Incompatible when single replacement character


codpot:
Hi, First of all, thank you for developing mysql2.
I'm having trouble when passing parameter with only 0.
I thought it was because 0 was treated as false.

```js
// lib/connection.js#L500
const rawSql = this.format(cmdQuery.sql, cmdQuery.values || []);
```

When cmdQuery.values = 0
0 || [] goes to [] not 0

Also having same problem on 'promise.js'.
If you don't mind, I'll make PR for this issue.
Thank you.

```js
// when using mysql
const mysql = require('mysql');
const connection = mysql.createConnection({
  host: 'localhost',
  user: 'user',
  password: 'password',
  database: 'database'
});
connection.query('SELECT ? AS result', 0, (err, results) => {
  if (err) throw err;
  console.log(results);
  connection.end();
});
```
```
[ RowDataPacket { result: 0 } ]
```
```js
// when using mysql2
const mysql = require('mysql2');
const connection = mysql.createConnection({
  host: 'localhost',
  user: 'user',
  password: 'password',
  database: 'database'
});
connection.query('SELECT ? AS result', 0, (err, results) => {
  if (err) throw err;
  console.log(results);
  connection.end();
});
```
```
  if (err) throw err;
           ^

Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '? AS result' at line 1
    at Packet.asError (C:\Users\user\Workspace\code\node_modules\[4mmysql2[24m\lib\packets\packet.js:712:17)
    at Query.execute (C:\Users\user\Workspace\code\node_modules\[4mmysql2[24m\lib\commands\command.js:28:26)
    at Connection.handlePacket (C:\Users\user\Workspace\code\node_modules\[4mmysql2[24m\lib\connection.js:417:32)
    at PacketParser.onPacket (C:\Users\user\Workspace\code\node_modules\[4mmysql2[24m\lib\connection.js:75:12)
    at PacketParser.executeStart (C:\Users\user\Workspace\code\node_modules\[4mmysql2[24m\lib\packet_parser.js:75:16)
    at Socket.<anonymous> (C:\Users\user\Workspace\code\node_modules\[4mmysql2[24m\lib\connection.js:82:25)
[90m    at Socket.emit (events.js:315:20)[39m
[90m    at addChunk (_stream_readable.js:297:12)[39m
[90m    at readableAddChunk (_stream_readable.js:273:9)[39m
[...]

PR #1130, opened 2020-03-14
Fix incompatibility when zero parameter (#1129)

codpot:
There was an error when passing replace parameter with 0.

0 !== undefined

0 || [] goes to [] not 0
if(0) goes to else


sidorares:
Thanks @codpot !

PR #1133, opened 2020-03-25
PoolCluster support restoreNodeTimeout option

anchel:
make PoolCluster support restoreNodeTimeout option

### DIFF
--- a/lib/pool_cluster.js
+++ b/lib/pool_cluster.js
@@ -4,6 +4,19 @@ const Pool = require('./pool.js');
 const PoolConfig = require('./pool_config.js');
 const EventEmitter = require('events').EventEmitter;
 
+function getMonotonicMilliseconds() {
+  let ms;
+
+  if (typeof process.hrtime === 'function') {
+    ms = process.hrtime();
+    ms = ms[0] * 1e3 + ms[1] * 1e-6;
+  } else {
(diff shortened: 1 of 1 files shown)


sidorares:
Hey @anchel thanks for submitting this PR! Would be really good to add some unit test for this option, can you take a look how `mysqljs/mysql` is doing this? You'll need to tweak test setup since `FakeServer` is different here, but as a starting point you can use https://github.com/mysqljs/mysql/blob/ad014c82b2cbaf47acae1cc39e5533d3cb6eb882/test/unit/pool-cluster/test-query-restore.js and https://github.com/mysqljs/mysql/blob/ad014c82b2cbaf47acae1cc39e5533d3cb6eb882/test/unit/pool-cluster/test-restore.js

rushvora:
@sidorares What will it take to merge this PR? The missing unit tests like you mentioned above? Due to this option not working, if our MySQL 8 instance restarts, our application starts breaking, as we are using the pool cluster feature heavily.

gautiermorel:
Hi ! Thanks @rushvora for you PR ! Do we have a release planned for this ? It will be very useful to have `restoreNodeTimeout` back ! Thanks !

anchel:
等其他人提pr吧

Issue #1136, opened 2020-04-04
Connection.connect callback may never be executed


cressie176:
I've found a subtle inconsistency between mysql and mysql2...

[mysql](https://github.com/mysqljs/mysql#user-content-connection-options) separates the creation of a Connection object and the establishment of a database connection, i.e.
```js
var mysql = require('mysql');

// Create the connection object
var connection = mysql.createConnection({
  host     : 'example.org',
  user     : 'bob',
  password : 'secret'
});

// Actually establish a database connection
connection.connect(function(err) {
  if (err) {
    console.error('error connecting: ' + err.stack);
    return;
  }

  console.log('connected as id ' + connection.threadId);
});
```
However mysql2 both creates the Connection object and establishes the database connection in the [Connection constructor](https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L22-L129), emitting a [connect](https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L110) event when successful. 

mysql2 attempts to maintain compatibility with mysql by adding a [connect](https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L716-L735) function, which listens for the **connect** event and executes the callback, however this creates a potential race condition where if the **connect** event is emitted before `connection.connect(cb)` is called, the callback will never be executed. The following code demonstrates the problem...

```js
const mysql = require('mysql2');

const connection = mysql.createConnection({
  host: 'localhost',
  user: 'root',
});

connection.on('connect', () => {
  console.log('Connection established');

  connection.connect(() => {
    console.log('Connect callback executed'); // <------ NEVER REACHED
  });
})

setTimeout(() => {
  console.log('Timeout');
  process.exit(1);
}, 5000);
```

```
$ node index.js 
Connection established
Timeout
```

I discovered this problem while trying to switch [marv-mysql-driver](https://github.com/guidesmiths/marv-mysql-driver/blob/master/index.js#L30-L38) to mysql2. I can easily rewrite my code to not go async between initiating the connection and calling connect, but thought it was worth mentioning, as it could affect other applications, and could be fixed by executing the callback immediately if the connection was currently active, or with an error if the connection had already been closed.

sidorares:
good catch @cressie176 

[...]

PR #1137, opened 2020-04-04
Fix Connection.connect callback may never be executed (#1136)

cressie176:
I wasn't able to find a way to test _fatalError - it wasn't set when either destroying or closing the server connection, however _protocolError was set, so I was at least able to test this.

I suspect there will still be some cases where an error occurs by neither _fatalError or _protocolError are set. In this case `connection.connect` will pass a closed / invalid connection to the callback and any attempt to use this should result in an error, which I think is still acceptable.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -717,6 +717,12 @@ class Connection extends EventEmitter {
     if (!cb) {
       return;
     }
+    if (this._fatalError || this._protocolError) {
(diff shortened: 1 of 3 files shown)


sidorares:
thanks @cressie176 !

cressie176:
My pleasure. Thanks for a great library.

PR #1153, opened 2020-05-03
Implement sha256_password authentication plugin

### LINKED ISSUE #1152: Error: Server requests authentication using unknown plugin sha256_password
Started receiving this error from Ubuntu 20.04's mysql 8.0.19. My users are using `caching_sha2_password` and the server default is also set to that, however it's still requesting `sha256_password`. This may be a bug in Ubuntu's mysql build, but I went ahead an made an implementation of the auth plugin for node-mysql2 and will post a PR.

```
Error: Server requests authentication using unknown plugin sha256_password. See TODO: add plugins doco here on how to configure or author authentication plugins.
    at Object.authSwitchRequest (node_modules/mysql2/lib/commands/auth_switch.js:46:11)
    at ClientHandshake.handshakeResult (node_modules/mysql2/lib/commands/client_handshake.js:150:22)
    at ClientHandshake.execute (node_modules/mysql2/lib/commands/command.js:39:22)
    at Connection.handlePacket (node_modules/mysql2/lib/connection.js:417:32)
```

kellycampbell:
Fixes #1152

### DIFF
--- a/lib/auth_plugins/sha256_password.js
+++ b/lib/auth_plugins/sha256_password.js
@@ -1,0 +1,60 @@
+'use strict';
+
+const PLUGIN_NAME = 'sha256_password';
+const crypto = require('crypto');
+const { xor } = require('../auth_41');
+
+const REQUEST_SERVER_KEY_PACKET = Buffer.from([1]);
+
+const STATE_INITIAL = 0;
+const STATE_WAIT_SERVER_KEY = 1;
+const STATE_FINAL = -1;
+
+function encrypt(password, scramble, key) {
+  const stage1 = xor(
+    Buffer.from(`${password}\0`, 'utf8').toString('binary'),
+    scramble.toString('binary')
+  );
+  return crypto.publicEncrypt(key, stage1);
+}
+
+module.exports = (pluginOptions = {}) => ({ connection }) => {
+  let state = 0;
+  let scramble = null;
+
+  const password = connection.config.password;
+
+  const authWithKey = serverKey => {
+    const _password = encrypt(password, scramble, serverKey);
+    state = STATE_FINAL;
+    return _password;
+  };
+
+  return data => {
+    switch (state) {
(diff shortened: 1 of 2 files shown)


sidorares:
I'd like to test it locally first, can you give instructions on how to set up a server with  sha256_password user?

kellycampbell:
```
create user 'sha256_test'@'%' identified with sha256_password by 'password';
```

https://dev.mysql.com/doc/refman/8.0/en/sha256-pluggable-authentication.html

sidorares:
given that the plugin is deprecated another option might be to ship this plugin as external module. This would require some extra setup for users that need it but won't increase size/complexity of the module for those who don't need sha256_password auth

kellycampbell:
I couldn't find the root cause in mysql but did see other similar reports. Even though it is deprecated, it's still in use and likely causing problems for users who need to connect to databases using it. Default Ubuntu 20.04 released a few weeks ago is where I'm experiencing it out of the box.

I found one team at my company using a workaround of making all the team's devs create their mysql users with `IDENTIFIED WITH mysql_native_password` added.

The `mysql` npm module has a similar problem but doesn't report it well. I can't tell from the error message or stack if it's failing on cached_sha2 or sha256. I don't think they implement either yet.
https://stackoverflow.com/questions/50093144/mysql-8-0-client-does-not-support-authentication-protocol-requested-by-server

The implementation is tiny and almost identical to the caching_sha2 one (because I copied most of the code and trimmed out the stuff that didn't apply). It would be nice if the library just worked out of the box for what seems to be a common issue.

sixmen:
I'm using sha256_password with MySQL 5.7. So I'm glad to include this plugin to the module.

(Currently, I'm using extracted plugin function and custom authPlugins settings)

sidorares:
I'm happy to include it @kellycampbell @sixmen , sorry didn't have much time lately to work through backlog of PRs

PR #1164, opened 2020-06-04
Move require to top level of the module

cabbiepete:
Conditionally importing does not seem to be required (constants only).

When running tests in parallel with Jest and a mysql2 connection we receive warnings as below

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

      at new ResultSetHeader (node_modules/mysql2/lib/packets/resultset_header.js:39:32)
      at Query.resultsetHeader (node_modules/mysql2/lib/commands/query.js:102:16)
      at Query.execute (node_modules/mysql2/lib/commands/command.js:39:22)
      at Connection.handlePacket (node_modules/mysql2/lib/connection.js:417:32)
      at PacketParser.onPacket (node_modules/mysql2/lib/connection.js:75:12)
```

If the require statement is moved to the top of the file as per this commit the issue is resolved.

### DIFF
--- a/lib/packets/resultset_header.js
+++ b/lib/packets/resultset_header.js
@@ -8,6 +8,7 @@ const ClientConstants = require('../constants/client.js');
 const ServerSatusFlags = require('../constants/server_status.js');
 
 const EncodingToCharset = require('../constants/encoding_charset.js');
+const sessionInfoTypes = require('../constants/session_track.js');
 
 class ResultSetHeader {
   constructor(packet, connection) {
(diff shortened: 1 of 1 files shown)


sidorares:
it saves a bit of startup time if libraries you use try to require inly the subset of the code you actually need when you need it, but I guess in this case few ms of benefit not worth trouble ( Jest / webpack / other bundlers )

@cabbiepete is `require('../constants/session_track.js')` the only one that caused issues for you? Jest problems were reported few times in the past ( some related to a sub dependency - iconv also requires files lazily )

cabbiepete:
@sidorares thanks for the reply.

I found those reports relating to iconv when investigating this one but have not had the same exact issues related to iconv.

Interesting that was the reason for doing that. For my use case definitely a few ms on startup not a priority, I suppose in a server less environment perhaps it's higher priority my thinking would be the connection establishment would be a lot longer than a JS code import.

You're definitely the expert for this package but would be great to get this change in. Let me know if you'd like any more changes to make this PR easier to merge.

sidorares:
> connection establishment would be a lot longer than a JS code import.

yes, definitely

By itself it's a small number, but if your app uses 100s of similar libraries this can make a difference of restart time to be 200ms or 2s

PR #1170, opened 2020-06-19
Pass in the callback when ending the pool connection.

vikasgarghb:
This fixes an issue when you close a connection using replication, the `end()` method on pool do not resolve the promise coz of missing callback.


martzcodes:
Any updates on this? would be useful for a project I'm working on

PR #1181, opened 2020-07-21
Fix readme

Squarific:
createPool already returns a promised wrapped pool


Squarific:
Oh nvm, this is only if you include the Promise version.

Issue #99, opened 2014-04-30
dateStrings not implemented. 


JemiloII:
``` js
mysql.createConnection({
    host        : 'localhost',
    user        : 'root',
    password    : 'pass',
    database    : 'dbname',
    dateStrings : 'DATE' //this does nothing
});
```

I tried the other dateStrings commands and they didn't work. So I just installed the original and the dateStrings method works just fine.

kbsanders:
I really only want DATE fields to be treated as strings but not DATETIME fields. I see that [mysqljs/mysql](https://github.com/mysqljs/mysql) now supports an array of types for `dateStrings` in addition to a boolean value. Will we see the same option added to mysql2?

In their [Connection Options](https://github.com/mysqljs/mysql#connection-options):

> dateStrings: Force date types (TIMESTAMP, DATETIME, DATE) to be returned as strings rather than inflated into JavaScript Date objects. Can be true/false **or an array of type names** to keep as strings. (Default: false)

sidorares:
wow, I missed that this was added to mysqljs/mysql

it's relatively simple change - would you like to try adding it @kbsanders ? Let me know if you need any help navigating the code

kbsanders:
@sidorares Thanks for the quick reply. If I find time in the next few days I will submit a PR.

In the meantime, I'm using this workaround of providing a `typeCast` in the options:

```js
// NOTE: only works for .query() calls and not .execute() calls.
typeCast: function ( field, next ) {

    if ( field.type === "DATE" ) {
        return field.string();
    }
    
    return next();
}
```

Along the way I discovered that `typeCast` was only being called for `.query()` and not `.execute()` so I had to refactor my code to use `.query()`. I don't know if this is a limitation of the way prepared statements work or not, but I guess that is a topic for a new issue.

sidorares:
> I don't know if this is a limitation of the way prepared statements work or not, but I guess that is a topic for a new issue.

see discussion in https://github.com/sidorares/node-mysql2/issues/649 and https://github.com/sidorares/node-mysql2/issues/717

PR #1200, opened 2020-09-12
feat: allow using `dateStrings` with specific types

B4nan:
This aligns the `dateStrings` option with mysqljs, allowing to set `dateStrings: ['DATE']` to return
the DATE type as string, but keeping other datetime types processed to Date.

Related: #99

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -48,3 +48,16 @@ function printDebugWithCode(msg, code) {
 }
 
 exports.printDebugWithCode = printDebugWithCode;
+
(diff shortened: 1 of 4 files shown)


sidorares:
Thanks @B4nan !

B4nan:
Do you have a rough ETA for a release? I am making some performance optimisations in MikroORM that required this change, and would rather not ship changes that depend on unreleased version :]

sidorares:
@B4nan your change is published now

PR #1206, opened 2020-09-18
Add the authPlugins types to ConnectionOptions

kanyuga:
My builds started failing today with error 

> TS2339: Property 'authPlugins' does not exist on type 'ConnectionOptions'. 

I've copied the `authPlugins` type declaration from @types/mysql2


sidorares:
thanks! published as v2.2.2

PR #1215, opened 2020-09-23
Pool should extend mysql.Pool, not mysql.Connection

fluffynuts:
The current `index.d.ts` shows `Pool` extending `mysql.Connection`, which leads to faulty typings -- in particular, a `Pool` does not have transaction methods (`beginTransaction`, `commit`, `rollback`) where a `Connection` _does_

It's a minor correction, but I just spent time wondering why my code was complaining about missing methods at runtime when the typings suggested they should be there ):

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -144,6 +144,7 @@ export interface Pool extends mysql.Pool {
   getConnection(
     callback: (err: NodeJS.ErrnoException, connection: PoolConnection) => any
   ): void;
+  releaseConnection(connection: PoolConnection): void;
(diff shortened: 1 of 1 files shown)


dctrotz:
@fluffynuts  - any chance you can also add `releaseConnection(connection: PoolConnection): void;` to the `Pool` declarations? Otherwise I will add it myself and PR it but these seem to go together.

fluffynuts:
@dctrotz sorry, just seen this now -- added the definition as per your request (:

fluffynuts:
@dctrotz is this still useful, or should I close it out?

fluffynuts:
closing due to inactivity

Issue #1216, opened 2020-09-23
ERR_PACKAGE_PATH_NOT_EXPORTED with version > 2.1.0


haganbt:
If I install `mysql2@2.2.5` and run the following:

```js
"use strict";

const mysql = require("mysql2");
const ClientFlags = require("mysql2/lib/constants/client.js");
```

It throws:

```
internal/modules/cjs/loader.js:518
  throw new ERR_PACKAGE_PATH_NOT_EXPORTED(basePath, mappingKey);
  ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/constants/client.js' is not defined by "exports" in /Users/VisualStudio/node_modules/mysql2package.json
```

With `mysql2@2.1.0` there is no error. 

Not sure if its the cause but note the missing slash at the end of the error: `node_modules/mysql2package.json`

pubmikeb:
I have the same issue but I'm using `mysql2/promise.js`:
```js
import mysql from "mysql2/promise.js";
```
Defining a pool and executing the query:
```
const pool = await mysql.createPool({
	connectionLimit: CONFIG_ENV.dbConnLimit,
	database: CONFIG_ENV.dbName,
	host: CONFIG_ENV.dbHost,
	password: CONFIG_ENV.dbPass,
	queueLimit: 0,
	user: CONFIG_ENV.dbUser,
	waitForConnections: true
});

await pool.query(QUERIES.myQuery, […]);
```

When I run my app, I get the following error:

> (Use `node --trace-warnings ...` to show where the warning was created)
> internal/process/esm_loader.js:74
>     internalBinding('errors').triggerUncaughtException(
>                               ^
> Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './promise.js' is not defined by "exports" in C:\Users\User\IdeaP
> rojects\TMS\node_modules\mysql2\package.json imported from C:\Users\User\IdeaProjects\TMS\src\server\bin\DAL\DALDriver.m
> js
>     at new NodeError (internal/errors.js:253:15)
>     at throwExportsNotFound (internal/modules/esm/resolve.js:290:9)
>     at packageExportsResolve (internal/modules/esm/resolve.js:513:3)
>     at packageResolve (internal/modules/esm/resolve.js:644:14)
>     at moduleResolve (internal/modules/esm/resolve.js:696:18)
[...]

PR #1217, opened 2020-09-24
Add 'promise.js' to exports

chrisveness:
Prior to 2.2.0, an ES6 import of the promise wrapper had to specify a file
extension ("import mysql from 'mysql2/promise.js';"), which made #1100 a
breaking change as only the extensionless version was included in the exports;
this restores promise.js as a valid import.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
   ],
   "exports": {
     ".": "./index.js",
-    "./promise": "./promise.js"
+    "./promise": "./promise.js",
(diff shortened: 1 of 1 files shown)


sidorares:
Thanks!

pubmikeb:
@chrisveness, so what will be the right way to import mysql2 with promises support?
Currently there is a related issue: https://github.com/sidorares/node-mysql2/issues/1216.

PR #1222, opened 2020-10-03
mysql_clear_password can have string type

tomoima525:
- mysql_clear_password in authPlugins currently only supports
Promise<string> type. However AWS.RDS.Signer.getAuthToken returns string, which
causes type error "Type 'string' is not assignable"

getAuthToken Type
Screen Shot 2020-10-03 at 11 23 46 AM

Error message:
Screen Shot 2020-10-03 at 11 27 26 AM

- This fix will add support for string type as well

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -154,7 +154,9 @@ export interface Pool extends mysql.Connection {
 type authPlugins = (pluginMetadata: {
   connection: Connection;
   command: string;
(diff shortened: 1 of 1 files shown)


sidorares:
Thanks! Can you also add `Buffer | Promise<Buffer> | null` as possible return type for a plugin?

tomoima525:
@sidorares I see, sure! 3526352

PR #1235, opened 2020-10-26
Mysql cluster 8.0.22 + mysql2 bug of collation

Perlover:
In my project i found very strange bug after upgrade `mysql-cluster` 8.0.21 -> 8.0.22 `mysql-cluster` package (`Ubuntu 18.04`)
My mysql statements throwed SQL errors "Illegal mix of collations"
I researched all docs of Mysql. There is [Rerertoire](https://dev.mysql.com/doc/refman/8.0/en/charset-repertoire.html) term. 
This bug manifests itself as if this mechanism does not work in mysql-cluster 8.0.22. But i cannot reproduced by hand through `mysql` command. 
This gave me the idea that the problem lies somewhere either in your module, or not in compatibility due to a change in mysql-cluster 8.0.22. I wrote a test for this problem and i was able to reproduce the problem. What is most interesting, it manifests itself only in `execute` (not `query`) and when the table has a second field. So i think it's prepared statement related problem.

I got error in tests:

```
[0:00:18 0 41/108 38.0% node test/integration/connection/test-execute-bind-ascii.js]    
  
  /home/perlover/tmp/node-mysql2/test/integration/connection/test-execute-bind-ascii.js:21
        throw err;
        ^
  
  Error: Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8mb4_unicode_ci,COERCIBLE) for operation '='
      at Packet.asError (/home/perlover/tmp/node-mysql2/lib/packets/packet.js:712:17)
      at Execute.execute (/home/perlover/tmp/node-mysql2/lib/commands/command.js:28:26)
      at Connection.handlePacket (/home/perlover/tmp/node-mysql2/lib/connection.js:425:32)
      at PacketParser.onPacket (/home/perlover/tmp/node-mysql2/lib/connection.js:75:12)
      at PacketParser.executeStart (/home/perlover/tmp/node-mysql2/lib/packet_parser.js:75:16)
      at Socket.<anonymous> (/home/perlover/tmp/node-mysql2/lib/connection.js:82:25)
      at Socket.emit (events.js:314:20)
      at addChunk (_stream_readable.js:298:12)
      at readableAddChunk (_stream_readable.js:273:9)
      at Socket.Readable.push (_stream_readable.js:214:10) {
    code: 'ER_CANT_AGGREGATE_2COLLATIONS',
    errno: 1267,
    sqlState: 'HY000',
    sqlMessage: "Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8mb4_unicode_ci,COERCIBLE) for operation '='"
  }
```

### DIFF
--- a/test/integration/connection/test-execute-bind-ascii.js
+++ b/test/integration/connection/test-execute-bind-ascii.js
@@ -15,15 +15,6 @@ connection.query(`INSERT INTO ${table} (a, b) VALUES ('test', 1)`);
 connection.query(`INSERT INTO ${table} (a, b) VALUES ('super', 2)`);
 connection.execute(
   `SELECT a FROM ${table} WHERE a=? AND b=?`,
-  ['DE', 1],
-  err => {
-    if (err) {
-      throw err;
-    }
-  }
-);
-connection.execute(
-  `SELECT a FROM ${table} WHERE a=? AND b=?`,
   ['test', 1],
   (err, _rows) => {
     if (err) {
--- a/test/integration/connection/test-execute-bind-ascii.js
+++ b/test/integration/connection/test-execute-bind-ascii.js
@@ -16,7 +16,7 @@ connection.query(`INSERT INTO ${table} (a, b) VALUES ('super', 2)`);
 connection.execute(
   `SELECT a FROM ${table} WHERE a=? AND b=?`,
   ['DE', 1],
-  (err) => {
+  err => {
     if (err) {
       throw err;
     }
--- a/test/integration/connection/test-execute-bind-ascii.js
+++ b/test/integration/connection/test-execute-bind-ascii.js
@@ -1,0 +1,39 @@
+'use strict';
+
+const common = require('../../common');
+const connection = common.createConnection();
+const assert = require('assert');
+
+const table = 'asciitest';
+
+let rows;
+connection.query(`DROP TABLE IF EXISTS ${table}`);
+connection.query(
+  `CREATE TABLE IF NOT EXISTS ${table} (a char (64) ascii not null, b int unsigned not null) character set utf8`
+);
+connection.query(`INSERT INTO ${table} (a, b) VALUES ('test', 1)`);
+connection.query(`INSERT INTO ${table} (a, b) VALUES ('super', 2)`);
+connection.execute(
+  `SELECT a FROM ${table} WHERE a=? AND b=?`,
+  ['DE', 1],
+  (err) => {
+    if (err) {
+      throw err;
+    }
+  }
+);
+connection.execute(
+  `SELECT a FROM ${table} WHERE a=? AND b=?`,
+  ['test', 1],
+  (err, _rows) => {
+    if (err) {
+      throw err;
+    }
+    rows = _rows;
+    connection.end();
+  }
+);
+
+process.on('exit', () => {
+  assert.deepEqual(rows, [{ a: 'test' }]);
+});


Perlover:
I deleted the first `execute` in test - it was not important test statement
So last commit is improved version of test
I got there error: 

```[0:00:17 0 39/108 36.1% node test/integration/connection/test-execute-bind-ascii.js]    
  
  /home/perlover/tmp/node-mysql2/test/integration/connection/test-execute-bind-ascii.js:21
        throw err;
        ^
  
  Error: Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8mb4_unicode_ci,COERCIBLE) for operation '='
      at Packet.asError (/home/perlover/tmp/node-mysql2/lib/packets/packet.js:712:17)
      at Execute.execute (/home/perlover/tmp/node-mysql2/lib/commands/command.js:28:26)
      at Connection.handlePacket (/home/perlover/tmp/node-mysql2/lib/connection.js:425:32)
      at PacketParser.onPacket (/home/perlover/tmp/node-mysql2/lib/connection.js:75:12)
      at PacketParser.executeStart (/home/perlover/tmp/node-mysql2/lib/packet_parser.js:75:16)
      at Socket.<anonymous> (/home/perlover/tmp/node-mysql2/lib/connection.js:82:25)
      at Socket.emit (events.js:314:20)
      at addChunk (_stream_readable.js:298:12)
      at readableAddChunk (_stream_readable.js:273:9)
      at Socket.Readable.push (_stream_readable.js:214:10) {
    code: 'ER_CANT_AGGREGATE_2COLLATIONS',
    errno: 1267,
    sqlState: 'HY000',
    sqlMessage: "Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8mb4_unicode_ci,COERCIBLE) for operation '='"
  }
  
```

Perlover:
I checked my packages - i have mysql-server 8.0.22 not mysql-cluster

And i posted the bug here too:

https://bugs.mysql.com/bug.php?id=101346

I don't know which side the problem is on (mysql2 module or mysql server 8.0.22) so I decided to inform both the design of the problem.

sidorares:
to me looks like it's server error and driver is just returning something server is unhappy but I haven't investigated that in detail. Thanks for providing repro test and if you definitely know its server issue ( and there is no easy way to help avoiding of better messaging on the driver side ) let us know @Perlover

Perlover:
> to me looks like it's server error and driver is just returning something server is unhappy but I haven't investigated that in detail. Thanks for providing repro test and if you definitely know its server issue ( and there is no easy way to help avoiding of better messaging on the driver side ) let us know @Perlover

Most interestingly, the server [developers have reproduced the problem](https://bugs.mysql.com/bug.php?id=101346#c505904), but [they believe](https://bugs.mysql.com/bug.php?id=101346#c505925) it is a problem with your module :). Here is their answer:

> Just one update to understand things properly
> 
> - this connector is community project, not MySQL made
> - this connector uses classic mysql protocol but native nodejs, not using mysqlclient library
> - the MySQL's nodejs connector uses X protocol only
>
> Using other mysql clients we cannot reproduce this, so it is def. a bug in the node-mysql2 connector (not our project). We do want to help and will try to find out why this fails with .22 but the mysql as is works flawlesly.

And I noticed that this error happens in the "execute" but a "query" methods.
I think this problem is there near by prepare statements with placeholders. May be your side may be server side (protocol?)

Perlover:
May be [this will help](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-22.html#mysqld-8-0-22-optimizer) (changes from 8.0.21 to 8.0.22):

> Important Change: A prepared statement is now prepared only once, when executing PREPARE, rather than once each time it is executed. In addition, a statement inside a stored procedure is also now prepared only once, when the stored procedure is first executed. This change enhances performance of such statements, since it avoids the added cost of repeated preparation and rollback of preparation structures, the latter being the source of several bugs.

ruiquelhas:
@sidorares this seems to be related to the server-side changes regarding prepared statements (as mentioned by @Perlover). In any case, I'm not able to reproduce it using our other non-`libmysqlclient` and, as far as I can tell, there were no changes in the client-side to accommodate this new behaviour, which means it might be an issue with how this module actually implements the protocol. 

I'm digging a little bit into it. I'll try to submit a PR or point you to a specific fix.

ruiquelhas:
Apparently I was fooled by some client charset setting on my tests. This seems to be reproducible with other client implementations, including the official `mysql` CLI.

We are looking into it and will keep in touch.

ruiquelhas:
This has been verified as a [bug](https://bugs.mysql.com/bug.php?id=101346) in the MySQL server.

sidorares:
@Perlover I'm closing this for now as it was identified as a server issue but if you can think of anything that would help driver users encountering this problem please feel free to reopen

PR #1251, opened 2020-11-25
Allow unnamed placeholders even if the namedPlaceholders flag is enabled

oamaok:
In a project using this package we are trying to transition into using named placeholders. We would like to do this gradually, as some parts of the code cannot be easily transitioned. This change allows the use of unnamed placeholders even when the flag is enabled.

### DIFF
--- a/documentation/Extras.md
+++ b/documentation/Extras.md
@@ -2,7 +2,7 @@
 
 ## Named placeholders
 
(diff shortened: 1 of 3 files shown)


sidorares:
LGTM. I know it's a very small change but could you add a unit test to cover this? Also maybe update documentation here mentioning fallback to unnamed - https://github.com/sidorares/node-mysql2/blob/master/documentation/Extras.md#named-placeholders

oamaok:
All should be good now. I also renamed the `test-named-placeholders.js` test file as there was a typo.

sidorares:
thanks @oamaok !

Issue #1229, opened 2020-10-15
Pool option "enableKeepAlive" is ignored?


davidecaroselli:
Hi all!

I was experiencing several issues with Pool in production due to `read ECONNRESET` error and I finally found the response in this Github Issue: https://github.com/sidorares/node-mysql2/issues/683 (and, by the way, thanks to all for the amazing work done and the patience in debugging it! ;) )

So, I have updated from `2.0.1` to `2.2.5` (latest right now), but I found something strange. Looking at the code it seems that the new PoolConfig property `enableKeepAlive` is never actually queried.
I found this piece of code that I think should read this property:
https://github.com/sidorares/node-mysql2/blob/2be0eb6298bca5528749bcf4b0609052373974b3/lib/connection.js#L42

But as you can see the function is always called with a hardcoded `true`.
Now, this is fine with me, because I was about to enable it anyway, but my question is: is this intentional? Am I missing some obvious bit of code?

Thanks!

sidorares:
hm, looks like this is side effect of #1086 . see https://github.com/sidorares/node-mysql2/pull/1081#issuecomment-566607840 for reasons why it was added but to me looks like it should be moved inside `if (this.config.enableKeepAlive) {}` block

kn3ny:
I have the same doubt as @davidecaroselli during I was reading the code.
I believe the pull request https://github.com/sidorares/node-mysql2/pull/1258 will fix this issue.

vlasky:
@sidorares why can't we just have a single configuration option keepAliveInitialDelay that enables keepalive when it is set to a non-zero value? 

What additional benefit is there for having a separate enableKeepAlive configuration option?

PR #1258, opened 2020-12-06
Follow enableKeepAlive option

kn3ny:
It seems `enableKeepAlive` option is ignored since the setting for enabling keep-alive is hardcoded to "true".
https://github.com/sidorares/node-mysql2/blob/a640d471f043eb078c09ab0d6016030c315bd879/lib/connection.js#L42

This pull request changes the behavior: enabling TCP keep-alive when `enableKeepAlive` is not set, or `enableKeepAlive` is set to true. The `enableKeepAlive` is **true by default** even before this change so I also edit comments in the source code.

Related issue:
- https://github.com/sidorares/node-mysql2/issues/1229

### DIFF
--- a/typings/mysql/lib/Pool.d.ts
+++ b/typings/mysql/lib/Pool.d.ts
@@ -33,8 +33,7 @@ declare namespace Pool {
         queueLimit?: number;
 
         /**
-         * Enable keep-alive on the socket.  It's disabled by default, but the
-         * user can enable it and supply an initial delay.
+         * Enable keep-alive on the socket. It's enabled by default.
          */
         enableKeepAlive?: true;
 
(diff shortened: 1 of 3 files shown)


typings/mysql/lib/Pool.d.ts
```diff
@@ -33,8 +33,7 @@ declare namespace Pool {
         queueLimit?: number;
 
         /**
-         * Enable keep-alive on the socket.  It's disabled by default, but the
-         * user can enable it and supply an initial delay.
+         * Enable keep-alive on the socket. It's enabled by default.
          */
         enableKeepAlive?: true;
```
irepela, line 38: should be boolean instead of true

kn3ny: @irepela Thanks!

kn3ny:
I'll re-create pull request later.

PR #1260, opened 2020-12-08
exposing pool cluster in the promise-based implementations

alpacamybags118:
Originally my changes were to expose query as a function of PoolCluster (as it was in the mysqljs library - https://github.com/mysqljs/mysql/blob/3430c513d6b0ca279fb7c79b210c9301e9315658/lib/PoolNamespace.js#L58)

In order to achieve that, I had to expose PoolCluster in the promise based implementation. This exposes PoolCluster in promise.js, along with adding the types in the promise typing files.

This is my first time working on the code for this library, so any feedback would be appreciated! I tried to follow the same coding practices and styles as much as I could.

### DIFF
--- a/examples/promise-co-await/await.js
+++ b/examples/promise-co-await/await.js
@@ -56,13 +56,30 @@ async function test() {
   end = +new Date();
   console.log(end - start);
   await p.end();
+
+  const cluster = mysql.createPoolCluster();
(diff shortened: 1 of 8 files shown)


vlasky:
It would be great to have execute() as well.

alpacamybags118:
added execute as well!

PR #1261, opened 2020-12-11
Remove unused option acquireTimeout from typings

yonran:
Remove parameter that apparently has never actually been supported

If you try including `acquireTimeout` in the PoolOptions from the TypeScript interface, then `mysql2.createPool` logs this warning (from https://github.com/sidorares/node-mysql2/pull/815): `Ignoring invalid configuration option passed to Connection: acquireTimeout. This is currently a warning, but in future versions of MySQL2, an error will be thrown if you pass an invalid configuration option to a Connection`.


wellwelwel:
Currently `acquireTimeout` is no longer present in the typings.

Related:

- #3099
- #2095

PR #1267, opened 2020-12-20
fix the type of ResultSetHeader **insertId** attribute type

buuug7:
database table ID either **number** or **string**


testn:
Nope. LastInsertId has to be number.

https://dev.mysql.com/doc/internals/en/packet-OK_Packet.html

sidorares:
when `bigNumberStrings ` setting is set to true `insertId` is returned as string ( when it can't fit js number )

testn:
ah I forgot

sidorares:
I'm not a fan of this polymorphic results but nut quite sure how to improve ( and in backwards compatible way )

we can probably retire Long.js and return native BigInt here so the type will be `number | BigInt` ( still polymorphic )

PR #1288, opened 2021-02-15
Typing and readme docs for rowsAsArray

rathboma:
I love love love the `rowsAsArray` query / connection option.

However it took me a while to find it, so hoping this makes it easier for others to find.


sidorares:
thanks @rathboma !

Issue #1019, opened 2019-09-27
Packet.readDateTime cannot read length of undefined


Mattchewone:
Upgrading from `1.6.3` to `1.7.0` and I am hitting this error:

```
node_modules/mysql2/lib/packets/packet.js:286
    if (str.length === 10) {
            ^

TypeError: Cannot read property 'length' of undefined
    at Packet.readDateTime (/user/node_modules/mysql2/lib/packets/packet.js:286:13)
    at new BinaryRow (eval at line.toFunction (/user/node_modules/generate-function/index.js:172:21), <anonymous>:2431:41)
    at Execute.row (/user/node_modules/mysql2/lib/commands/query.js:230:17)
    at Execute.execute (/user/node_modules/mysql2/lib/commands/command.js:39:22)
    at Connection.handlePacket (/user/node_modules/mysql2/lib/connection.js:408:32)
    at PacketParser.Connection.packetParser.p [as onPacket] (/user/node_modules/mysql2/lib/connection.js:70:12)
    at PacketParser.executeStart (/user/node_modules/mysql2/lib/packet_parser.js:75:16)
    at PacketParser.executePayload (/user/node_modules/mysql2/lib/packet_parser.js:167:21)
    at Socket.Connection.stream.on.data (/user/node_modules/mysql2/lib/connection.js:77:25)
    at Socket.emit (events.js:198:13)
    at addChunk (_stream_readable.js:288:12)
    at readableAddChunk (_stream_readable.js:269:11)
    at Socket.Readable.push (_stream_readable.js:224:10)
    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
```

sidorares:
hi @Mattchewone can you reduce this to simple self contained example so I can run it on my side? Ideally include schema creation and initial data

grzegorziskierka:
I've replicated this error with table having default value for date or datetime columns e.g.

```
CREATE TABLE `user` (
  `userid` int(11) NOT NULL AUTO_INCREMENT,  
  `dob` date DEFAULT '0000-00-00'
  PRIMARY KEY (`userid`)
) ENGINE=InnoDB;
```
and when one or more records had default value set. This error was happening only when I was fetching records with limit applied. Work around was to pass 'dateStrings: true' in 'dialectOptions' property when creating connection to MySQL.

b2evandro:
[...]

PR #1305, opened 2021-03-17
fix(packet): fix issue #1019

metauro:
str is undefined when record include zero date on execute select statement


sidorares:
Superseded by #4258, which picks up this fix and adds an integration regression test.

Thank you @zhongyuanjia for the original fix!

sidorares:
Closing in favor of #4258.

Issue #1224, opened 2020-10-05
On Node 12.8 failing at runtime - missing ./lib/constants/types export


rbotla:
We are getting the following error on Node 12.8. 
```
[...]

Issue #1084, opened 2019-12-16
SSL support does not work with Node v12+


jjosef:
When using Node v12+ you will always get an error when trying to connect with SSL.
```
SSL routines:ssl_choose_client_version:unsupported protocol
```

The minimum TLS version in node 12 is 1.2, and this module needs to add support for this.

jjosef:
I am able to get this working using a connection string similar to this:
```
const connConfig = {
  user: 'username',
  password: 'password',
  host: 'hostname',
  ssl: {
    rejectUnauthorized: false,
    minVersion: 'TLSv1'
  }
}
```

However, this isn't ideal. I think it's worth making an effort to make this work properly.

sidorares:
does it fail with `minVersion: 'TLSv1'` but `rejectUnauthorized` unset?

related: #961 #960

jjosef:
Yes it does fail, btw this is using Amazon RDS. Came up with another solution that pulls the ca bundle from lib/constants/ssl_profiles.js

sidorares:
>  Came up with another solution

this one?
https://github.com/sidorares/node-mysql2/issues/960#issuecomment-562342922

@stephen-dahl did that work for you without setting `rejectUnauthorized` to false?

jjosef:
> > Came up with another solution
> 
> this one? [#960 (comment)](https://github.com/sidorares/node-mysql2/issues/960#issuecomment-562342922)
> 
> @stephen-dahl did that work for you without setting `rejectUnauthorized` to false?

Yes that's what we ended up doing. It's pretty hacky though. I can look into making it work properly with TLS 1.3 in node 12 if you don't have a plan to get his resolved

sidorares:
I don't have a plan yet. Do you thing RDS bundled configs need to be updated to include ` minVersion: 'TLSv1'`?

stephen-dahl:
I don't think it can be the default because it is only a problem with mysql < 5.6.46 and aurora mysql 5.6. Newer versions of mysql and aurora support the newer ssl versions. We don't want to make newer versions insecure to make older versions a little easier.

stephen-dahl:
We could add a new profile `Amazon RDS TLSv1`

jjosef:
Seems like a good idea, I think just adding some documentation on this for supporting older MySQL versions in newer node versions would suffice, and improving how to load the Amazon RDS cert.

jjosef:
[...]

PR #1313, opened 2021-04-07
Update index.js

karjaneth:
Adding other constants modules as top-level getters. This will resolve a number of open bugs, including

* https://github.com/sidorares/node-mysql2/issues/1224
* https://github.com/sidorares/node-mysql2/issues/1216
* https://github.com/sidorares/node-mysql2/issues/1084

In my own use case, we had been importing `mysql2/lib/constants/errors` in order to support retry-after-deadlock logic, which is now broken since the `exports` are defined in a more limited fashion in `package.json`. This will allow us to simply `import mysql from 'mysql2'` and use `mysql.Errors` for our situation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,15 +59,18 @@ exports.__defineGetter__(
   () => require('./promise.js').createPoolCluster
 );
 
+exports.__defineGetter__('CharsetToEncoding', () => require('./lib/constants/charset_encodings.js'));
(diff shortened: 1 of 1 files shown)


karjaneth:
@sidorares if you are OK with this change and merge it in, would you mind releasing a new version with it as soon as is reasonable? We are just hard-coding error codes until we can go back to importing them from `mysql2`.

PR #1338, opened 2021-06-09
Fix tests to work with the latest MySQL server versions

ruiquelhas:
The latest MySQL server versions (up to 8.0.25) introduced some changes that are causing some tests to fail. These changes are mostly related to the following:
  
  - new unexpected packets being sent from the server
  - deprecated column metadata details for certain data types such as integers
  - stricter SQL rules enforced on date and time values

Additional details about each issue being address are depicted in the corresponding commit message.

Although the changes are mostly constrained to the test suite, there is also a small change in the implementation in order to accommodate the introduction of an additional MySQL protocol packet sent by the server when the connection is closed because of inactivity.

In order to make the test suite more flexible and able to execute in different host and CI environments, for instance, where there is no way to access random ports on different hosts or containers, all instances of fake MySQL servers are now being spin in the same host where the tests are running. As far as I can tell, this causes no functional change to the tests themselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+# This file was modified by Oracle on June 14, 2021.
+# The changes involve the introduction of a local cache for Docker images.
+# Modifications copyright (c) 2021, Oracle and/or its affiliates.
+
 sudo: required
 dist: trusty
 
@@ -9,8 +13,18 @@ language: node_js
 cache:
   yarn: true
   directories:
+    - docker_images
     - node_modules
     - $HOME/.yarn-cache
+
+before_cache:
+  # save all docker images to a local cache in order to avoid the rate limit
+  # on Docker Hub
+  - docker save -o docker_images/images.tar $(docker images -a -q)
+
+before_install:
+  # load docker images from the local cache
+  - docker load -i docker_images/images.tar || true
 
 # Node.js version:
 # we test only maintained LTS versions
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1,3 +1,8 @@
+// This file was modified by Oracle on June 1, 2021.
+// The changes involve new logic to handle an additional ERR Packet sent by
+// the MySQL server when the connection is closed unexpectedly.
+// Modifications copyright (c) 2021, Oracle and/or its affiliates.
+
 'use strict';
 
 const Net = require('net');
@@ -188,7 +193,7 @@ class Connection extends EventEmitter {
     if (this.connectTimeout) {
       Timers.clearTimeout(this.connectTimeout);
       this.connectTimeout = null;
-    }    
+    }
     // prevent from emitting 'PROTOCOL_CONNECTION_LOST' after EPIPE or ECONNRESET
     if (this._fatalError) {
       return;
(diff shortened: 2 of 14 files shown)


ruiquelhas:
@sidorares looks like the Travis builds are failing because they are hitting the Docker Hub rate limit and are unable to spin the MySQL server container. I think the script will need to start manually [caching](https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-542672812) Docker images to avoid this.

License
```diff
@@ -17,3 +17,5 @@ Copyright (c) 2016 Andrey Sidorov (sidorares@yandex.ru) and contributors
  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  THE SOFTWARE.
+
+Copyright (c) 2021, Oracle and/or its affiliates.
```
sidorares, line 21: @ruiquelhas what's the reason this line is added? I'm not an IP lawyer and not aware of the consequences but on my side the intension is ti keep this library under MIT license https://opensource.org/licenses/MIT

ruiquelhas: In any case, let me double check the implications of this change and I'll let you know.

ruiquelhas: I'm still waiting to hear an official response from the horse's mouth but here's [something](https://softwareengineering.stackexchange.com/a/158011) I've found in the meantime. It's not exactly the same scenario since the goal is not to fork, but it mentions the following:

> To do so with the same MIT license: Just add your copyright notice after the 2012 Acme Corp copyright notice in the files you modified.

Which means that, probably, I just need to move the new copyright notice block and put it between the original copyright notice block and permission notice block.

dougwilson: @sidorares I think that if Oracle needs this done by their employees, you may need to provide a SLA for the employees. I would be especially concerned about having them explicitly agree that they are granting the use of any patents they may be contributing, even accidentally. My experience in OSS is that if an employer is forcing their employees to do this, you need to make them sign a SLA because they have a very leal mind :) A SLA is not bad, it just makes it explicit on whatis happening, especially since the MIT license would not grant patent usage if the employee (even accidentally) incorporated a company patent into the contribution.

dougwilson: Sorry, the above should say CLA, not SLA.

As far as copyright goes, the person who wrote the code automatically retains copyright already unless there is a CLA that reassigns it. I suspect the employer wants to retain copyright explicitly so that is why they are adding this line. Employer automatically retains copyright for works done as part of work, but clearly the employer wants to call it out, so if it were me I would make a CLA as I think the employer would like thay better anyway, as the agreement is clear.

ruiquelhas: My legal knowledge is pretty limited. What I can say is that yes, it's still Oracle's code, and they own the copyright but are granting the project the license to use it (under the same terms I assume).

With regards to patents, we are not allowed to submit contributions involve the use of material covered by any Oracle patents, issued or pending.

In any case, the CLA would be a possibility but it would also require additional burden on my end, and would probably void this contribution and delay any future one.

In any case, I'm actively trying to get to the bottom of this internally. I'm as interested and concerned as you guys.

ruiquelhas: Ok, I finally managed to get more details about the issue. Looks like this not the way to go either. What I should be doing instead is adding the copyright notice to each source file that is introduced or changed by the contribution. In case of an existing file, the notice should mention the modifications where it applies. The `License` file shouldn't, in the best case scenario, be changed.

This means there should be no interference with the project license. Again, although Oracle would own the copyright for the new changes, the project has the license to use it and (the project) retains whatever copyright it has for all the remaining code.

@sidorares does this work for you?

ruiquelhas:
@sidorares I've removed the copyright notice from the `License` file and added it on each source file that has been changed or introduced by the pull request. Please let me know if you still have any concerns about it and/or if you prefer to resort to a CLA.

PR #1352, opened 2021-07-05
Add proper handshake fatal error handling

### LINKED ISSUE #1263: Client-side authentication plugin errors not closing the connection
I was in the process of writing an authentication plugin to be used with the framework provided by mysql2, and I've realized that when a client-side plugin throws an error, the server connection is not closed.

```js
const mysql = require('mysql2')

  authPlugins: {
    caching_sha2_password () {
      throw new Error('boom')
    }
  },
  user: 'root' // with mysqld --initialize-insecure
})

connection.connect(err => {
  if (err) {
    console.error(err)
    return
  }

  console.log('done')
  connection.end()
})
```

Is this by design and are the plugins expected to close the connection themselves? In such a case, the only way the application can do it is by calling `process.exit()` or similar, otherwise if it calls `connection.end()` or `connection.destroy()` it will bump into an `Unexpected packet while no commands in the queue` error, which is somewhat noisy.

As far as I can tell, everytime the plugin "throws", you end up reaching the `handhakeResult` handler in `ClientHandshake` command.

https://github.com/sidorares/node-mysql2/blob/master/lib/commands/client_handshake.js#L157

The handler happens to emit an `error` event in the `connection` instance, for which, the event handler will only execute the callback with the error argument. However, the connection is already open by this point, and it does not close it.

https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L744

In the end, the fact that the client-side authentication plugin throws an error means that there was indeed an authentication error and there isn't really any way to recover from it unless creating a connection with different options. This should probably be no different than the case where the authentication error happens in the server and the server itself ends up closing the connection, allowing the client to also release resources. But we can disagree about this.

So, I won't claim this is a bug or a limitation, but rather an oversight or a design decision, and, if that is the case, is this something that can change?

ruiquelhas:
Unknown handshake errors such as ones caused by unexpected packets sent by the server and errors reported by client-side authentication plugins are now considered fatal (similar to what already happens for errors reported by the server during the authentication stage).

Additionally, any kind of fatal error reported during the handshake stage will result in the connection being closed.

Closes #1263.

### DIFF
--- a/lib/commands/client_handshake.js
+++ b/lib/commands/client_handshake.js
@@ -1,3 +1,8 @@
+// This file was modified by Oracle on June 17, 2021.
+// Handshake errors are now maked as fatal and the corresponding events are
+// emitted in the command instance itself.
+// Modifications copyright (c) 2021, Oracle and/or its affiliates.
+
 'use strict';
 
 const Command = require('./command.js');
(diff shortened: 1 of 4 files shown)


sidorares:
thanks @ruiquelhas !

PR #1355, opened 2021-07-07
Add getter function for connection private property _fatalError

alexsherman:
I ran into the problems described in https://github.com/knex/knex/pull/2175 recently and I saw that the proposed solution in this issue (expose _fatalError as part of the public api of connection -> use getter in knex) was never implemented. Here's a stab at exposing that property as a getter, and I'll follow up with a PR for knex if this is accepted.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -397,7 +397,7 @@ class Connection extends EventEmitter {
     this.emit('error', err);
   }
   
-  getFatalError() {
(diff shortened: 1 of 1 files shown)


sidorares:
hi @alexsherman can you add a simple unit test ( or modify existing one ) so your getter is not removed accidentally in the future

PR #1359, opened 2021-07-20
fix uncatchable exception thrown issue

shyunku:


Digicrat:
I was just about to create a new ticket when I found this.  Applying this patch looks to resolve my issue (see below).  I suspect this patch would also need to be applied to the query function, and possibly others.


My original message:

I'm executing a number of queries asynchronously, but when an error occurs I'm unable to catch the exception.  It causes the program to exit, without any of my error handlers being triggered. I just updated to the latest mysql2 version (2.3.2).

Neither of the following is working to catch an error ('Bind parameters cannot be undefined', or SQL reporting a field can't be null in another case), making it very difficult to debug.

`try { result = await db.execute(...); } catch(e) { ...block never gets called... }`
`result = await db.execute(...).catch(e => { ...also never gets called...}`

Is there a better way of doing this, or is this a bug (which doesn't preclude the former)? 

thanks

PR #1360, opened 2021-07-20
Remove calls to deprecated url.parse

### LINKED ISSUE #1357: Password, in URL connection string, that contain a colon are not supported
The connection code that parses `mysql://` strings does not handle passwords containing a colon `:`.

https://github.com/sidorares/node-mysql2/blob/9d22796c71bd0b4bb17f587cfdbd6f0bd25e7768/lib/connection_config.js#L244

* The URL is correctly parsed and any URI encoding is decoded correctly.
* The `auth` part of the URL is split on the `:` character, giving more than 2 parts if the password contains such a character.
* It should treat the first `:` as the only delimiter.

acdibble:
Fixes #1357.

`url.parse` and the legacy Node.js `Url` object were deprecated in Node 12.

By replacing it with the WHATWG URL API, username and password parsing comes for free. Additionally, uses of the `query` property had to be changed to use the `URLSearchParams` object.

### DIFF
--- a/benchmarks/FB/hello.js
+++ b/benchmarks/FB/hello.js
@@ -61,8 +61,8 @@ function sequelizeQuery(callback) {
 }
 
 function handlePrepared(req, res) {
-  const values = url.parse(req.url, true);
-  const queries = values.query.queries || 1;
+  const values = new url.URL(req.url);
+  const queries = values.searchParams.get('queries') || 1;
   const results = [];
   for (let i = 0; i < queries; ++i) {
     mysql2conn.execute(
--- a/benchmarks/http-select-and-render.js
+++ b/benchmarks/http-select-and-render.js
@@ -10,10 +10,10 @@ const port = process.env.PORT;
 
 http
   .createServer((req, res) => {
-    const q = url.parse(req.url, true);
+    const q = new url.URL(req.url);
     if (q.pathname === '/render') {
-      const sql = q.query.q;
-      const n = q.query.n;
+      const sql = q.searchParams.get('q');
+      const n = q.searchParams.get('n');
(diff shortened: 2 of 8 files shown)


acdibble:
@sidorares , I've fixed the parsing and added a new test to validate the behavior. Let me know if there is anything else I can do to help get this merged in.

sidorares:
@acdibble I think your editor config is slightly different to what's used here, can you discard commits unrelated to your main change ( replacing url.parse with URL )? appveyor build fails because of lint errors

acdibble:
@sidorares , I think one of the commit hooks changed some stuff. It seems like my local ESLint is in conflict with Appveyor's:
```
$ npx eslint .

~/node-mysql2/examples/promise-co-await/co.js
  17:9   error  Unexpected function expression  prefer-arrow-callback
  20:10  error  Unexpected function expression  prefer-arrow-callback

~/node-mysql2/examples/ssl/rds-ssl.js
  14:45  error  Unexpected function expression  prefer-arrow-callback

~/node-mysql2/examples/ssl/select-over-ssl.js
  19:34  error  Unexpected function expression  prefer-arrow-callback
  21:49  error  Unexpected function expression  prefer-arrow-callback

~/node-mysql2/tools/generate-charset-mapping.js
  33:30  error  Unexpected function expression   prefer-arrow-callback
  46:17  error  Unexpected string concatenation  prefer-template

✖ 7 problems (7 errors, 0 warnings)
  7 errors and 0 warnings potentially fixable with the `--fix` option.
```

I've reverted those changes so hopefully the build is green this time.

sidorares:
thanks @acdibble !

Issue #1365, opened 2021-07-29
TIME column returned in wrong format


martenjacobs:
As per the [mysql specification](https://dev.mysql.com/doc/refman/5.6/en/time.html), a TIME column is in `hh:mm:ss` format. However, when querying, what I see is that the actual format returned by node-mysql2 is `h:mm:ss` for amounts less than 10 hours. This makes it more difficult to parse (luxon's Duration.fromISOTime requires the leading zero, for example). 
I'd expect that either the documentation would clearly describe the return type and format for each column type, or that the library would follow the mysql spec, but in the case of the TIME column type, it appears that neither is the case.

sidorares:
looks like we need to add `leftPad(2, H)` here: https://github.com/sidorares/node-mysql2/blob/4e22e33495287c4c46336154ccfba01d3b23e9df/lib/packets/packet.js#L373

@martenjacobs would you like to contribute code change and a unit test to fix this?

martenjacobs:
@sidorares done, see #1366

PR #1366, opened 2021-07-30
Improved formatting for TIME-type columns

martenjacobs:
Solves #1365

This PR makes sure the default format for TIME columns adheres to the format outlined [here](https://dev.mysql.com/doc/refman/5.6/en/time.html): 
`[-][h]hh:mm:ss[.u[u[u[u[u[u]]]]]]`


sidorares:
thanks @martenjacobs !

PR #1372, opened 2021-08-06
add sqlMessage to QueryError interface

mxw:
it should be there; it's usually set. also delete redundant comments of the form `/** the variable name, but with spaces */`.

### DIFF
--- a/typings/mysql/lib/protocol/sequences/Query.d.ts
+++ b/typings/mysql/lib/protocol/sequences/Query.d.ts
@@ -79,19 +79,9 @@ declare namespace Query {
          */
         code: string;
 
-        /**
-         * The sql state marker
-         */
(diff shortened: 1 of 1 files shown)


sidorares:
thanks @mxw .
I'm not an expert in TS typings, would it be possible to somehow validate typings from unit tests? Currently if we accidentally break something it's not visible until somebody using this module with TS starts to complain

mxw:
i'm just barely aware that unit tests for typings exist, because i happened upon this today: https://github.com/discordjs/discord.js/blob/main/typings/tests.ts

but i'm not familiar with how they work or best practices; i'm a regular TS user but don't maintain any TS libraries myself.

sidorares:
I guess just running TS compiler against some codebase that uses typings would be a test

Issue #889, opened 2018-11-28
typeCast is up to 10x slower


gbelur:
When the typeCast option is used, queries tend to run up to 10x slower under load. A perf profile shows that most of the additional time is spent in column_defintion.js obtaining the schema and table name in the code below:

```
const addString = function(name) {
  Object.defineProperty(ColumnDefinition.prototype, name, {
    get: function() {
      const start = this['_' + name + 'Start'];
      const end = start + this['_' + name + 'Length'];
      return StringParser.decode(
        this._buf.slice(start, end),
        this.encoding === 'binary' ? this._clientEncoding : this.encoding
      );
    }
  });
};
```

These fields are meant to be computed lazily, but end up getting recomputed for every field access as part of the 'wrap' function implementation that needs field.schema, field.table and field.name

```
function wrap(field, type, packet, encoding) {
    return {
      type: type,
      length: field.columnLength,
      db: field.schema,
      table: field.table,
      name: field.name,
      string: function() {
        return packet.readLengthCodedString(encoding);
      },
      buffer: function() {
        return packet.readLengthCodedBuffer();
      },
      geometry: function() {
        return packet.parseGeometryValue();
      }
    };
  }

```
Given that these fields are used for every column, can these be precomputed?

testn:
@gbelur Can you try with 2.3.3-rc0 to see if it is ok now?

testn:
I thought I fixed this already @sidorares

sidorares:
I thought so as well
Need to test and close if can't confirm

wellwelwel:
Fixed in _PR_ #1400 (f2abaa89, released in v2.3.1). I don't see this test as useful for CI, but it helps confirm that all loadings properly handle lazy to close the issue with confidence:

```ts
import { Buffer } from 'node:buffer';
import { describe, it, strict } from 'poku';
import ColumnDefinition from '../../../lib/packets/column_definition.js';
import Packet from '../../../lib/packets/packet.js';

const lenenc = (value: string): Buffer => {
  const buf = Buffer.from(value, 'utf8');
  strict.ok(buf.length < 251);
  return Buffer.concat([Buffer.from([buf.length]), buf]);
};

const asPacket = (payload: Buffer): typeof Packet.prototype => {
  const buf = Buffer.concat([Buffer.alloc(4), payload]);
  return new Packet(0, buf, 0, buf.length);
};

const buildColumnDefinitionPayload = (): Buffer => {
[...]

PR #1400, opened 2021-10-06
Cache lazy-evaluated field

testn:
Fixing #889 

Reading the data from the buffer repeatedly.

Sequelize Before Change
Call to doSomething took 2571.7481999993324 milliseconds

Sequelize After Change
Call to doSomething took 1230.2330999970436 milliseconds


sidorares:
do you think the getter definition can be reduced to just one `defineProperty` call @testn ? Happy to see this merged, the reason I initially added lazy evaluated access was that buffer => string conversion was relatively high in the profiler results so caching it makes total sense

sidorares:
side note - I'm currently in the process of migration travis-ci to github actions powered CI, would like to finish that first before merging any other feature

testn:
the first defineProperty is to make it lazy. Once the property is evaluated for the first time, the property is modified again to cache the value. 

I tested different mechanism to cache the value including just caching the value like

  get: function() {
      const cacheVal = this[`_${name}Cache`];
      if (cacheVal) {
          return cacheVal;
      }
      const start = this[`_${name}Start`];
      const end = start + this[`_${name}Length`];
      return StringParser.decode(
        this._buf.slice(start, end),
        this.encoding === 'binary' ? this._clientEncoding : this.encoding
      );
}

but it is much slower since the hot method cannot inline the property value in addition to the cost of string concat and dictionary lookup everytime.

PR #1401, opened 2021-10-07
Optimize string decoding by removing the use of slice()

testn:
Previously, when we want to decode a string from a range in a Buffer, we
chose to slice it first and then decode it. However, Buffer.toString()
already does it internally already. In addition to that, toString()
seems to be more efficient when working with native buffer rather than
the wrapper FastBuffer created by slice()

From a simple test with large buffer and short string, it seems to yield
about 10-15% performance improvement


sidorares:
@testn this one needs rebasing again to resolve conflicts

PR #1402, opened 2021-10-07
Avoid leaking TextRow/BinaryRow object outside the framework

testn:
Avoid leaking TextRow/BinaryRow object outside the framework by
returning a new object every time instead of returning this

### DIFF
--- a/lib/parsers/text_parser.js
+++ b/lib/parsers/text_parser.js
@@ -95,7 +95,7 @@ function compile(fields, options, config) {
     parserFn('const _this = this;');
     for(let i=0; i<fields.length; ++i) {
       const field = fields[i];
-      const encodingExpr = `fields[${i}].encoding`;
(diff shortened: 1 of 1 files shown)


testn:
re: accessing every field, sequelize currently copies every field from TextRow object into its own object. Therefore, every field is already accessed.

sidorares:
I'll try to benchmark locally, also wonder how reliable numbers are when run inside GH action

sidorares:
FYI @testn - I added benchmark workflow in https://github.com/sidorares/node-mysql2/pull/1406 but as far as I understand the way it works it compares benchmarks from current run to the previous one and I see big variations in results even with no changes in the code. I'll see if I can change it ( instead of saving bench results in the cache for later access, always perform 2 benchmarks in the same run, one for master and one for current commit ). If performance alert comments become too noisy I'll disable them

Issue #1239, opened 2020-10-29
Incorrect arguments to mysqld_stmt_execute


perry-mitchell:
Hi, love this library!

Have been using it successfully across a large number of projects and it's worked flawlessly for me so far. Just now I've seen that I have a single test case failing on the CI server (Gitlab), MySQL 8, failing with the following error:

```
VError: Failed executing query: \"SELECT * FROM message WHERE chat_id = ? ORDER BY sent DESC LIMIT ? OFFSET ?\" [01ensmg6m4dea0gh7gsjgaa5gb, 50, 0]: Incorrect arguments to mysqld_stmt_execute
    at ConnectionPool.<anonymous> (/data/node_modules/@my-company/mysql-connection-pool/dist/ConnectionPool.js:128:27)
    at Generator.throw (<anonymous>)
    at rejected (/data/node_modules/@my-company/mysql-connection-pool/dist/ConnectionPool.js:6:65)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
caused by: Error: Incorrect arguments to mysqld_stmt_execute
    at PromiseConnection.execute (/data/node_modules/mysql2/promise.js:110:22)
    at ConnectionPool.<anonymous> (/data/node_modules/@my-company/mysql-connection-pool/dist/ConnectionPool.js:117:58)
    at Generator.next (<anonymous>)
    at fulfilled (/data/node_modules/@my-company/mysql-connection-pool/dist/ConnectionPool.js:5:58)
```

Now this error is wrapped by my own logic, but the cause I think it quite clear: `Incorrect arguments to mysqld_stmt_execute`. The query being passed to the `execute` method on the `mysql2/promise` library is `SELECT * FROM message WHERE chat_id = ? ORDER BY sent DESC LIMIT ? OFFSET ?` and the values are `["01ensmg6m4dea0gh7gsjgaa5gb", 50, 0]`. For some reason this only fails on the CI, and not locally on any of my (or my colleagues') machines.

If I change this to `query` instead of `execute`, it works on the CI. If I form this query manually using the values that were passed-in, and run it, it also works. 

Any idea what's happening here? Is there some failure in how the parameters are being transferred to the MySQL service before being inserted?

perry-mitchell:
Not sure it helps, but this is the usage:

```typescript
export async function getChatMessages(
    chatID: string,
    limit: number,
    offset: number = 0,
    connection = getConnection()
): Promise<Array<MessageSlow>> {
    const [
[...]

PR #1407, opened 2021-10-11
Change the way types for prepared statement parameters are calculated

sidorares:
Currently types are inferred based on what is in the input (Date -> mysql date, js number -> mysql double, everything else - string ). This used to work ok until mysql server version 8.0.22 

Calling `execute('SELECT * from foo limit ?' , [1])` results in `Incorrect arguments to mysqld_stmt_execute` error. The type of the parameter server expects ( LONGINT ) is incompatible to what is actually sent ( DOUBLE ).

The fix is to use parameters column definitions returned from the previous `prepare()` call and use types from there instead of inferring type from the provided input

### DIFF
--- a/lib/packets/execute.js
+++ b/lib/packets/execute.js
@@ -43,12 +43,16 @@ function toParameter(value, encoding, timezone, parameterHints) {
         return Packet.prototype.writeDate.call(this, value, timezone);
       };
       return { value, type, length, writer };
-    } else if (isJSON(value)) {
+    }
+     
+    if (isJSON(value)) {
       value = JSON.stringify(value);
       type = Types.JSON;
       length = Packet.lengthCodedStringLength(value, encoding);
       return { value, type, length, writer };
-    } else if (Buffer.isBuffer(value)) {
+    } 
+
+    if (Buffer.isBuffer(value)) {
       length = Packet.lengthCodedNumberLength(value.length) + value.length;
       writer = Packet.prototype.writeLengthCodedBuffer;
       type = Types.VAR_STRING;


sidorares:
ref #1239

kimf:
Hello @sidorares 
For what it's worth, I had the exact same issue as mentioned in #1239 and tried this branch with great success.

mlucic:
Any updates on this?

sidorares:
@mlucic unfortunately no progress here (but its quite high in my priority)

The way I want to implement this currently is different from this PR though.

The plan is:

1) Introduce "a typed parameter" helpers which would hold the value you pass together with type information. Not sure about the exact api, but it might look like this:

```js
const mediumintparam = mysql.TypedParameter.MEDIUMINT(123);
const floatparam =  mysql.TypedParameter.FLOAT(123);

conn.execute('select ?+? as sum', [mediumintparam, floatparam]);
```
when used via type helper the type sent with COM_EXECUTE is always what you specify

2) when untyped js value is passed we have 2 options: infer from current value ( string / date / number / null / non-null object ) or use response from COM_PREPARE. The problem "type hint from the server" is that its not always accurate ( for example, there are scenarios when the server has no way of telling parameter type - `select ?+? as sum` ). I'll need to think when this hint might be useful and when not ( any feedback is appreciated! ), ideally we need a matrix where there is js type on one axis, COM_EXECUTE type hint on another axis mapped to what the driver will actually use for type

step 1) can be done now and would help solving current edge cases with a bit of manually added typings ( as in mysql types metadata in users code, not typescript )

after that we can add step 2 to make default map work with no surprises most of the time. Mysql is quite capable of casting back to a correct value on a server side, in earlier versions of this driver all `.execute()` parameters were converted to a string ( sometimes its even more efficient, "1" takes 2 bytes over the wire where double precision number is 8 )

PR #1410, opened 2021-10-12
Update collation list up to MySQL 8.0.26

testn:
I noticed that some collations are missing so I pulled up the latest collation list from MySQL by 

```
mysql> select concat('exports.',UPPER(COLLATION_NAME),' = ',id,';')
    ->        FROM INFORMATION_SCHEMA.COLLATIONS
    ->        ORDER BY ID;
```


testn:
utf8_general50_ci  and ucs2_general50_ci seem to be removed.

https://bugs.launchpad.net/percona-server/+bug/1163324

testn:
I just tried to store 😂 into utf8_unicode_ci and it said
```
    code: 'ER_TRUNCATED_WRONG_VALUE_FOR_FIELD',
    errno: 1366,
    sqlState: 'HY000',
    sqlMessage: "Incorrect string value: '\\xF0\\x9F\\x98\\x82' for column 'field' at row 1",
    sql: "INSERT INTO `test-charset-encoding` (field) values('😂')"
```

sidorares:
I believe I was able to insert non-BMP 💩 using utf8 mysql encoding ( see https://github.com/sidorares/node-mysql2/pull/374#issuecomment-241748134 ), so not sure if that confirms "mysql utf8 = cesu-8", utf8mb3 = same as cecu-8 but only BMP is supported, utf8mb4 = "normal modern utf8"

sidorares:
unit test: https://github.com/sidorares/node-mysql2/blob/694e1003732f048c91a2ab99f7a7667d88a02549/test/integration/connection/encoding/test-non-bmp-chars.js#L9-L15

sidorares:
can column name use other encodings ( win1251, koi8 for Cyrillic as an example )? If yes not sure if worth adding exception / custom logic for perf benefits

dogancan-yigit:
Hello, would this be merged? I'm having problems with MySQL 8.x

testn:
Let me check.

ahmedbodi:
any update on this?

sidorares:
@ahmedbodi I think the discussion got sidetracked a bit into more complex issue, updated collation list LGTM to me. Merging now

Issue #1418, opened 2021-10-15
Sequelize broken with v2.3.1


ddolcimascolo:
Hello,

Long-time user of sequelize + mysql2 here.
Our build starts to fail when upgrading to v2.3.1 of your library. I didn't investigate deeply yet but the error we're getting is

```
SequelizeDatabaseError: Cannot read property 'parse' of undefined
  at Query.formatError (node_modules/sequelize/lib/dialects/mysql/query.js:265:16)
  at Query.run (node_modules/sequelize/lib/dialects/mysql/query.js:77:18)
```

I can also open an issue there if you feel the issue is on their side, but you released a _minor_ version wich isn't supposed to contain breaking changes.

I can provide logs, reproduction repo, anything you'd need to debug.

Cheers,
David

rnkhouse:
It's breaking with the 2.3.1 version. The error we get is: `uncaughtException: The value of "offset" is out of range. It must be >= 0 and <= 1248. Received 1252`

edgedemon:
sequelize is broken with version 2.3.1 not with 2.3.0.

the problem comes from this file:
https://github.com/sidorares/node-mysql2/blob/master/lib/parsers/text_parser.js
you are using readCodeFor which applies JSON.parse 
![image](https://user-images.githubusercontent.com/5783253/137540036-5f048c82-cead-4600-998f-88c67c40a055.png)
and then assigning that to _this which ends with 
![image](https://user-images.githubusercontent.com/5783253/137539897-127f052d-2160-4259-b557-afb62a73d2b0.png)
_this.JSON.parse(packet.readLengthCodedString("utf8"))
instead of 
JSON.parse(_this.packet.readLengthCodedString("utf8"))
![image](https://user-images.githubusercontent.com/5783253/137540115-f908dc89-7322-42e0-82c2-bd891566b8ce.png)

sidorares:
@testn looks like #1402 is causing this regression ( typeCast + JSON field )

sidorares:
my attempt to fix: #1420 
cc @testn @edgedemon @ddolcimascolo

testn:
@sidorares sorry I must have missed a case with JSON.

sidorares:
@testn I'm a bit worried about `uncaughtException: The value of "offset" is out of range. It must be >= 0 and <= 1248. Received 1252` - could it be we are reading data from wrong packet reference?

testn:
Let me take a look.

testn:
[...]

PR #1419, opened 2021-10-15
fix(json parsing): moved json parse to fix issue 1418

edgedemon:
fixes the issue with JSON data types that was introduced on version v2.3.1 as reported on issue https://github.com/sidorares/node-mysql2/issues/1418, notice the issue was first reported with sequelize but JSON data type will likely fail with all enviroments.


sidorares:
@edgedemon closing as already fixed by #1420

Issue #1413, opened 2021-10-13
Bad logic to decode the column name


testn:
1. Create a table with the following definition

```CREATE TABLE `test-encoding2` (
  `平仮名` varchar(100) CHARACTER SET tis620 NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=tis620```

2. Insert data

```INSERT INTO `test-encoding2`
VALUES ('กขค');```

2. Set the result charset to null
```SET chracter_set_results=NULL;```

3. select data

```select * from `test-encoding2`;```

4. it will decode the column name as `'ๅนณไปฎๅ��'` instead of `平仮名` as expected

testn:
you have to set 'SET chracter_set_results=NULL;'

sidorares:
doh. yep, same results as you posted now

testn:
That's right.

testn:
BTW, I have put a thought around utf8mb3/utf8mb4 a little. I have this thought....

1. anything encoded with utfmb3 can be "decoded" by either cesu8, utf8 without any problem as it has no non-BMP characters
2. for encoding utf8mb3 character, mysql2 should either (choice to implement)
- if the connection charset is utf8mb4, there is no problem. Any non-BMP characters will fail to be encoded in utf8mb3 on the server side
- if the connection charset is utf8mb3 or something more limiting, we have a couple choices
     - replace the invalid character with ?
     - encode a non-bmp character as 6-byte. This will enable you to jam in non-BMP characters where you are not supposed to. But it will likely create a problem later.

testn:
Oh really? So we just need to add `character_set_results`?

testn:
Let me test this out!

testn:
@sidorares  although the fix is not ready yet as I'm adding more tests, feel free to take a look at the diff.

test-track-encoding test had a couple problems
1. it relied on connection.option.charSet rather than tracking the session change from resultset_header.js
2. it used utf8 to encode the query instead of koi8r as the code expected

PR #1424, opened 2021-10-16
Fix Bad logic to decode the column name #1413

testn:
Introduce character_set_results tracking to ensure that we can
read the column name properly

###### still in the draft mode ########

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -55,7 +55,7 @@ class Query extends Command {
 
     const cmdPacket = new Packets.Query(
       this.sql,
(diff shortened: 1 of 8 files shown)


testn:
@sidorares  this is the first attempt trying to fix #1413. It is still rough but it should demonstrate what we can address this problem.

testn:
@sidorares  that exposes the problem that I'm working on a fix. Basically, when you specify character_set_client, you have to encode your SQL in that encoding as well.

Nikhilkk1994:
What the need for this ?

PR #1428, opened 2021-10-17
Add geometry typecast test

testn:
Add addition typecast tests for geometry

### DIFF
--- a/test/integration/connection/test-typecast.js
+++ b/test/integration/connection/test-typecast.js
@@ -5,19 +5,29 @@ const connection = common.createConnection();
 const assert = require('assert');
 
 connection.query('CREATE TEMPORARY TABLE json_test (json_test JSON)');
-connection.query('INSERT INTO json_test VALUES (?)', JSON.stringify({ test: 42 }));
+connection.query(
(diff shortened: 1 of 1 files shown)


testn:
@sidorares any idea why geo stuff is broken in 5.7? I couldn't find any related release notes.

sidorares:
@testn it fails at `CREATE TEMPORARY TABLE geom_test (p POINT SRID 0, g GEOMETRY NOT NULL SRID 0)` query, I assume this is newer syntax and 5.7 server does not understand it

I can't see geometry types in https://dev.mysql.com/doc/refman/5.7/en/data-types.html
( there are geometry functions though - https://dev.mysql.com/doc/refman/5.7/en/functions.html )

We could check server version and skip the test if 5.7 or earlier, or somehow feature detect via sql

sidorares:
CI logs from my copy of your commit - https://github.com/sidorares/node-mysql2/actions/runs/3906857317/jobs/6675468924

testn:
oh sure... let me do that

wellwelwel:
Closing due to #2395.

Thanks, @testn 🙋🏻‍♂️

PR #1434, opened 2021-10-25
Hotfix: incorrect argument type in PoolCluster add method

claudio-at-float:
`createPoolCluster` method takes `BasePoolCluster.PoolClusterOptions` [type](https://github.com/sidorares/node-mysql2/blob/07a429d9765dcbb24af4264654e973847236e0de/typings/mysql/lib/PoolCluster.d.ts#L8) as argument to create a [pool cluster](https://github.com/mysqljs/mysql#poolcluster-options).

When adding nodes to the PoolCluster using the `add` method this parameter type is not correct since we need to pass options to define where nodes are and how to authenticate against them (`host`, `port`, etc.). 
These options are defined in the `PoolOptions` [type](https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/Pool.d.ts#L10) which extends the `Connection.ConnectionOptions` class.

### DIFF
--- a/typings/mysql/lib/PoolCluster.d.ts
+++ b/typings/mysql/lib/PoolCluster.d.ts
@@ -1,7 +1,7 @@
 
-import Connection = require('./Connection');
 import PoolConnection = require('./PoolConnection');
 import {EventEmitter} from 'events';
+import { PoolOptions } from './Pool';
 
 declare namespace PoolCluster {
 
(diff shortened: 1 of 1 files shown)


santosh1997:
Hi @sidorares,
When will this be released? Any specific reason not releasing it till now?
I would be happy to help on it and it released

sidorares:
@santosh1997 no particular reason, just no time to prepare the release ( somewhat manual process, partially because I want a bit of manual check partially because it's not automated yet )

you can help with automating - create a PR identical to https://github.com/sidorares/json-bigint/pull/77

currently another manual step in addition to publishing is to check all merged PRs and add summary to readme. Would be cool to have a gh-actions workflow that acts as a chat bot and 1) comments with links to PRs merged since last release 2) allows ( potentially by someone in the pre-approved list ) respond with edited list and merge that into changelog and publish a release ( tag, npm push and github release with description, same as what was added to changelog )

sidorares:
@santosh1997 I'll try to make a release tonight or at most this weekend. You can help by doing what imaginary automation workflow would do - check last release tag, verify what's latest on npm corresponds to it, prepare list of changes and comment here

PR #1436, opened 2021-10-26
Add support for multi-factor authentication

ruiquelhas:
MySQL 8.0.27 introduced support for multi-factor authentication, as described in the following link:

https://dev.mysql.com/doc/refman/8.0/en/multifactor-authentication.html

User accounts can now be configured with up to three authentication plugins and corresponding passwords or additional options. Although not all authentication plugins can be used for the 2nd and 3rd factor, any MySQL client should be flexible enough to accommodate those that can.

This proposal follows the same convention used by the official MySQL CLI and introduces three new optional connection options for users to provide passwords for the corresponding _nth_ authentication factor, namely:

- `password1` (symbolic replacement for `password`)
- `password2`
- `password3`

The client handshake workflow was extended to account for an additional `AuthNextFactor` packet that is sent by the server (when the `MULTI_FACTOR_AUTHENTICATION` capability is enabled) after a successful 1st and/or 2nd factor authentication steps.

The packet exchange for each factor remains unchanged, and after the last authentication factor, the server either replies with an [`OK_Packet`](https://dev.mysql.com/doc/internals/en/packet-OK_Packet.html) or an [`ERR_Packet`](https://dev.mysql.com/doc/internals/en/packet-ERR_Packet.html) depending on whether or not the access is granted, similar to what already happens in a single-factor authentication setup.

### DIFF
--- a/documentation/Authentication-Switch.md
+++ b/documentation/Authentication-Switch.md
@@ -1,15 +1,45 @@
 # Authentication switch request
 
-During connection phase the server may ask client to switch to a different auth method.
-If `authSwitchHandler` connection config option is set it must be a function that receive
-switch request data and respond via callback. Note that if `mysql_native_password` method is
-requested it will be handled internally according to [Authentication::Native41]( https://dev.mysql.com/doc/internals/en/secure-password-authentication.html#packet-Authentication::Native41) and
-`authSwitchHandler` won't be invoked. `authSwitchHandler` MAY be called multiple times if
-plugin algorithm requires multiple roundtrips of data exchange between client and server.
-First invocation always has `({pluginName, pluginData})` signature, following calls - `({pluginData})`.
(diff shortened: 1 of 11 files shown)


ruiquelhas:
@sidorares any thoughts about my last comment?

sidorares:
sorry @ruiquelhas this PR disappeared from my active memory

I think I'm ok with password1/2/3, the more aligned with other tools the better. we can document it as experimental / not stable yet and see how it goes

Re typescript - all manual for now. Maybe at least add password1/2/3 as possible entries in config so that compiler won't throw type error

ruiquelhas:
@sidorares I've introduced the TypeScript definitions for the new connection properties. I'm not sure why the checks are failing though. The ones I'm able to easily run locally (like `lint`) seem fine.

ruiquelhas:
@sidorares did you manage to give this a look? Anything I can help with?

Issue #1328, opened 2021-05-28
Feature request: mysql_reset_connection


asmgit:
Feature request: mysql_reset_connection
https://dev.mysql.com/doc/c-api/8.0/en/mysql-reset-connection.html

PR #1437, opened 2021-10-28
Implement CMD_RESET_CONNECTION

testn:
Address #1328

still in the draft... 

To add integration tests


sidorares:
hey @testn I might have some time this weekend to clear backlog of PRs - do you need any help with integration tests for this?

testn:
> hey @testn I might have some time this weekend to clear backlog of PRs - do you need any help with integration tests for this?

Awesome. I still cannot figure out why it failed in some CI setting.

Also, connection.end() is not "async". However, to end() the connection, I would like to make a reset() call which is async network call. What should we do with it?

danielgindi:
I just stumbled upon a bug caused by this. I was wondering "do they reset the connection?", and the answer is apparently  "no".
Now discussing a "no surprises" methodology should always result to a "default: true". As no surprises mean: I expect a clean connection from the pool. How the heck should I know what is currently pending on a *random* connection I just got from the pool?
Anyone expecting the opposite, is working in some extreme case where there is zero session and wants to gain 1-2% more in perf.

sidorares:
@danielgindi 100% agree ( though would be great to have some tests to support 1-2% perf hit assumption ). Any preference on the name for this flag? `resetOnAcquire` ( default true ) ?

danielgindi:
> @danielgindi 100% agree ( though would be great to have some tests to support 1-2% perf hit assumption ). Any preference on the name for this flag? `resetOnAcquire` ( default true ) ?

As for the naming - this should just be "connectionReset". Other connectors do this too: https://mysqlconnector.net/connection-options/.
I'd like to have a `connectionResetOnRelease` option, to automatically avoid deadlocks (if using GET_LOCK etc.), but this could introduce an additional overhead for connections that will not be used again, and should be opt-in.

As for the perf hit - I tested about a year ago with the .net connector, with and without the reset. I believe it's safe to assume it would be the same here. Anyway it's a non-issue, as you never over-optimize at safety's expense.

danielgindi:
@sidorares Note that in some cases, the perf hit may be larger. i.e:
1. When the roundtrip (`ping mysql_server`) is too big. Because in this case the code has to wait for an additional roundtrip until it can start querying.
2. When there's a pending transaction that will be rolled back by the `reset connection` command, and the rollback is a large set of rows and takes some time.

In these cases, it may even be better to **always** reset the connection when *returning* to the pool.
And then we can just mark it as "not ready yet for acquiring" until the reset has completed.

You can see some discussion here: https://github.com/mysql-net/MySqlConnector/issues/178

sidorares:
> I'd like to have a connectionResetOnRelease option, to automatically avoid deadlocks (if using GET_LOCK etc.), but this could introduce an additional overhead for connections that will not be used again, and should be opt-in.

hm, doing reset on connection release might actually be better for performance ( or rather consumer latency ) as in this case there is a chance that reset is happening while no one is waiting

lib/connection.js
```diff
@@ -707,6 +707,10 @@ class Connection extends EventEmitter {
     return this.addCommand(new Commands.Ping(cb));
   }
 
+  reset(cb) {
```
sidorares, line 710: I believe this need to perform some additional client cleanup, for example prepared statements need to be discarded.

testn: ok let me take a look

testn: Preparedstatement is discarded automatically
https://mariadb.com/kb/en/com_reset_connection/

danielgindi:
@testn I believe that `releaseConnection` should look like this:

```js
releaseConnection(connection) {
    let cb;
    if (!connection._pool) {
      // The connection has been removed from the pool and is no longer good.
      if (this._connectionQueue.length) {
        cb = this._connectionQueue.shift();
        process.nextTick(this.getConnection.bind(this, cb));
      }
    } else if (this._connectionQueue.length) {
      cb = this._connectionQueue.shift();
      process.nextTick(cb.bind(null, null, connection));
    } else {
      connection.reset(() => {
        this._freeConnections.push(connection);
        this.emit('release', connection);
      });
    }
  }
```

lib/commands/reset_connection.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  start(packet, connection) {
+    const req = new Packets.ResetConnection();
+    connection.writePacket(req.toPacket());
+    return ResetConnection.prototype.resetConnectionResponse;
+  }
+
+  resetConnectionResponse() {
+    if (this.onResult) {
```
andrey-sidorov-iag, line 19: re statement cache:
1) add private connection helper method connection._initialiseStatementCache({ disableDispose: boolean })

something along the lines:

```js
_initialiseStatementCache({ disableDispose }) {
  if (this._statements) {
    if (disableDispose) {
       this._statements.dispose = null;
    }
    this._statements.clear();
  }
  this._statements = new LRU({
    max: this.config.maxPreparedStatements,
    dispose: function(key, statement) {
      statement.close();
    }
  })
}
```

2. replace line [66 in connection.js](https://github.com/sidorares/node-mysql2/blob/299ed3d95576da72ce71933ea26cb76f8ab21f79/lib/connection.js#L66) with `this._initialiseStatementCache({ disableDispose: false });`

3) in `ResetConnectionCommand.resetConnectionResponse(packet, connection)` add call `connection._initialiseStatementCache({ disableDispose: true })` at the beginning before calling `onResult()`

lib/connection.js
sidorares: yes, server side, but we need to delete statements cache otherwise it might be used incorrectly with `execute()` after reset

I think this is what needs to be done

1) add private connection helper method `connection._initialiseStatementCache({ disableDispose: boolean })`

something along the lines:

```js
_initialiseStatementCache({ disableDispose }) {
  if (this._statements) {
    if (disableDispose) {
       this._statements.dispose = null;
    }
    this._statements.clear();
  }
  this._statements = new LRU({
    max: this.config.maxPreparedStatements,
    dispose: function(key, statement) {
      statement.close();
    }
  })
}
```

2. replace line [66 in connection.js](https://github.com/sidorares/node-mysql2/blob/299ed3d95576da72ce71933ea26cb76f8ab21f79/lib/connection.js#L66) with `this._initialiseStatementCache({ disableDispose: false });`

3) in `ResetConnectionCommand.resetConnectionResponse(packet, connection)` add call `connection._initialiseStatementCache({ disableDispose: true })` [at the beginning](https://github.com/sidorares/node-mysql2/blob/299ed3d95576da72ce71933ea26cb76f8ab21f79/lib/commands/reset_connection.js#L19) before calling `onResult()`

testn: Ah... got it

testn: I thought statement can be reused even though the connection is reset

sidorares: I need to double check but I'm pretty sure statement id is sequential 1, 2, 3, ... for each `prepare()` command ( can't remember if 0 or 1 based ) and after reset it goes again as 1, 2, 3, ...

sidorares: small correction: set dispose to noop instead of null, e.i `this._statements.dispose = () => {};`

danielgindi:
We've tried this in a production environment, after modifying the `releaseConnection` function -> it looks like problem solved!

lib/connection.js
testn: gotcha

danielgindi:
> @testn I believe that `releaseConnection` should look like this:
> 
> 
> 
> ```js
> 
>   releaseConnection(connection, reset = true) {
> 
>     let cb;
> 
>     if (!connection._pool) {
> 
>       // The connection has been removed from the pool and is no longer good.
> 
>       if (this._connectionQueue.length) {
> 
>         cb = this._connectionQueue.shift();
> 
>         process.nextTick(this.getConnection.bind(this, cb));
> 
>       }
> 
>     } else {
> 
>       if (reset) {
> 
>         connection.reset(() => {
> 
>           this.releaseConnection(connection, false);
> 
>         });
> 
>       } else {
> 
>         if (this._connectionQueue.length) {
> 
>           cb = this._connectionQueue.shift();
> 
>           process.nextTick(cb.bind(null, null, connection));
> 
>         } else {
> 
>           this._freeConnections.push(connection);
> 
>           this.emit('release', connection);
> 
>         }
> 
>       }
> 
>     }
> 
>   }
> 
> ```

@testn  have you taken a look at this? This is to actually trigger the reset.

testn:
> > @testn I believe that `releaseConnection` should look like this:
> > ```js
> >   releaseConnection(connection, reset = true) {
> > 
> >     let cb;
> > 
> >     if (!connection._pool) {
> > 
> >       // The connection has been removed from the pool and is no longer good.
> > 
> >       if (this._connectionQueue.length) {
> > 
> >         cb = this._connectionQueue.shift();
> > 
> >         process.nextTick(this.getConnection.bind(this, cb));
> > 
> >       }
> > 
> >     } else {
> > 
> >       if (reset) {
> > 
> >         connection.reset(() => {
> > 
> >           this.releaseConnection(connection, false);
> > 
> >         });
> > 
> >       } else {
> > 
> >         if (this._connectionQueue.length) {
> > 
> >           cb = this._connectionQueue.shift();
> > 
> >           process.nextTick(cb.bind(null, null, connection));
> > 
> >         } else {
> > 
> >           this._freeConnections.push(connection);
> > 
> >           this.emit('release', connection);
> > 
> >         }
> > 
> >       }
> > 
> >     }
> > 
> >   }
> > ```
> 
> @testn have you taken a look at this? This is to actually trigger the reset.

Can you explain a little bit?

danielgindi:
@testn we want the connector to reset a connection in the background before making it available again. 
So what I did there is added a flag, that makes it reset the connection, and then call the logic again to return to the pool.
It could of course be split into two functions instead, like releaseConnection and _releaseConnectionImpl.

Now if there was an error, even in the `reset` command, then it will have its `_pool` prop removed, which will make the function release the connection without returning to the pool.

We can make this flag based, but this should be the default behavior anyway.

mscdex:
What's left to get this over the finish line?

sidorares:
continuing in https://github.com/sidorares/node-mysql2/pull/4148

PR #1445, opened 2021-11-05
Fix 100% CPU problem

testn:
Addressing #1432  where it was a regression bug from #1402


testn:
@sidorares Once I manage to re-create a problem, I will submit another PR for the test

Issue #805, opened 2018-07-14
MySQL operation failed: Unknown prepared statement handler (${#}) given to mysqld_stmt_execute


hellopat:
I have a service that has been running in production for a month with no issues.  Last evening, the service failed after all queries using `.execute` were throwing this error.  Restarting the service resolved the issue.

Let me know if I can provide more information.

sidorares:
interesting, never seen anything like this.

"prepared statement handler" is just integer, it's local to connection and usually starts  with `1`. When you do `connection.execute('select 1+?', [1])` for the first time this is what happens:

```
client: prepare "select 1 + ?"
server: statement id = 1, expect one parameter etc etc
client: execute id=1, param = 1
server: ok, 1 column, 1 row = "2" 
```

When later you do `connection.execute('select 1+?', [123])` same happens but instead of prepare id is looked in prepared statement cache

```
client, after realising there is already ps id for 'select 1+?': execute id=1, param = 1 
server: ok, 1 column, 1 row = "124" 
```

if you think you can possibly reproduce this, can you add `console.log(this.statement.id)` somewhere here https://github.com/sidorares/node-mysql2/blob/224c74de8184f78c540f343c432c0f37986c8aea/lib/commands/execute.js#L42 ?

This might indicate there is a bug in a way we cache statement id, but could be something else

hellopat:
Thanks for the info @sidorares.

I found it strange myself that this occurred, especially after the service was running fine at load for a month.

I'll see if I can get something into place to debug if the issue occurs again.

dekyfin:
I have been getting the same error on my app. The error occurs when script is called multiple times in a short period of time. I think module is using a cached statement even though the statement was close. I would appreciate it if I could get a work-around

This is a simplified version of the code
``` javascript
// A list of rows to be inserted into table
var data = [
    {col1: "val1", col2: "val2"},
    {col1: "val3", col2: "val4"}
];

var stmnt, conn; // Statement and connection variables to be used in this function

// Get a connection from Pool
Pool.getConnection( function( errCon, connection ){

    // Interupt the script if error occurs
    if( errCon ){
        finalize( errCon );
        return;
    }
    conn = connection; 

    // Prepare statement
    connection.prepare(
        "INSERT into table( col1, col2 ) VALUES ( ?, ? ) ",
        function( err, statement ){

            // Interupt the script if error occurs
            if( err ){
                finalize( err );
                return;
            }

            stmnt = statement;

[...]

PR #1493, opened 2022-01-14
fix(typing & documentation): prepared statements usage

xWTF:
## Issue
1. The prepared statement typings are missing
2. The documentation of prepare() usage are kinda wrong

Let's say that we have following code (assume we've fixed the typings):
```typescript
async function updateStuffs(updates: [string, string][]) {
    const statement = await conn.prepare('UPDATE Users SET A = ? WHERE B = ?');
    for (const update of updates) {
        await statement.execute(update);
    }
    statement.close(); // <-- Notice this line
}

await updateStuffs(...);
await updateStuffs(...); // Call multiple times
```

The first call to `updateStuffs` works fine, but later calls (assume the connection didn't get closed) produces the following error:
```
Error: Unknown prepared statement handler (#XXX) given to mysql_stmt_precheck
```

I've searched a bit and found this issue: https://github.com/sidorares/node-mysql2/issues/805, where @rallfo [found the exact cause of this](https://github.com/sidorares/node-mysql2/issues/805#issuecomment-447871245). And the fix is super simple:
```typescript
async function updateStuffs(updates: [string, string][]) {
    const statement = await conn.prepare('UPDATE Users SET A = ? WHERE B = ?');
    for (const update of updates) {
        await statement.execute(update);
    }
    conn.unprepare('UPDATE Users SET A = ? WHERE B = ?'); // <-- Fixed!
}

await updateStuffs(...);
await updateStuffs(...); // Works now
```

According to [current docs](https://github.com/sidorares/node-mysql2#using-prepared-statements), I believe there's a documentation error.

## Fix
1. Added typings of `Connection.prepare`, `Connection.unprepare` and `PreparedStatementInfo` to promised version, with a warning in `PreparedStatementInfo.close` jsdoc
2. Updated the documentation to use `unprepare` instead of `close`

### DIFF
--- a/documentation/Prepared-Statements.md
+++ b/documentation/Prepared-Statements.md
@@ -13,8 +13,11 @@ connection.execute('select 1 + ? + ? as result', [5, 6], (err, rows) => {
 // close cached statement for 'select 1 + ? + ? as result'. noop if not in cache
 connection.unprepare('select 1 + ? + ? as result');
 ```
+Note that `connection.execute()` will cache the prepared statement for better performance, remove the cache with `connection.unprepare()` when you're done.
 
 ## Manual prepare / execute
+
+Manually prepared statements doesn't comes with LRU cache and SHOULD be closed using `statement.close()` instead of `connection.unprepare()`.
 
 ```js
 connection.prepare('select ? + ? as tests', (err, statement) => {
@@ -27,9 +30,9 @@ connection.prepare('select ? + ? as tests', (err, statement) => {
     // -> [ { tests: 3 } ]
   });
 
-  // do not use statement.close(), use connection.unprepare() to remove lru cache properly.
+  // don't use connection.unprepare(), it won't work!
   // note that there is no callback here. There is no statement close ack at protocol level.
-  connection.unprepare('select ? + ? as tests');
+  statement.close();
 });
 ```
 Note that you should not use statement after connection reset (`changeUser()` or disconnect). Statement scope is connection, you need to prepare statement for each new connection in order to use it.
--- a/documentation/Prepared-Statements.md
+++ b/documentation/Prepared-Statements.md
@@ -27,8 +27,9 @@ connection.prepare('select ? + ? as tests', (err, statement) => {
     // -> [ { tests: 3 } ]
   });
 
+  // do not use statement.close(), use connection.unprepare() to remove lru cache properly.
   // note that there is no callback here. There is no statement close ack at protocol level.
-  statement.close();
+  connection.unprepare('select ? + ? as tests');
 });
 ```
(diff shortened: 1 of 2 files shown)


xWTF:
IDK why the checks are failing -- This PR only touches documentation and typing definitions, and neither are associated with runtime stuffs.
Seems like windows CI tests failed on date type cast asset, and coverage test fails with `github.GithubException.GithubException: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue-comment"}`, no idea since I didn't touch anything related to those

sidorares:
@xWTF I'll try to rerun. Coverage job is known to fail, I need to fix or disable it

sidorares:
I've been thinking about "manual prepare caching the result and .close() is not cleaning the cache" and leaning towards "son't cache for manual prepare" solution vs advising to use `connection.unprepare()`. While potentially backwards incompatible seems like it's a more expected behaviour. Let me fix that first and we'll review your documentation update

xWTF:
I think it's a good idea for `.close()` to remove cache, not removing cache should be considered as a bug based on my review on the impl of `.prepare()`, `.unprepare()`, and `.execute()`.
However, since the cache is associated with connection object, I'd suggest you to only cache stuffs for `execute` instead of `prepare`. For example make `conn.addCommand` accept another parameter indicating it not to cache the cmd. In that case, `stmt.close()` can be safely used without having to go back to connection object and search for cache.

sidorares:
> make Prepare.prepareDone not to cache when using manual prepare.

Yes, my intention exactly. The reasoning here is manual `.prepare()` is typically when user wants their own caching implementation instead of built in and ps lifecycle in the case of `connection.execute()` if focused on default convenience + performance. The fact that `connection.prepare()` saves result in the cache currently was a mistake on my side

xWTF:
Fine, I've updated the doc update according to this new behaviour

jmordica:
At first I thought this was a duplicate of #1565 but it seems to fix documentation issues as well related to prepared statements. Can this be merged?

sidorares:
thanks for the ping @jmordica , I think this change should be merged

TomMiller-mas:
I am having the same problem running a stored procedure every 1m 30s.  But I am not preparing it manually.  

     mdbConn = await mdbPool.getConnection();
     // Use a stored procedure to select and set a job number to prevent duplicate sending of emails.
      const resultSpPrepare = await mdbConn.execute("CALL EM_Email_Prepare_To_Send_U_sp(?);", [
        data.RequestUserId,
      ]);
     const jobNumber = resultSpPrepare[0][0].JobNumber;
     dlog("processEmails-mdb Job Number", jobNumber); //Console.log

PR #1497, opened 2022-01-29
Expose authPlugins to enable example from docs

MasterOdin:
The example of using `authPlugins` given in https://github.com/sidorares/node-mysql2/blob/master/lib/auth_plugins/caching_sha2_password.md is not possible, as this module does not expose a top level `authPlugin` object. This PR adds that top-level object, as well as fully typing for the existing plugins.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -164,13 +164,6 @@ export interface Pool extends mysql.Connection {
   config: mysql.PoolOptions;
 }
 
-type authPlugins = (pluginMetadata: {
-  connection: Connection;
-  command: string;
-}) => (
-  pluginData: Buffer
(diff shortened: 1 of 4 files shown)


MasterOdin:
Not sure why the Windows tests are failing on a date test. That should be unrelated to this PR?

sidorares:
@MasterOdin definitely unrelated, I saw that started to happening in other PRs. Need to investigate that

sidorares:
@MasterOdin could you rebase please? I have accidentally merged code in the master that fails linter but that should be fixed now

MasterOdin:
@sidorares I've rebased the PR onto the latest master.

sidorares:
I don't believe perf issue is real, nothing in the PR can contribute to it. Likely a slower CI runner

PR #1506, opened 2022-02-09
Allow connectAttributes to be set through URI config

mamof:


sidorares:
hey @mamof , not sure about kay:value pair encoding, maybe we should be consistent with how for example `ssl` option is passed to connection string:

`mysql://username:password@us-east.connect.psdb.cloud/dbname?ssl={"rejectUnauthorized":true}`

Maybe just `JSON.stringify()`? `mysql://username:password@us-east.connect.psdb.cloud/dbname?ssl={"rejectUnauthorized":true}&connectAttributes={"program_name":"testProg","second_attr":"example"}`

PR #1517, opened 2022-02-16
Add minVersion for ssl option.

cweijan:
I need to configure minVersion option, there is currently no option for this, I can only configure it via `as SslOptions as any` in  typescript, it's bad. so I created it.


cweijan:
I think a better way is by referencing `SecureContextOptions` in the type definition file, because ca also supports buffer, but currently only supports string.

sidorares:
Thanks @cweijan !

PR #1518, opened 2022-02-16
Reference SecureContextOptions.

cweijan:
Hi, this module defines the SSL Option, but I think a better way is to directly refer to the SecureContextOptions of the node tls module, because now this option is incomplete, for example, it does not support ca files in the form of buffers.
image
image


cweijan:
This means that the user has to use @types/node, I'm not sure if it will work correctly.

PR #1527, opened 2022-03-01
Fix precision bug in parseFloat affecting select(DOUBLE) in prepared statements

wesgarland:
This corrects Issue 1525 by falling back to the JS engine's own implementation of parseFloat.

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -645,44 +645,17 @@ class Packet {
     return new Date(`${str}${timezone}`);
   }
 
-  parseFloat(len) {
-    if (len === null) {
-      return null;
-    }
-    let result = 0;
+  parseFloat(len)
+  {
+    const start = this.offset;
     const end = this.offset + len;
-    let factor = 1;
-    let pastDot = false;
-    let charCode = 0;
+    this.offset += len;
+
     if (len === 0) {
       return 0; // TODO: assert? exception?
     }
-    if (this.buffer[this.offset] === minus) {
-      this.offset++;
-      factor = -1;
-    }
-    if (this.buffer[this.offset] === plus) {
-      this.offset++; // just ignore
-    }
-    while (this.offset < end) {
-      charCode = this.buffer[this.offset];
-      if (charCode === dot) {
-        pastDot = true;
-        this.offset++;
-      } else if (charCode === exponent || charCode === exponentCapital) {
-        this.offset++;
-        const exponentValue = this.parseInt(end - this.offset);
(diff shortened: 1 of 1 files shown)


wesgarland:
@wellwelwel Exactly what kind of tests are you looking for, here?

The current code is demonstrably wrong.  Testing the implementation of parseFloat() is covered by the V8 team.

wellwelwel:
> @wellwelwel Exactly what kind of tests are you looking for, here?
> 
> The current code is demonstrably wrong. Testing the implementation of parseFloat() is covered by the V8 team.

Hi, @wesgarland. As a fix, to effectively check and ensure that the bug does not recur, it's a useful practice, when possible, ensuring automated tests that reproduce the failure and wait for the failure if the proposed correction is not applied. This also helps to ensure that the correction actually fixes the proposed error.

wellwelwel:
I _forced_ the current tests to run through this _PR_. For me, just removing the variables that are no longer used after the fix is fine. I'm going to take a break to carefully read the discussion that led to this PR in #1525.

wesgarland:
@wellwelwel Thanks for picking this up and running with it. I don't think there was a CI pipeline when I patched this ~ 4 years ago or I would have picked up on the lint errors. Or maybe I just missed it... anyhow.

To summarize "how we got here", I discovered that selecting float64 values from mysql would give different values in node in some circumstances. The reason is that in those circumstances, they are transmitted over the network as ASCII strings, and sidorares wrote his own implementation of parseFloat in JavaScript. This implementation was marginally faster than that the one built into V8 because it cheated, and made mistakes. My patch removes that code, and uses the parseFloat builtin.

Parsing floating point numbers is a very tricky topic, and writing a test to ensure that a given parser does not make mistakes is a technically challenging task. I do think, however, that assuming that the JS engine implements it correctly is a reasonable assumption.

In retrospect, I'm not sure why I chose to use parseFloat() instead of Number(). I should have documented. It's plausible that Number() would be a few nanoseconds faster.

wellwelwel:
> Parsing floating point numbers is a very tricky topic, and writing a test to ensure that a given parser does not make mistakes is a technically challenging task. I do think, however, that assuming that the JS engine implements it correctly is a reasonable assumption.

I agree. If there were to be a method that competes with **V8**, I think a dedicated library would be more appropriate, since it is not so simple to test and cover all scenarios.

sidorares:
Closing due to a similar change in https://github.com/sidorares/node-mysql2/pull/4135

PR #1539, opened 2022-03-22
Add taggedExecute helper

sidorares:
RFC at this stage, but I thing the feature was mentioned / requested few times in the past and looks like a safe and useful addition

At this stage I intentionally adding this to "real" prepared statements only ( AKA `.execute()` ) and also to promise wrapper only

examples:

```js
const pool = mysql.createPool({ user: 'root', password: 'test' });
const $ = pool.taggedExecute();
const input = 'test\" --';
const rows = await $`select ${input} as solution`;
```

```js
const connection = await mysql.createConnection({ user: 'root', password: 'test' });
const $ =connection.taggedExecute();
const input = 'test\" --';
const rows = await $`select ${input} as solution`;
```

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -124,6 +124,13 @@ class PromiseConnection extends EventEmitter {
     });
   }
 
+  taggedExecute() {
+    return (strings, ...keys) => {
+      const statementQuery = strings.join('?');
+      return this.execute(statementQuery, keys).then(result => result[0])
+    }
+  }
+
   end() {
     return new this.Promise(resolve => {
       this.connection.end(resolve);
@@ -370,6 +377,13 @@ class PromisePool extends EventEmitter {
         corePool.execute(sql, done);
       }
     });
+  }
+
+  taggedExecute() {
+    return (strings, ...keys) => {
+      const statementQuery = strings.join('?');
+      return this.execute(statementQuery, keys).then(result => result[0]);
+    }
   }
 
   end() {


sidorares:
Something that I don't like but not sure how to handle well in a backwards compatible way.

With addition of tagged templates based PS there will be 4 ways of sending parametrised sql, which potentially makes everything very confusing:

```js
// 1 - client side interpolation, PS-like but not "real" prepared statements:
const [rows1, columns1] = await connection.query('select ? as solution', ["parameter"]);

// 2 - real PS, prepare(sql) + execute(param) command the first time its called, execute(param) after
//  api aims to mimic (1) but due to limitations of prepared statements there are differences
const [rows2, columns2] = await connection.query('select ? as solution', ["parameter"]);

// 3 - named placeholders. Same as (2) but a bit easier to read
const [rows3, columns3] = await connection.query('select :parameter as solution', { parameter: "parameter"});

// 4 - tagged template
const rows = await $`select ${"parameter"} as solution`;
```

Is having (3) and (4) too much? Given that (3) is relatively unknown, maybe we should deprecate it in favour of (4) ?

sidorares:
might be a good idea to model api to match https://github.com/porsager/postgres#await-sql---result

not sure though if it belongs to base driver or independent wrapper library

sidorares:
add to docs: this extension should provide syntax highlighting with no additional config - https://marketplace.visualstudio.com/items?itemName=frigus02.vscode-sql-tagged-template-literals

sidorares:
I'm leaning towards _not_ merging this and creating separate package instead

sidorares:
Existing external modules as of Nov 2023:
- https://github.com/nearform/sql
- https://github.com/blakeembrey/sql-template-tag

PR #1554, opened 2022-04-07
Add identify verification option for TLS

dbussink:
Right now the hostname presented on a TLS certificate is never validated. This means that man-in-the-middle attacks are trivial when using for example public CAs to sign certificates.

The issue is that anyone can have a public CA sign a certificate for a hostname they own, which is key to the security model there. But with the current functionality in the `mysql2` driver, the hostname is never verified.

This means an attacker could do a man in the middle attack and only has to present a signed certificate but doesn't have to match the hostname.

It does not only apply to public CAs in system roots, but also for example Amazon RDS where a single root chain is also used.

The changes here add an option to explicitly verify the identity on the certificate. This option is available in other MySQL drivers as well, such as `libmysqlclient` and everything based on that (often drivers for Ruby, Python etc.). Also other languages like Go, Java, .Net etc. provide an option to have strict identity verification.

Ideally of course, this validation would be the default and always enforced, but that also would likely be a significant breaking change, so right now in the code here the setting is opt-in for having a secure connection to MySQL.

Additionally, it adds SNI on the connection which is a best practice as well.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -342,6 +342,9 @@ class Connection extends EventEmitter {
       minVersion: this.config.ssl.minVersion
     });
     const rejectUnauthorized = this.config.ssl.rejectUnauthorized;
+    const verifyIdentity = this.config.ssl.verifyIdentity;
+    const host = this.config.host;
+
     let secureEstablished = false;
     const secureSocket = new Tls.TLSSocket(this.stream, {
       rejectUnauthorized: rejectUnauthorized,
@@ -349,6 +352,9 @@ class Connection extends EventEmitter {
       secureContext: secureContext,
       isServer: false
     });
+    if (typeof host === 'string') {
+      secureSocket.setServername(host);
+    }
     // error handler for secure socket
     secureSocket.on('_tlsError', err => {
       if (secureEstablished) {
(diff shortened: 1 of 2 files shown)


dbussink:
@sidorares Any feedback on this change?

dbussink:
> @sidorares Any feedback on this change?

@sidorares Any update on this one?

dbussink:
> > @sidorares Any feedback on this change?
> 
> @sidorares Any update on this one?

@sidorares Any updates on this one?

sidorares (approved):

sidorares:
@dbussink sorry for the wait, ( and thanks for your persistence )

Your change looks good to me and I plan to have it as part of v3.0.0 ( and probably enable by default in 3.1.0 )

dbouchierarcad:
Hello , do you have an expected delivery date for the correction of this vulnerability? I understand you'll fix it in 3.1.0? Thanks

sidorares:
@dbouchierarcad I'll do one more 3.0.0 rc with this change, should be part of final v3.0

sidorares:
in terms of ETA - the only blocker is release-please configuration, I can manually publish `rc.2` any time

Issue #1384, opened 2021-09-12
decodeURIComponent after parsing URL for initial connection


kwanhs:
It appears the URL package is not decoding some special characters correctly. In my case, it was `@` sign used in the password.
Regardless of whether `@` or `%40` (its encoded counterpart) is used, the attributes obtained from parsed URL is still `%40`.

sidorares:
which version are you using? I'd expect this issue would be fixed by https://github.com/sidorares/node-mysql2/pull/1360 can you double check the version you have is v2.3.0?

kwanhs:
Thanks for your quick response! From my package.json: `"mysql2": "^2.3.0",`

sidorares:
can you clarify `the attributes obtained from parsed URL is still %40` part? Are you able to connect using url where password is urlencoded?

kwanhs:
It was the `password` that remained url-encoded

sidorares:
@acdibble should [this line](https://github.com/sidorares/node-mysql2/pull/1360/files#diff-1978d11697f58179bc32b66c6ac7f21cbdfd6649a8f013248a1f9feee25858feR241) be `password: unescape(parsedUrl.password)` or something similar?

acdibble:
We might have to pass both username and password to `decodeURIComponent`. However, if a password or username contains `%` and it is not properly escaped before building the URL, `decodeURIComponent` will throw an error, e.g.:

```javascript
> password = 'pass!@$%^&*+()\\word:'
'pass!@$%^&*+()\\word:'
> conn = new URL(`test://user:${password}@www.example.com/path`)
URL {
  href: 'test://user:pass!%40$%%5E&*+()%5Cword%3A@www.example.com/path',
  origin: 'null',
  protocol: 'test:',
  username: 'user',
  password: 'pass!%40$%%5E&*+()%5Cword%3A',
  host: 'www.example.com',
  hostname: 'www.example.com',
  port: '',
  pathname: '/path',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
> decodeURIComponent(conn.password)
Uncaught URIError: URI malformed
    at decodeURIComponent (<anonymous>)
> conn = new URL(`test://user:${encodeURIComponent(password)}@www.example.com/path`)
URL {
  href: 'test://user:pass!%40%24%25%5E%26*%2B()%5Cword%3A@www.example.com/path',
  origin: 'null',
  protocol: 'test:',
  username: 'user',
[...]

PR #1560, opened 2022-04-18
fix uri connection unavaliable

yuhj86:
when password contains some characters

e.g. `mysql://root:pa%ss=wo%rd@localhost/test`

#1384


yuhj86:
i missed  `path percent-encode set `

PR #1565, opened 2022-04-24
Add type declarations for Prepare & PrepareStatementInfo

### LINKED ISSUE #1564: Missing prepare/unprepare
Looked at the documentation at https://github.com/sidorares/node-mysql2/blob/master/documentation/Prepared-Statements.md when I try to run a prepared statement

```js
connection.execute('statement', (error, results) => {

});

connection.unprepare('statement');
```

I get the following error

`
 error TS2339: Property 'unprepare' does not exist on type 'Connection'.
`

The prepare method also doesn't exist

zordtk:
This adds type declarations for Prepare & PrepareStatementInfo and then adds prepare and unprepare to the Connection declarations. Fixed #1564

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -186,6 +186,14 @@ export interface ConnectionOptions extends mysql.ConnectionOptions {
   };
 }
 
+export interface ConnectionConfig extends ConnectionOptions {
+  mergeFlags(defaultFlags: string[], userFlags: string[] | string): number;
+  getDefaultFlags(options?: ConnectionOptions): string[];
+  getCharsetNumber(charset: string): number;
+  getSSLProfile(name: string): { ca: string[] };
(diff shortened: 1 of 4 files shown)


sidorares:
I have to check other command interfaces but there is no `determinePacket` method in mysql2 commands, my guess is that its part that was originally in `mysqljs/mysql` typings

Also, what was the reason to remove `ConnectionConfig` interface?

zordtk:
> Also, what was the reason to remove `ConnectionConfig` interface?

Didn't realize I did.

> I have to check other command interfaces but there is no `determinePacket` method in mysql2 commands, my guess is that its part that was originally in `mysqljs/mysql` typings

I actually copied that from the type declarations from the Query command (typings/mysql/lib/protocol/sequences/Query.d.ts)

jmordica:
Can this get merged?

sidorares:
> Can this get merged?

yes, thanks for the ping

PR #1576, opened 2022-05-14
Fix Simplified Chinese document button parsing error in Github Pages

WenisOK:
I found an issue with the Simplified Chinese documentation hyperlinks on Github Pages. It was originally a hyperlink, but was parsed into a table. It is caused by `|`, so i removed it.


WenisOK:
image

Issue #1585, opened 2022-05-30
CI: coverage-comment-action premissions problem


yuxizhe:
https://github.com/sidorares/node-mysql2/blob/ab3feb75bb4af1f6d288c0494a99aa2483ac24d5/.github/workflows/ci-linux.yml#L118-L122

there is a bug in coverage-comment-action
https://github.com/ewjoachim/coverage-comment-action/issues/26

external users creating PRs will trigger actions whose github token won't have the right to post comments.

so external users PRs always fail. 

eg: https://github.com/sidorares/node-mysql2/runs/6649645051?check_suite_focus=true

```
github.GithubException.GithubException: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue-comment"}
Operating on refs/pull/1584/merge
Commenting on the coverage on PR 1584
Previous coverage results not found, cannot report on evolution.
```

sidorares:
I'm thinking to completely disable this workflow unfortunately
Maybe I'll give https://github.com/ewjoachim/python-coverage-comment-action a try first

sidorares:
added `continue-on-error: true` via #1697 , PRs from external users won't have a coverage message but at least should not see failure if everything else is ok

PR #1589, opened 2022-07-02
fix: webpack projects no longer show warning for cardinal dependency

mikedidomizio:
I made a [comment here on a fix](https://github.com/sidorares/node-mysql2/issues/846#issuecomment-1172958633) for the cardinal optional dependency issue with Webpack.  I decided to go ahead with the PR as I won't have the configuration set up for too long on my end.

This fixes the issue for at least Webpack, I haven't checked with other bundlers.

I've added a video below to show how my project running Next12/Webpack5 would show the warning, and how after the fix the warning would be gone.  On the right is the node-mysql2 project, and the top left console is my project which has it npm linked so I can test changes.

https://user-images.githubusercontent.com/5728044/177018917-d3ff624a-37c8-4651-8de8-61316c762098.mov

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -25,7 +25,9 @@ exports.srcEscape = srcEscape;
 let highlightFn;
 let cardinalRecommended = false;
 try {
-  highlightFn = require('cardinal').highlight;
+  // the purpose of this is to prevent projects using Webpack from displaying a warning during runtime if cardinal is not a dependency
+  const REQUIRE_TERMINATOR = '';
+  highlightFn = require(`cardinal${REQUIRE_TERMINATOR}`).highlight;
 } catch (err) {
   highlightFn = text => {
     if (!cardinalRecommended) {


mikedidomizio:
Not sure why the job failed here.  I attempted the [same PR in my fork](https://github.com/mikedidomizio/node-mysql2/pull/1) and succeeded.  I see this [other PR](https://github.com/sidorares/node-mysql2/pull/1590) hit the same issue.  Maybe the limit on GitHub actions minutes?

sidorares:
> Not sure why the job failed here. I attempted the [same PR in my fork](https://github.com/mikedidomizio/node-mysql2/pull/1) and succeeded. I see this [other PR](https://github.com/sidorares/node-mysql2/pull/1590) hit the same issue. Maybe the limit on GitHub actions minutes?

known problem - #1585
I might just disable this workflow for now unless there is a better fix

Issue #1044, opened 2019-10-23
In caching sha2 password xor, b length can be less than a.


normano:
https://github.com/sidorares/node-mysql2/blob/93df7a8f5c7f56d80285ebdd001508567cbe84e9/lib/auth_plugins/caching_sha2_password.js#L35

You could use xorRotating func to fix:
https://github.com/normano/node-mysql2/blob/72aaf4a07680d26edb5ac6bb828027207e97e004/lib/auth_41.js#L137

sidorares:
Good catch @normano !
I wonder if mysql server actually expects xor rotated this way. Are you able to test this scenario?
Scramble is usually 20 bytes, this means you need to create a user with > 20 char password to be able to test

normano:
There is a link to Preparing Your MySQL Community Connector in the code but
the site is not currently working. You can search for "MySQL 8
caching_sha2_password xor rotated in Google to find reference to this.

normano:
https://webcache.googleusercontent.com/search?q=cache:CBqPiiqmUO4J:https://insidemysql.com/preparing-your-community-connector-for-mysql-8-part-2-sha256/+&cd=1&hl=en&ct=clnk&gl=us

sidorares:
Thanks, do you want to create a PR with this change? Note that currently xor function is not shared between different auth plugins, maybe worth refactoring now

sidorares:
I think this is only affects caching_sha2_password where seed is xored against password, in all 4.1 auth scenarios both arguments to xor are sha1 hashes exactly 32 bytes.

I can confirm that current code fails to login for a user with `01234567890123456789` ( or longer ) password and logs in no issues with rolling xor  ( Mysql 8.0.17, caching_sha2_password default auth )

normano:
Just got home, I can pull in your changes and attempt to implement.

PR #1592, opened 2022-07-14
Update sha256_password.js

cavenator:
In short, auth_plugin `sha256_password` has the problem as defined in issue #1044.  This swaps `xor` with `xorRotating` just like `caching_sha2_password`.


sidorares:
Thanks @cavenator !

PR #1604, opened 2022-08-10
Ability to use `createServer` in typescript

jmordica:
When using typescript and/or ESM it's nice to be able to:
```js
import { createServer } from 'mysql2';
```

This PR hoists the `createServer` module.

Let me know if this is way is unacceptable and I can possibly define this within `typings/mysql/...`


sidorares:
Thanks @jmordica !

PR #1584, opened 2022-05-29
fix(type): add promise missing typings and simplify code

yuxizhe:
add promise wrapper missing typings and simplify code.

fix [ #1184](https://github.com/sidorares/node-mysql2/issues/1184)   #804

PR #1606, opened 2022-08-10
More complete way of adding typings for the `Server` module

jmordica:
- Conformed to the way typings are being added currently in the project
- Fixed an issue importing `PrepareStatementInfo` (while we're at it)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,8 +7,6 @@ import {
 import * as mysql from './typings/mysql';
 export * from './typings/mysql';
 
-import Server from './lib/server.js';
-
 export interface Connection extends mysql.Connection {
   execute<
     T extends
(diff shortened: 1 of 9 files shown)


sidorares:
not sure why https://github.com/sidorares/node-mysql2/actions/workflows/code-coverage.yml is still failing, I disabled in manually

Could you add some typings unit tests for types covered in this pr? ( And perhaps ones discussed in https://github.com/sidorares/node-mysql2/pull/1584#issuecomment-1210935539 )

jmordica:
Looks like checks are failing because the package-lock.json isn't synced with the package.json. I can `npm install` and add the `package-lock.json`

jmordica:
Now that the actual errors are fixed related to typings, i'll add the test for the Server typings.

jmordica:
@sidorares I've added a new test environment for typings specifically and went ahead and included the `Server` module. So going forward, anyone that creates new typings will be encouraged to add a test in `typings/test` and modify the `tests.json`.

Let me know if this seems over-engineered or if you have any questions. Based on your response, I can also add proper tests for the changes in PR #1584.

PR #1610, opened 2022-08-17
Types, tests, and commands Oh my! 🌪️ 

jmordica:
This PR invalidates #1584 (based on whether or not this PR can be merged). It adds the missing typings and aims to simplify things a bit.

It also takes care of added tests for the common types, fixes a packet issue related to `prepare`, and adds additional listeners for both `prepare` and `execute` when using the module as a MySQL proxy.

These changes are necessary in order to build a more complete (batteries included) MySQL proxy (in typescript) with support for prepared statements (using mysql2 module as a client along with other common clients).

### DIFF
--- a/lib/commands/server_handshake.js
+++ b/lib/commands/server_handshake.js
@@ -71,7 +71,7 @@ class ServerHandshake extends Command {
 
   _isStatement(query, name) {
     const firstWord = query.split(' ')[0].toUpperCase();
-    return firstWord == name;
+    return firstWord === name;
   }
 
   dispatchCommands(packet, connection) {
--- a/lib/commands/server_handshake.js
+++ b/lib/commands/server_handshake.js
@@ -122,7 +122,13 @@ class ServerHandshake extends Command {
       case CommandCode.QUERY:
         if (connection.listeners('query').length) {
           const query = packet.readString(undefined, encoding);
-          connection.emit('query', query);
+          if (this._isStatement(query, 'PREPARE') || this._isStatement(query, 'SET')) {
+            connection.emit('stmt_prepare', query);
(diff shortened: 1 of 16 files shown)


jmordica:
Looking into the lint issue.

lib/commands/server_handshake.js
```diff
@@ -75,6 +76,90 @@ class ServerHandshake extends Command {
...
+            code: Errors.HA_ERR_INTERNAL_ERROR,
+            message:
+              'No query handler for prepared statements.'
+          });
+        }
+        break;
+      case CommandCode.STMT_EXECUTE:
+        if (connection.listeners('stmt_execute').length) {
+          console.log(packet)
+          const stmtId = packet.readInt32();
```
sidorares, line 94: this should be moved to `PrepareStatement.fromPacket`, added [here](https://github.com/sidorares/node-mysql2/blob/5022b54b012872d062b260a0cafb2a134139dd30/lib/packets/prepare_statement.js#L24) similar to https://github.com/sidorares/node-mysql2/blob/5022b54b012872d062b260a0cafb2a134139dd30/lib/packets/prepare_statement.js#L24

jmordica: This was added to `/lib/packets/execute.js`. Hopefully this is good enough.

jmordica:
@sidorares let me know what I'm missing here.

Thanks!

lib/commands/server_handshake.js
sidorares: thanks! ( yep, meant execute packet, not prepare )

sidorares:
thanks @jmordica , great work!

PR #1621, opened 2022-08-31
Adds decodeURIComponent in ConnectionConfig#parseUrl

### LINKED ISSUE #1384: decodeURIComponent after parsing URL for initial connection

Paultje52:
Currently, you cannot use special characters like `@` in usernames or passwords. This is because those characters are getting uri encoded by the URL nodejs package.

This PR adds `decodeURIComponent()` to `ConnectionConfig#parseUrl`, for `options.user` and `options.password`. This way, users can use special characters like `@` in usernames and passwords!

This also resolves #1384.
It also makes sure mysql2 can be used in nodejs apps hosted by pterodactyl with databases. When creating a database in pterodactyl, it generates a password automatically. That password also contains special characters, which make them useless when using mysql2.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -7,9 +7,9 @@
 
 'use strict';
 
-const { URL } = require('url');
 const ClientConstants = require('./constants/client');
 const Charsets = require('./constants/charsets');
+const { URL } = require('url');
 let SSLProfiles = null;
 
 const validOptions = {
@@ -249,13 +249,31 @@ class ConnectionConfig {
   }
 
   static parseUrl(url) {
-    const parsedUrl = new URL(url);
+    // First, split the url. Take the mysql:// part. The username is after that.
+    // Then, split at the : - usernames cannot contain a :
+    let user = url.split("mysql://")[1];
+    if (user) user = user.split(":")[0];
+    else throw new Error("Invalid connection string");
+
+    // Now, get the password. Remove the mysql://<USER>: part from the connection string,
+    // Continue until the last @ - that indicates the host is after it.
+    let password = url.substring(8 + user.length + 1);
+    if (password) {
+      const passwordParts = password.split("@");
+      password = passwordParts.splice(0, passwordParts.length-1).join("@");
+    } else throw new Error("Invalid connection string");
+
+    // Now, let node handle the rest of the url parsing (for the host, port and database)
+    // Also, remove the username and password and replace them with placeholders.
+    // This way, we do not get an error when the username or password contains special characters. 
+    const parsedUrl = new URL(url.replace(user, "_PLACEHOLDER_").replace(password, "_PLACEHOLDER_"));
+
     const options = {
       host: parsedUrl.hostname,
       port: parsedUrl.port,
       database: parsedUrl.pathname.slice(1),
-      user: decodeURIComponent(parsedUrl.username),
-      password: decodeURIComponent(parsedUrl.password)
+      user,
+      password
     };
     parsedUrl.searchParams.forEach((value, key) => {
       try {
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -254,8 +254,8 @@ class ConnectionConfig {
       host: parsedUrl.hostname,
       port: parsedUrl.port,
       database: parsedUrl.pathname.slice(1),
-      user: parsedUrl.username,
-      password: parsedUrl.password
+      user: decodeURIComponent(parsedUrl.username),
+      password: decodeURIComponent(parsedUrl.password)
     };
     parsedUrl.searchParams.forEach((value, key) => {
       try {


kykungz:
@Paultje52 I'm really looking forward to seeing this PR merged, as I cannot use `mysql2` with connection string at all because my password contains special characters (generated) to be considered a strong password.

I have tested with [this list](https://kb.iu.edu/d/aphg) of available characters in MySql strong passwords (`~!@$^&*()_-+={}[]<>,.;':|#?/%`), here's the result:

For this list `~!@$^&*()_-+={}[]<>,.;':|` (without `#`, `?`, `/`, `%`), it is successfully parsed with no errors:
```ts
const password = "~!@$^&*()_-+={}[]<>,.;':|"
const connectionString = `mysql://username:${password}@127.0.0.1/database`
const parsedUrl = new URL(connectionString)
const parsedPassword = decodeURIComponent(parsedUrl.password)

console.log(parsedPassword === password) // true
```

But if the password is including one of `#`, `?`, `/`, `Invalid URL` error will be thrown at `new URL( ... )` due to the fact that `#`, `?`, `/` is a path identifier of a URL.
```ts
const password = "~!@$^&*()_-+={}[]<>,.;':|" + "#?/"
const connectionString = `mysql://username:${password}@127.0.0.1/database`
const parsedUrl = new URL(connectionString) // Invalid URL
```

Also if the password is including `%`, `URI malformed` will be thrown at `decodeURIComponent( ... )` because `%` is not encoded properly from `new URL( ... )`.
```ts
const password = "~!@$^&*()_-+={}[]<>,.;':|" + "%"
const connectionString = `mysql://username:${password}@127.0.0.1/database`
const parsedUrl = new URL(connectionString)
const parsedPassword = decodeURIComponent(parsedUrl.password) // URI malformed
```

However we can fix the `%` problem by using `encodeURI( ... )` before passing it to `new URL( ... )`
```diff
- const parsedUrl = new URL(url);
+ const parsedUrl = new URL(encodeURI(url));
const options = {
  host: parsedUrl.hostname,
  port: parsedUrl.port,
  database: parsedUrl.pathname.slice(1),
  user: decodeURIComponent(parsedUrl.username),
  password: decodeURIComponent(parsedUrl.password)
};
```
This will make `%` available in the password:
```ts
const password = "~!@$^&*()_-+={}[]<>,.;':|" + "%"
const connectionString = `mysql://username:${password}@127.0.0.1/database`
const parsedUrl = new URL(encodeURI(connectionString))
const parsedPassword = decodeURIComponent(parsedUrl.password)

console.log(parsedPassword === password) // true
```

@sidorares Despite not covering every edge case (`#`, `?`, `/` still unusable), we should still push forward this PR as it has already covered a wide range of special characters (`~!@$^&*()_-+={}[]<>,.;':|%`) while the current version does not do well with them.

| Version | ✅ Supported                   | ❌ Unsupported      |
| ------- | :---------------------------- | ------------------ |
| Current | `~!$&*()_-+,.%'`              | `@^={}[]<>;:\|#?/` |
| This PR | `~!$&*()_-+,.%'@^={}[]<>;:\|` | `#?/`              |

sidorares:
Thanks @Paultje52 !

Would you be able to add a unit test for this?

Paultje52:
@kykungz and @sidorares I changed the parser to support all special characters. Let me know what you think!

sidorares:
@Paultje52 whats the reason you decided to switch to a manual parsing? Can you give an example where `password = decodeURIComponent(url.password); username = decodeURIComponent(url.username);` would not work?

kykungz:
@sidorares @Paultje52 After digging around with MySQL's [URI-Like Connection Strings](https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html#connecting-using-uri), I found that both solution (https://github.com/sidorares/node-mysql2/pull/1621/commits/f08b3be0182bb6e32800ab5900597c76403bea1a and https://github.com/sidorares/node-mysql2/pull/1621/commits/ed0cb99b78d798cd7d04a24e2faa144cdc9312fb) will never cover every use-case.

For example, using manual parsing will cover more use-cases, but it will open up another limitation of:

> username cannot contain `:`

while MySQL natively supported `:` in a username.

Although there will be only a few people who hit this wall, it's not great that a widely-used lib like this cannot match what MySQL can support.

It is [explicitly stated](https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html#connecting-using-uri) in MySQL docs that you **must** encode all special characters before using it:
> ### Important
> Percent encoding must be used for reserved characters in the elements of the URI-like string. For example, if you specify a string that includes the @ character, the character must be replaced by %40. If you include a zone ID in an IPv6 address, the % character used as the separator must be replaced with %25.

So you cannot use `mysql://username:p@$$w&rd@127.0.0.1:3306/database` natively with MySQL's tools:
```sh
mysqlsh --uri "mysql://username:p@$$w&rd@127.0.0.1:3306/database"
# ❌ Error - Invalid URI
```

You must `encodeURIComponent` manually for **each** field:
```sh
mysqlsh --uri "mysql://username:p%40%24%24w%26rd@127.0.0.1:3306/database"
# ✅ Connected
```

It wouldn't make sense if you can connect to MySQL through [node-mysql2](https://github.com/sidorares/node-mysql2) with `p@$$w&rd`, but couldn't do so on native MySQL tools. That sounded like a magic feature of this lib, but it's a feature that causes other unsolvable bugs.

So I think we should re-consider this and make this lib consistent with MySQL's standard, to eliminate all confusions about the lib vs actual MySQL. Instead of **auto parsing connection string**, we should be thinking of how to **support encoded connection string**.

We can simply revert back to https://github.com/sidorares/node-mysql2/pull/1621/commits/f08b3be0182bb6e32800ab5900597c76403bea1a, and maybe add more `decodeURIComponent` so that we also support encoded [IPv6 Zone ID](https://howdoesinternetwork.com/2013/ipv6-zone-id) in `host` and special characters in `database`:
```diff
const options = {
-  host: parsedUrl.hostname,
+  host: decodeURIComponent(parsedUrl.hostname),
   port: parsedUrl.port,
-  database: parsedUrl.pathname.slice(1),
+  database: decodeURIComponent(parsedUrl.pathname.slice(1)),
  user: decodeURIComponent(parsedUrl.username),
  password: decodeURIComponent(parsedUrl.password)
};
```
After that we can add more documentation on how to use URI-Like Connection Strings (every special character must be encoded) by referring to [this document](https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html#connecting-using-uri).

What do you think?

sidorares:
@kykungz I'm happy with the proposed solution ( throwing in few `decodeURIComponent` )

while we are at this, might worth checking/adding support for socket path in uri connection:

>  socket: The path to a Unix socket file or the name of a Windows named pipe. Values are local file paths. In URI-like strings, they must be encoded, using either percent encoding or by surrounding the path with parentheses. Parentheses eliminate the need to percent encode characters such as the / directory separator character. For example, to connect as root@localhost using the Unix socket /tmp/mysql.sock, specify the path using percent encoding as root@localhost?socket=%2Ftmp%2Fmysql.sock, or using parentheses as root@localhost?socket=(/tmp/mysql.sock). 

I think at this stage its more important to keep compatibility with mysqljs/mysql rather than mysql cli or other tools but ideally we should be following the spec in the https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html

estyrke:
Any progress on this one? Running into this right now, unable to connect to my database at all due to crazy passsword...

estyrke:
I ended up hand-patching connection_config.js and it worked perfectly, FWIW. :)

Issue #1629, opened 2022-09-08
Migrate deque from denque to Js-sdsl Deque


ZLY201:
Hey! I'm the developer of Js-sdsl. Official website: https://js-sdsl.github.io/.

Now, we published the version 4.1.4.

I see you are using denque.

In [benchmark](https://js-sdsl.github.io/#/test/benchmark-result?id=deque), we have confirmed that Js-sdsl is **several times faster** than `denque` and **nearly equal** to `Array.push` in the case of push elements.

We would like to invite you to migrate deque related functions to Js-sdsl v4.1.4 and I am willing to submit a pull request for this change.

Looking forward to your reply! :D

sidorares:
Thanks for the offer, I'll try to evaluate Js-sdsl!
It would be interesting to see real world impact on performance with some typical patterns of connection / prepared statements / pool use.

ZLY201:
Hey, I done th migrate.

See https://github.com/sidorares/node-mysql2/pull/1633.

How can I know the result about benchmark?

ZLY201:
@js-sdsl/deque published!

sidorares:
@ZLY201 I don't want to change core functionality that already works fine without convincing benchmarks ( in the context of drivers use case ). I believe Js-sdsl is more performant than denque but need to see some visible difference (with all other components factored in). if its below statistical error not worth the change

It's not that I'm against this change, just until there is a good benchmark its a low priority for me

ZLY201:
> @ZLY201 I don't want to change core functionality that already works fine without convincing benchmarks ( in the context of drivers use case ). I believe Js-sdsl is more performant than denque but need to see some visible difference (with all other components factored in). if its below statistical error not worth the change
> 
> It's not that I'm against this change, just until there is a good benchmark its a low priority for me

Thanks to your reply!

We are trying to find a good way to test the performance of Js-sdsl in MySQL like database connections, including Redis and MongoDB.

This will be our main work in the next stage. So make this progress to be pending is reasonable.

trasherdk:
@ZLY201 you got your definitions wrong
![image](https://user-images.githubusercontent.com/5003891/204112320-98617548-ed0a-48db-b427-204eaf8f2499.png)

It's `stack` - first in last out, as in `push` `pop`
and `queue` first in first out, as in `push` `shift`.

ZLY201:
[...]

PR #1633, opened 2022-09-11
perf: migrate deque to js-sdsl

ZLY201:
Linked issue: https://github.com/sidorares/node-mysql2/issues/1629.


sidorares:
not worth doing given not obvious benefit and security audit overhead

PR #1642, opened 2022-09-30
renamed done to onResult

### LINKED ISSUE #1638: behavior of connection.end() on closed connection differs from mysqljs/mysql.
In mysqljs/mysql, calling:
```
connection.end();
connection.end( err => { if ( err ) console.info( err.message ); } );
```
results in:
```
Cannot enqueue Quit after invoking quit
```
In mysql2, the second call to `end()` doesn't invokes the callback.

After dropping in mysql2 to replace mysql, this was detected by a unit test expecting a call to `end()` on a closed connection to result in an error.

Andrew-Opplaud:
Resolves #1638 to the extent that it invokes callback, although the mysql2 error message is "Can't add new command when connection is in closed state" instead of the more specific mysqljs/mysql error message "Cannot enqueue Quit after invoking quit.".

### DIFF
--- a/lib/commands/quit.js
+++ b/lib/commands/quit.js
@@ -7,7 +7,7 @@ const Packet = require('../packets/packet.js');
 class Quit extends Command {
   constructor(callback) {
     super();
-    this.done = callback;
+    this.onResult = callback;
   }
 
   start(packet, connection) {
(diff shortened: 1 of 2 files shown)


sidorares:
Thanks @Andrew-Opplaud! Happy to merge just this change but would be great to have a unit test that fails without the above change and passes with

Andrew-Opplaud:
added!

Issue #1631, opened 2022-09-08
Releases?


ErisDS:
Hi there,

My team and I (Ghost) are trying to understand why this repo doesn't get releases?

There are regular updates to the repo but noone is benefitting from the changes - the last release was back in November 2021. At which point the releases had been every couple of weeks.

We have users specifically waiting on https://github.com/sidorares/node-mysql2/pull/1438, which was merged 10days after the last release, now almost a year ago. 

Is there a problem with the release process? Something we could possibly help with? cc @daniellockyer

We really appreciate all of the hard work going on here - we would love to make use of it!

sidorares:
Mainly because I'm very busy and publishing is currently not fully automatic. I'm going to try to find some time and configure release-please or something similar over the weekend

sidorares:
Re help - any release workflow automation would be appreciated.  Also one blocker is coverage check that is failing on PRS where source branch outside of this repo. Need to fix the issue or remove the check

ErisDS:
Hey @sidorares, I appreciate how difficult these things are when they're not automated. I believe @daniellockyer fixed a similar issue with PR coverage checks on our repos so hopefully he can take a look at that next week. 

He's also a dab hand at workflow automation and I'm sure he can help more.

daniellockyer:
Hey @sidorares 👋🏻 Apologies for stepping in much later. I'd be more than happy to help given we also fixed up the node-sqlite3 release process too 🙂 

I see you've pushed a commit for the coverage check permissions. Shall I take a look into the release-please GHA?

sidorares:
Thanks @daniellockyer!

yes, I was thinking to add release-please

I'll describe current manual release process and we can discuss what automated would look like:

1) clone master locally, go through PR/commits and extract what I think needs to be communicated into https://github.com/sidorares/node-mysql2/blob/master/Changelog.md 
2) commit changelog as vX.Y.Z changes, run `npm version patch` ( or minor/major based on the changes )
3) push changelog commit to master + commit and tag generated by `npm version`
4) `npm publish`

Sometimes where I feel changes are too big there is an `-rc` version publish followed by few days of waiting.

I still prefer manual changelog, my ideal workflow would be this:

- create issue with "release" in the title
- GH action pulls all the commits and PRs in a commit
- I ( or somebody else in the allowed list) picks relevant changelog / adds edits
- steps similar to 2-4 are performed automatically ( + extra step - create github release )
- issue is closed if all goes well

[...]

PR #1647, opened 2022-10-12
ci: added workflow to integrate release-please

daniellockyer:
refs https://github.com/sidorares/node-mysql2/issues/1631

- this workflow integrates Google's `release-please-action`, which seems like the best option for using release-please and GHA
- if a release was created, it'll also automatically publish to npm

---

@sidorares There is a [lot of configuration](https://github.com/google-github-actions/release-please-action#configuration) available for the Action. Might be worth giving that list a check before merging to ensure it is set up how you would like

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+on:
+  push:
+    branches:
+      - master
+name: release-please
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          release-type: node
+          package-name: mysql2
+          changelog-path: 'Changelog.md'
+
+      - uses: actions/checkout@v3
+        if: ${{ steps.release.outputs.release_created }}
+
+      - uses: actions/setup-node@v3
+        if: ${{ steps.release.outputs.release_created }}
+        with:
+          node-version: 16
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: npm ci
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+          node-version: 16
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: npm ci
+        if: ${{ steps.release.outputs.release_created }}
+
+      - run: npm publish
+        if: ${{ steps.release.outputs.release_created }}
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
daniellockyer, line 32: Not sure what you've named the token in the GitHub interface so please check this 🙂

sidorares: image

sidorares:
Thanks @daniellockyer !

What would be release process with current setup ( more specifically, what conditions set `steps.release.outputs.release_created` to true )

daniellockyer:
@sidorares From the [docs](https://github.com/google-github-actions/release-please-action#automating-publication-to-npm):

> With a few additions, the Release Please action can be made to publish to npm when a Release PR is merged:

So I think the release process would be:

1. collect all the commits in `master`
2. the release PR is created along the way
	1. it'll also contain the changes to changelog.md - but you should be good to push your edits to it before merging
4. Merge the PR
5. GitHub Release is created and the package is published to npm

Given `master` doesn't use conventional commits right now, I'm not sure what version it'll try to output, but we could always do a manual one and start from fresh 🙂

Issue #935, opened 2019-03-18
API incompatibility re aggregate functions sum and avg return string type instead of numeric


igolfs:
SELECT SUM(amt), AVG(amt)
    FROM MYTABLE;

 mysqljs returns numeric columns but mysql2 returns strings.

sidorares:
Is that using text binary or protocol ( `.query()` or `.execute()` ) ?

Also can you log field types? In your callback `function(err, rows, fields) {}` add `console.log(fields)`

igolfs:
Here's my DDL:

    create table test (qty int);
    insert into test (qty) values (1), (2), (3), (4), (5);

I ran 3 other tests with MySQLjs.query, MySQL2.query and MySQL2.execute for:

    SELECT sum(qty), avg(qty) FROM test
Functions min(), max() and count() behave correctly and return numeric values in all 3 tests.

Results for sum() and avg():
 
    ======================================================================
    MySQLjs query
    ======================================================================
    Results: [ RowDataPacket { 'sum(qty)': 15, 'avg(qty)': 3 } ]
    FieldPacket {
      catalog: 'def',
      db: '',
      table: '',
      orgTable: '',
      name: 'sum(qty)',
      orgName: '',
      charsetNr: 63,
      length: 33,
      type: 246,
      flags: 128,
      decimals: 0,
      default: undefined,
      zeroFill: false,
      protocol41: true }
      FieldPacket {
      catalog: 'def',
      db: '',
      table: '',
      orgTable: '',
      name: 'avg(qty)',
      orgName: '',
      charsetNr: 63,
      length: 16,
      type: 246,
      flags: 128,
      decimals: 4,
      default: undefined,
      zeroFill: false,
      protocol41: true }
    
    ======================================================================
    MySQL2 query
    ======================================================================
    Results: [ TextRow { 'sum(qty)': '15', 'avg(qty)': '3.0000' } ]
    { catalog: 'def',
      schema: '',
      name: 'sum(qty)',
      orgName: '',
      table: '',
      orgTable: '',
      characterSet: 63,
      columnLength: 33,
      columnType: 246,
      flags: 128,
      decimals: 0 }
      { catalog: 'def',
      schema: '',
      name: 'avg(qty)',
      orgName: '',
      table: '',
      orgTable: '',
      characterSet: 63,
      columnLength: 16,
      columnType: 246,
      flags: 128,
      decimals: 4 }
    
    ======================================================================
    MySQL2 execute
[...]

PR #1649, opened 2022-10-13
highlighted known incompatibility with node-mysql

timrobinson33:
I was caught by a quite subtle bug after upgrading from node-mysql to mysql2:

I was retrieving some `sum` values from the DB and then sorting by that field in JS code. because of the way I was processing the data after the sort, the sums were automatically converted from string to number. however, the sort itself was being done on the raw values.

In most of my tests the sum values were all 4 digits long so the output was identical when using mysql and mysql2. It was only when I happened to have a test where some of them where 3 digits long and some were 4 that the bug became apparent (999 would be sorted before 1000 when using mysql but "999" would be sorted after "1000" when using mysql2.

for this reason I think it warrants a warning on the main documentation page where it says it's "mostly compatible" to point out this potentially serious incompatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,9 @@ con.query({ sql: 'select 1 as foo, 2 as foo', rowsAsArray: true }, function(err,
 
 MySQL2 is mostly API compatible with [Node MySQL][node-mysql]. You should check their API documentation to see all available API options.
 
-If you find any incompatibility with [Node MySQL][node-mysql], Please report via Issue tracker. We will fix reported incompatibility on priority basis.
+One known incompatibility is that `sum` and `avg` results are returned as strings rather than numbers. This is deliberate to avoid loss of precision - see https://github.com/sidorares/node-mysql2/issues/935.
+
+If you find any other incompatibility with [Node MySQL][node-mysql], Please report via Issue tracker. We will fix reported incompatibility on priority basis.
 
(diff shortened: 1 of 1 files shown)


sidorares:
I wonder why this happens - could you check what is the return type on the `sum` or `avg` column? mysql2 does not cast `DECIMAL` type to a number ( this might be documented somewhere, need to link )

timrobinson33:
it seems to be investigated and quite clearly described (by yourself) here https://github.com/sidorares/node-mysql2/issues/935#issuecomment-474604432

> The value is stored internally and transmitted over the wire as string, I just wanted to preserve this property in the driver, it's up to the user to decide what to do with that value ( If we convert to number immediately there can be easily loss of precision )

This is why I'm not complaining about the fact that it is returned as string, but I think it should be pointed out prominently because the behavior is deliberately different from node-mysql driver and in a way that could cause subtle bugs.

sidorares:
Maybe we change wording in that paragraph focusing on the column types that are incompatible and the reasoning why it is like that, and then add example queries / function names that are known to return this types ( I'll let it to you to find best words @timrobinson33 , english is not my first language )

timrobinson33:
You're right. I have updated the description accordingly.

sidorares:
Thanks @timrobinson33 !

Issue #1650, opened 2022-10-13
await .ping() return [undefined, undefined]


ansleehk:
I have a question in regard to the ping function. I found out that the response of the async ping function is [undefined, undefined]

```js
async function a(){
    const b = await mysql.createConnection({
        "host": "127.0.0.1",
        "user": "admin",
        "password": "xxxxxxxx"
    });
    const c = await b.ping();
    console.log(c);
}

a();
```

`
Output: [ undefined, undefined ]
`

Is it that I missed something? I think I have implemented the functions correctly. The MYSQL server is running when I execute the above code. 

I cannot find any documentation in this module about how can I execute the ping function asynchronously so I purely add the await in front of ping().

sidorares:
this is because its reusing `makeDoneCb` helper from query/execute 

https://github.com/sidorares/node-mysql2/blob/ec4ed3ee04a94839216c50dbbb6d84efaa26b4fa/promise.js#L17

would you like to volunteer and improve this @ansleehk?

Either make a "void" version of makeDoneCb or just have similar code inline in ping() here https://github.com/sidorares/node-mysql2/blob/ec4ed3ee04a94839216c50dbbb6d84efaa26b4fa/promise.js#L165 

Also we need to check the change is covered in typings ( or correct types are added if there are none )

ansleehk:
Working on it!

ansleehk:
This is not an issue purely related to the makeDoneCb function. Even though I run the following code, the three values from the callback are still "undefined".

```js
  ping() {
    const c = this.connection;
    const localErr = new Error();
    return new this.Promise((resolve, reject) => {
        c.ping((a,b,c)=>{
          console.log(a,b,c);
        });
    });
  }
```
I think this issue is related to the so-called core package.
Therefore, I dig into the ping.js, which is located in lib/commands/, I think no values have been put into the callback functions, so the values of three are undefined.

sidorares:
why do you expect any data from ping callback?

This is how its called https://github.com/sidorares/node-mysql2/blob/33073c187fd26f81f8ecfe2114cf0b67db4f6983/lib/commands/ping.js#L30

This is `COM_PING` documentation

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_ping.html

ansleehk:
[...]

PR #1651, opened 2022-10-16
Fix async ping function #1650

ansleehk:
This is a fix related to #1650


sidorares:
small nit from the linter, change `(err) =>` to `err =>`

sidorares:
Thanks @ansleehk!

Issue #1334, opened 2021-06-06
Warning: got packets out of order. Expected 10 but received 0


chriscousinsdev:
Error has been occuring for a while, happens after a while where the bot will error. I recently upgraded from mysql to mysql2 to see if it would fix the problem, still nothing. When it errors nothing can be read from the database until the bot is restarrted

Flydexo:
I have the same issue but only a long time (10 hours) after launching the server. 
`const connection = mysql.createPool({
    host     : 'db',
    user     : 'test',
    password : '',
    database : ''
});

module.exports.connection = connection;`. I set the host to db because I use docker-compose. I export this connection/pool and I use it in my code with connection.query. But I have the same error. Does someone know why ?

bttger:
Possible duplicates
https://github.com/sidorares/node-mysql2/issues/528
https://github.com/sidorares/node-mysql2/issues/653
https://github.com/sidorares/node-mysql2/issues/778

ElijahShadbolt:
I have the same warning message. I'm testing a possible cause, but in case I forget to post a solution: I suspect it might be because I'm keeping one global variable for the connection, so maybe something is going wrong with my node http server calling `execute` and `query` asynchronously, while another request is using the same connection.

gramakri:
I am able to reproduce this when using `close()` instead of `end()` .

Here's a simple example using promises:
```
'use strict';
const mysql2p = require('mysql2/promise');

async function bugp() {
    let db = await mysql2p.createConnection({
        host: '172.18.0.2',
        user: 'root',
        password: 'helloworld'
    });

    await db.close(); // await db.end() makes it work
}
(async () => await bugp())();
```
Running the above with MySQL 8.0.27 (ubuntu 20.04), prints "Warning: got packets out of order. Expected 3 but received 0". If you replace db.close() with db.end() it without warning. Admittedly, I have no clue where I got db.close from since docs only refer to db.end()

If it's interesting, the equivalent callback code causes an error. Maybe that's why the above code is erroring differently:
```
'use strict';
const mysql2 = require('mysql2');
function bug(cb) {
    let db = mysql2.createConnection({
        host: '172.18.0.2',
        user: 'root',
        password: 'helloworld'
    });

    db.close(cb);
}

bug(() => {});
```

The above prints:
```
node:events:368
      throw er; // Unhandled 'error' event
      ^

Error [ERR_STREAM_WRITE_AFTER_END]: write after end
[...]

PR #1660, opened 2022-10-26
Fix issue #1334: warning "packets out of order" when handling ER_CLIENT_INTERACTION_TIMEOUT

huycn:
To fix issue #1334

From MySQL 8.0.24, an [error packet with code 4031](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_client_interaction_timeout) is sent to client before connection is closed. This packet has sequenceId == 0, thus the connection will issue a warning "packets out of order", which is not correct.

The changes here is just move the block checking & increasing sequence id to go after check for protocol error, since in case of error, we close the connection and return right away.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -407,19 +407,6 @@ class Connection extends EventEmitter {
       this._paused_packets.push(packet);
       return;
     }
-    if (packet) {
-      if (this.sequenceId !== packet.sequenceId) {
-        const err = new Error(
-          `Warning: got packets out of order. Expected ${this.sequenceId} but received ${packet.sequenceId}`
-        );
-        err.expected = this.sequenceId;
-        err.received = packet.sequenceId;
-        this.emit('warn', err); // REVIEW
-        // eslint-disable-next-line no-console
-        console.error(err.message);
-      }
-      this._bumpSequenceId(packet.numPackets);
-    }
     if (this.config.debug) {
       if (packet) {
         // eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


sidorares:
Oh wow, many thanks for tracking that down @huycn! Is there any way we can have a repro example for the issue and fix? (I'll try to create it myself but if you already have something please  post it here

huycn:
Sorry for the delay, I was taken away by a tough week.

Here is a working example to reproduce the issue:
```
const mysql = require("mysql2/promise");

async function main() {
    var pool = mysql.createPool({
        host: "localhost",
        port: 3306,
        user: "mysql_user",
        password: "mysql_password",
        database: "mysql_schema",
    });
    pool.execute("SELECT version()");

    await new Promise(resolve => setTimeout(resolve, 10000));
}
main();
```
On MySQL server 8.0.24 or later, set `wait_timeout` and `interactive_timeout` to 5s:
```
[mysqld]
...
wait_timeout = 5
interactive_timeout = 5
```
and then restart the server before running the code above.

trasherdk:
I'm a bit confused here. I don't understand the problem.

If you want a async API `const mysql = require("mysql2/promise")`
But don't want to wait for requests to resolve.

The `createPool()` seems to be synchronous.
```js
    host: DB_HOST,
    user: DB_USER,
    password: DB_PASS,
    database: DB_NAME
  });
```
as it's ready for the `pool.execute("SELECT version()")`

My test code:
```js
import dotenv from 'dotenv'

dotenv.config()
const { DB_HOST, DB_USER, DB_PASS, DB_NAME } = process.env

async function main () {
    host: DB_HOST,
    user: DB_USER,
    password: DB_PASS,
    database: DB_NAME
  });

  console.log("SELECT version()")

  let version = pool.execute("SELECT version()")

  console.log('version:', version)
  /*
  console.log('wait for promised timeout')
  await new Promise(resolve => setTimeout(() => {
    console.log('resolve timeout')
    resolve
  }, 2000));
  console.log('version:', version)

  * Anything after await new Promise() never executes.
   */
  version = await version;
  console.log('awaited version:', version[0])

  const version2 = await pool.execute("SELECT version()");
  console.log('awaited execute():', version2[0])

}

main();
```
And the output:
```js
[nodemon] starting `node mysql/src/test-1660.js`
SELECT version()
version: Promise { <pending> }
awaited version: [ { 'version()': '5.7.39' } ]
awaited execute(): [ { 'version()': '5.7.39' } ]
[nodemon] clean exit - waiting for changes before restart
```

GingerAdonis:
Any update on this PR? Would be great to have this fix merged.

sidorares:
Finally encountered this issue in my setup ( PlanetScale connection ), and your change fixed it, thanks [huycn](https://github.com/huycn)!

Issue #1661, opened 2022-10-27
JSON Data Type Parsing in typeCast Compatibility Wrapper


jyoonPro:
This issue is similar to #409, in that the JSON field will not natively parse in `utf-8`.

Implementing a custom typeCast for JSON will result in incorrect parsing, so adding an option to override the default encoding seems suffice.

Current code:
```
function wrap(field, type, packet, encoding) {
  return {
    type: type,
    length: field.columnLength,
    db: field.schema,
    table: field.table,
    name: field.name,
    string: function() {
      return packet.readLengthCodedString();
    },
    buffer: function() {
      return packet.readLengthCodedBuffer();
    },
    geometry: function() {
      return packet.parseGeometryValue();
    }
  };
}
```

An example of custom typeCasting:

```
function typeCastFunction(field, next) {
  switch (field.type) {
    case 'JSON': {
      return JSON.parse(field.string());
    }

    default: {
      return next();
    }
  }
}
```

Proposal: add an optional parameter to override the encoding directly inside the `field.string()` method. For example, `field.string("utf8")`.

jyoonPro:
Will create a PR and link to this issue.

PR #1662, opened 2022-10-27
fix: typeCast string() parameters for JSON

jyoonPro:
#1661

### DIFF
--- a/lib/parsers/text_parser.js
+++ b/lib/parsers/text_parser.js
@@ -87,8 +87,8 @@ function compile(fields, options, config) {
       name: field.name,
       string: function(encoding = field.encoding) {
         if (field.columnType === Types.JSON && encoding === field.encoding) {
-          // JSON columns are always returned as BINARY, so we have to
-          // handle them according to JSON specs and use "utf8",
+          // Since for JSON columns mysql always returns charset 63 (BINARY),
+          // we have to handle it according to JSON specs and use "utf8",
           // see https://github.com/sidorares/node-mysql2/issues/1661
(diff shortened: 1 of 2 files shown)


sidorares:
would you be able to add a simple unit test, ideally in the context of json parsing ( e.i: create temp table with json column, insert data, read it using typeCast ) The test need to fail without proposed change to `lib/parsers/text_parser.js`

The change itself looks good to me. ( I wonder if we should be more helpful proactively? A warning in typeCast if the `string()` is used without charset AND field type is JSON )

Also could you try to follow https://www.conventionalcommits.org/en/v1.0.0/ in commit messages @jyoonPro?

jyoonPro:
@sidorares How should we warn the user? I'm not sure if I should use console.log() since no other part of the code seems to be using it.

sidorares:
> @sidorares How should we warn the user? I'm not sure if I should use console.log() since no other part of the code seems to be using it.

This is how we warn when the user `await`s non-promise query ( also this is immediately followed by `throw` so maybe not the best example )
https://github.com/sidorares/node-mysql2/blob/241cb10658aa32b410d4f4a3b618156e8bec6ff5/lib/commands/query.js#L42

lib/parsers/text_parser.js
```diff
@@ -85,8 +85,15 @@ function compile(fields, options, config) {
       db: field.schema,
       table: field.table,
       name: field.name,
-      string: function() {
-        return _this.packet.readLengthCodedString(field.encoding);
+      string: function(encoding = field.encoding) {
+        if (field.columnType === Types.JSON && encoding === field.encoding) {
```
Marsup, line 89: Shouldn't you also check that encoding is not already utf8? I have a case where `field.encoding` is, and passing utf8 explicitly still raises the warning.

PR #1666, opened 2022-11-05
fix(connection): ensure pooled connections get released

### LINKED ISSUE #1664: TypeOrm: This socket has been ended by the other party
I've seen there a lot of open issues considering this, i may have stumbled upon something interesting while investigating the very same error.

### Our context

We are using an AWS serverless architecture where our mysql server is behind an [RDS Proxy](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy-managing.html#rds-proxy-connection-pooling-tuning.idleclienttimeout).

First i was thinking we were missing the configuration to keep our connections in the pool alive, see this [issue](https://github.com/sidorares/node-mysql2/pull/1081) for more info. 

Although that seems functional it did not have any effect on our particular situation.

We have our idleclientTimeout for the RDS proxy configured around 120 secs, so the issue is really a big issue with us. Although we could up the timeout to something more reasonable we would keep finding ourselves in the same situation after lets say 30 minutes.

Because of the specific sleepstate of lamda's on aws, the keepalive makes no difference, because no user codes executes anymore after running the function.

I've seen we did some handling in [connection.js](https://github.com/sidorares/node-mysql2/blob/1eef0a71084da297a1db627c277b4d94609591b2/lib/connection.js#L99) of the connection's `close` event.

### Findings

The socket has been ended happens when the connection is used by a query, so i changed something in the connection.js class and added a listener on the `end` event

```
    this.stream.on('end', () => {
      console.warn('connection recycle');
    });

    // below is existing code ;)
    this.stream.on('close', () => {
      // we need to set this flag everywhere where we want connection to close
      if (this._closing) {
        return;
      }
      if (!this._protocolError) {
        // no particular error message before disconnect
        this._protocolError = new Error('Connection lost: The server closed the connection.');
        this._protocolError.fatal = true;
        this._protocolError.code = 'PROTOCOL_CONNECTION_LOST';
      }
      this._notifyError(this._protocolError);
    });

```

With that change, i could see that when starting the function the log contained a 'connection recycle' item, just before the function fired. And off course failed with the 'socket has been ended shizzle'

From the time is changed that function to contain the same logic as the `on('close`` handler the problem of the socket has been ended disappears and everything works is it should.

```
    this.stream.on('end', () => {
      console.warn('connection recycle');
      // we need to set this flag everywhere where we want connection to close
      if (this._closing) {
        return;
      }
      if (!this._protocolError) {
        // no particular error message before disconnect
        this._protocolError = new Error('Connection lost: The server closed the connection.');
        this._protocolError.fatal = true;
        this._protocolError.code = 'PROTOCOL_CONNECTION_LOST';
      }
      this._notifyError(this._protocolError);
    });

```

Please advise, i don't mind creating a pull for this if you think this is useful :)

i don't think this is the actual code change that is needed but i just seems to solve the issue, it has something to do with the setting of the `_fatalError`. 

### Update

when i compare the handling with the `mysql` package i see them handling the `end` event too, so i guess we need to do something there

tommarien:
The PoolConnection is listening for the end event to be emitted to close the connection, but the connection is not listening to the underlying end event emitted by the net.Socket.

fixes #1664

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -96,6 +96,10 @@ class Connection extends EventEmitter {
       }
       this.packetParser.execute(data);
     });
+    this.stream.on('end', () => {
+      // emit the end event so that the pooled connection can close the connection
+      this.emit('end');
+    });
     this.stream.on('close', () => {
       // we need to set this flag everywhere where we want connection to close
       if (this._closing) {
--- a/test/integration/test-pool-end.js
+++ b/test/integration/test-pool-end.js
@@ -1,0 +1,22 @@
+'use strict';
+
+const { createPool } = require('../common.js');
+const assert = require('assert');
+
+const pool = createPool();
+
+pool.getConnection((err, conn) => {
+  assert.ifError(err);
+
+  assert(pool._allConnections.length === 1);
+  assert(pool._freeConnections.length === 0);
+
+  // emit the end event, so the connection gets removed from the pool
+  conn.stream.emit('end');
+
+  assert(pool._allConnections.length === 0);
+  assert(pool._freeConnections.length === 0);
+
+  // As the connection has not really ended we need to do this ourselves
+  conn.destroy();
+});


tommarien:
@sidorares anything i can help on the github.GithubException.GithubException: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue-comment"} ?

tommarien:
@sidorares is it because coverage regressed by it, i can try to put up an integration test that verifies the behavior ?

sidorares:
I believe coverage fails because of the autogenerated access token permissions preventing it from posting coverage report when initiator is external repo, so its usually a false positive. I should really try to fix that or disable workflow. Please ignore it for now ( But if you can integration test would be really really great )

tommarien:
Will try later tonight to add an integration test for it thanks for the update @sidorares

tommarien:
@sidorares Thanks for merging, are you going for another rc or is the 3.0.0 planned in the coming weeks ?

sidorares:
Yes, I'll push rc hopefully in the next couple of days. Unfortunately release-please not ready yet, otherwise I'd do releases more frequently

tommarien:
@sidorares Ok thanks 🙏

tommarien:
Hi @sidorares, any chance you could release a new rc beginning of januari, we want to test drive it in our enviroments ?

tommarien:
@sidorares Do you have any update on this? Sorry to bother you again with this, but the issue is really blocking us, and we need to move forward, but i would like to prevent forking etc ;)

Issue #1668, opened 2022-11-06
node-mysql2 does not work with Node.JS 18.12.1 LTS


ericvgit:
First, thank you a lot for developing mysql2.
I had various node-mysql2 code blocks to connect to MySQL 8 on Ubuntu 20.04 LTS, which worked with Node.js 16 (LTS). 
None of those code blocks work with Node.js 18 LTS. I isolated the problem to the simple fact that it could not connect to MySQL. I switched to Node.js 16 (LTS), and everything worked again. 
This is very concerning. After investing a lot in moving from PHP to Node.js, now, I have no reliable means for connecting to MySQL when I use an LTS version of Node.js. And this occurred after being forced to move to mysql2 from mysql after installing a newer version of MySQL (version 8). 
It is also concerning because I don't see this problem being reported so far (or not abundantly, so I could find it by Googling or in StackOverflow. And this may be the result of a lack of interest in working with MySQL by the Node.js community and the bias toward using relational databases (e.g., MongoDB).
I even considered Rust to write my connection module, but the fun ends when you have to import a MySQL module similar to mysql2. Eventually, when I considered pure C, I found the same, plus the exciting part: you have to free a memory allocated by the module (but where). 
Bottom line mysql2 has become as essential as express.js, I assume, for those who depend on MySQL. 

Best regards,
Eric V

sidorares:
Hey @ericvgit I added node 18 to the build matrix and tests seem to be running fine
Can you elaborate a bit more what is failing in your setup ( that works with node v16 )? Any errors or other log messages or example code you can share?

ericvgit:
Hello @sidorares
Thank you for developing mysql2 and looking into this problem:
I have two methods to connect to MySQL via mysql2, which both work fine with Node 16 and not with Node 18.12.1 LTS. Also, I use a real web server NGINX/Ubuntu 20.04 LTS and MySQL 8

### **Method 1:**

**_Module mysqlDB.js_**

```
const mysql = require('mysql2');

function DBconnect() {
    const db = mysql.createConnection({
        host: 'localhost',
        user: 'xxx',
        password: 'xxx',
        database: 'xxx'
    })

    // Connect

    db.connect((err) => {
        if(err) {
            //throw err
            console.log("MySQL Error in DBconnect()")
        } else {
            console.log('MySql Connected ...')
        }        
    })
    return db
}

module.exports = DBconnect
```

**_mysql_test1.js_**

```
[...]

PR #1670, opened 2022-11-07
Add node 18 to the matrix

sidorares:
Investigating issue in #1668 ( also node 18 been around for more than 6 months now and in LTS )


sidorares:
I'll leave this to dependabot

PR #1671, opened 2022-11-07
Add node 18 to the build matrix

sidorares:
Investigating issue in https://github.com/sidorares/node-mysql2/issues/1668 ( also node 18 been around for more than 6 months now and in LTS )


sidorares:
I'll repurpose this PR for "node 18 matrix"

PR #1549, opened 2022-03-30
added tests and fixes for #1546

stevemandl:

Issue #1546, opened 2022-03-29
column type support


stevemandl:
```js
import mysql from "mysql";
const dbConfig = {
    connectionLimit: 10,
	host: "db",
	user: "db-read",
	password: "Secr3t",
	database: "test",
};
const pool = mysql.createPool(dbConfig);
const cbk = (err, rows, fields) => {
	if (err) console.log(err);
	if (rows.length) {
		var data = {};
		data.meta = fields.map((f)=>{
			return {"text": f.name,"type": mysql.Types[f.type]};
		});
		data.rows = rows.map((row)=>{
			return data.meta.map((col)=>{
				return row[col.text];
			});
		});
		console.log( data);
	}
	else console.log("No data");
	process.exit(0);
};
const sql = \`select 'ABCDEF', 123, 123.45, cast('2022-01/01 11:59' as datetime) x, cast('{"name": "Steve"}'  as json) y\`;
console.log('running query...');
pool.query(sql, cbk);
```

When I run the program above, I get the following output:

```js

{
  meta: [
    { text: 'ABCDEF', type: 'VAR_STRING' },
    { text: '123', type: 'LONGLONG' },
    { text: '123.45', type: 'NEWDECIMAL' },
    { text: 'x', type: 'DATETIME' },
    { text: 'y', type: 'JSON' }
  ],
  rows: [
    [
      'ABCDEF',
      123,
      123.45,
      2022-01-01T16:59:00.000Z,
      '{"name": "Steve"}'
    ]
  ]
}

```

If I change the import to "mysql2" and run the program, I get this output:

```js
{
  meta: [
    { text: 'ABCDEF', type: undefined },
    { text: '123', type: undefined },
    { text: '123.45', type: undefined },
    { text: 'x', type: undefined },
    { text: 'y', type: undefined }
  ],
  rows: [ [ 'ABCDEF', 123, '123.45', 2022-01-01T16:59:00.000Z, [Object] ] ]
}
```

The mysql.Types is not supported as it was in mysql, and the field.type also seems to be missing, or replaced by a columnType attribute. For backward compatibility, could you add a field.type alias and implement a Types dictionary similar to this:

```js
const Types = {
	0x00: 'DECIMAL', // aka DECIMAL 
	0x01: 'TINY', // aka TINYINT, 1 byte
	0x02: 'SHORT', // aka SMALLINT, 2 bytes
	0x03: 'LONG', // aka INT, 4 bytes
[...]

PR #1672, opened 2022-11-07
Fix wrong field type definition.

cweijan:
Hi, in the callback function of query, field does not have a type field, only columnType and columnLength.

image

### DIFF
--- a/typings/mysql/lib/protocol/packets/FieldPacket.d.ts
+++ b/typings/mysql/lib/protocol/packets/FieldPacket.d.ts
@@ -15,7 +15,8 @@ declare interface FieldPacket {
     orgTable: string;
     protocol41: boolean;
     table: string;
-    type: number;
+    columnType: number;
+    columnLength: number;
     zerofill: boolean;
 }
 


sidorares:
I believe that was unintentional incompatibility between mysqljs/mysql and mysql2 ( `type` in one and `columnType` fields in another ), and later we added `type` as an alias to `columnType` for compatibility

https://github.com/sidorares/node-mysql2/blob/198f3710be97e414fc2f067009f9279f91b3c1ba/lib/packets/column_definition.js#L54-L55

- #1546 
- #1549

cweijan:
But the strange thing is that the type is empty, please check the screenshot of my debug.
image

cweijan:
I used version 3.0.0-rc.1, it is not empty now, no new version has been released for a year after 2.3.3.

cweijan:
Are you considering adding columnType to the type definition?

sidorares:
> Are you considering adding columnType to the type definition?

I'd probably deprecate it, not sure what would be the best approach from typings point of view

Issue #1669, opened 2022-11-06
Is there no way to create a connection pool using a database URI?


cmmartin:
I'm using planetscale which injects a `DATABASE_URL` env variable. `createConnection` accepts a connection string, but `createPool` does not. Is there any way to create a connection pool with a database url?

rafipiccolo:
What makes you think it doesnt work ?

they both work :

```
const connectionstring = 'mysql://xxxxxx:xxxxx@xxxxxx.fr/xxxx?charset=utf8mb4&connectionLimit=10'
const pool = mysql.createPool(connectionstring);
const connection = mysql.createConnection(connectionstring);
```

sidorares:
PoolConfig does accept string argument:

https://github.com/sidorares/node-mysql2/blob/21b13a0b25086b97a63cac2ca65b5cb6e37bbe74/index.js#L22

https://github.com/sidorares/node-mysql2/blob/21b13a0b25086b97a63cac2ca65b5cb6e37bbe74/lib/pool_config.js#L8

cmmartin:
@rafipiccolo @sidorares You're right it does work. But according to the typescript types in the promise package, `createConnection` can take a string or config, but `createPool` can only take a config...

https://github.com/sidorares/node-mysql2/blob/master/promise.d.ts#L144-L148

```tyepscript
export function createConnection(connectionUri: string): Promise<Connection>;
export function createConnection(
  config: ConnectionOptions
): Promise<Connection>;
export function createPool(config: PoolOptions): Pool;
```

PR #1675, opened 2022-11-07
Add TS type for createPool to accept a string

cmmartin:
Related: https://github.com/sidorares/node-mysql2/issues/1669


sidorares:
Thanks @cmmartin !

PR #1697, opened 2022-11-12
chore(build): don't fail if coverage action unable to post a comment

sidorares:


sidorares:
not exactly a fix for #1585 but better that false positive

Issue #1706, opened 2022-11-21
Adding list of undefined parameter names to the 'Bind parameters must not contain undefined' error message


theirongiant:
Currently the error message shown is:
`Error: Bind parameters must not contain undefined. To pass SQL NULL specify JS null`

I've added some code to list which named parameters are undefined so the error message becomes:
`Error: Bind parameters must not contain undefined (:id, :email). To pass SQL NULL specify JS null`

Contributing guidelines says it's best to start an issue on changes before creating a PR so thought I'd mention here to see if it's of interest.

sidorares:
What do you want to change here?

Mysql does not have `null` type, and for that reason driver wants it to be explicit. If you want to send "unset" variables pass null instead, e.i `conn.execute(sql, [param1, param2, id ?? null, email ?? null]`)

theirongiant:
I'm not changing how it works just adding more detail to the error message so that it'll list which of the named parameters in the query were undefined. Ideally in a live system you shouldn't really be seeing this error but if it does occur it's probably helpful to know which parameter was unexpectedly undefined.

sidorares:
yes happy to see anything that would help users fix the problem
I assume you can only know name of the parameter when named placeholders are used? ( in case of array parameters we can still add something helpful, at least parameter index ). I'll see what metadata can be extracted from prepared statement info

theirongiant:
Yeah the code I've written only works in the case of named parameters but happy to take a look to see if I can do something similar for array parameters as well.

sidorares:
something similar to this message:

<img width="702" alt="image" src="https://user-images.githubusercontent.com/173025/203222180-8038678b-2054-4324-bd1c-57a7cbc87f26.png">

I remember some time ago that `reading 'test'` was not in the error message

theirongiant:
Yeah that's the idea, got it working for array parameters now as well. I'll run some tests to make sure that there are no bugs and can create a PR for it.

PR #1707, opened 2022-11-22
More verbose undefined parameter error messages

theirongiant:
Adds list of undefined named parameters/indexes to error message. 

For named parameters the error message will be
`Bind parameters must not contain undefined (parameters :id, :email). To pass SQL NULL specify JS null`
and for array parameters it will be
`Bind parameters must not contain undefined (indexes 0, 1). To pass SQL NULL specify JS null`

Also added cross-env to devDependencies to allow doing
`cross-env FILTER='test-timestamp' npm run test` 
on windows machines.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -550,6 +550,8 @@ class Connection extends EventEmitter {
       if (convertNamedPlaceholders === null) {
         convertNamedPlaceholders = require('named-placeholders')();
       }
+      // Store list of named parameters in the order they appear in the sql so we can reference against value array if needed
+      options.names = options.sql.match(/:\w*/g);
       unnamed = convertNamedPlaceholders(options.sql, options.values);
       options.sql = unnamed[0];
       options.values = unnamed[1];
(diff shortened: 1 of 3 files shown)


sidorares:
ref #1706

sidorares:
I can't remember if `.execute()` support non-array single parameter, with `.query()` you allowed to do this:

```js
var post  = {id: 1, title: 'Hello MySQL'};
// you can do post as well as [post]
var query = connection.query('INSERT INTO posts SET ?', post, function (error, results, fields) {
  if (error) throw error;
  // Neat!
});
```

theirongiant:
I wasn't aware of that, I'll try out some tests and see how I can handle it.

theirongiant:
Doesn't look like execute allows that, I get this error when I attempt it: Bind parameters must be array if namedPlaceholders parameter is not enabled.

Interestingly if I do
`
const connect = createConnection({namedPlaceholders: true});
try {
  connect.execute({sql: 'insert into tig_test_table set :user'}, {user: {id:1, name:'test', email:'test@test.com'}}, (err, results, fields) => {
    console.log(err, results, fields);
  });
} catch (e) {
  console.log(e);
}`

There's no error and nothing is inserted into the db but that's probably beyond the scope of what I'm doing here.

sidorares:
one of the unit tests never finishes with your changes, could you try running tests locally?

theirongiant:
Apologies for that, I've put a fix in for that issue and it no longer hangs on test-execute-bind-functions. I'm not able to get a clean run on a full test run but I think that's down to me running a real old version of mysql (5.6.45). I'll see if I can get a more up to date version if there are any further issues.

PR #1724, opened 2022-12-04
Add PromisePoolCluster typings to promise.d.ts

nebkat:
Tested locally on 2.3.3, untested as submitted.

Want to get confirmation this is accidentally missing - I see PoolCluster typings are also missing from index.d.ts?

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -6,7 +6,8 @@ import {
   QueryOptions,
   ConnectionOptions,
   PoolOptions,
-  Pool as CorePool
+  Pool as CorePool,
+  PoolCluster as CorePoolCluster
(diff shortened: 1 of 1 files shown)


sidorares:
@nebkat I currently don't maintain typings and type coverage is just started to gradually improve by contributors like yourself. Could you please add simple typing smoke test ( fails with type error without your change and passes with )? Also can you make sure commit message is following [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) so that release-please can format release post and changelog automatically

nebkat:
Fixed by 7f38496097fa6d9cfbced604fe0ddc392b1b1979

PR #1752, opened 2022-12-19
Fix backpressure when using TLS

vobarian:
Rebase of PR #1559 which GitHub automatically closed :(

This PR fixes two related issues when using TLS:

Using LOAD DATA LOCAL, backpressure is not respected, causing out of memory errors if the data stream supplied by infileStreamFactory is faster than the upload to the database server
Using connection.query().stream(), backpressure is not respected if data is returned by the server faster than the client is handling it
Both issues result from the fact that the backpressure mechanisms of pause(), resume(), emit('pause'), on('drain'), etc., were being applied to the original stream object, which have no effect after that stream has been wrapped inside a TLSSocket, according to my experiments. This can lead to an out of memory error if the buffer grows too large.

See: backpressure not handled https://github.com/sidorares/node-mysql2/issues/1134
Prior fix, which did not cover TLS, because the startTLS function was replacing the write method: handle backpressure when loading data from file https://github.com/sidorares/node-mysql2/pull/1167

Note: There is still the legacy TLS support at the bottom of connection.js which I did not touch. The docs state TLSSocket was added in Node.js v0.11.4 which was released in 2013. It is so old I did not feel like looking into it.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -382,9 +382,7 @@ class Connection extends EventEmitter {
     secureSocket.on('data', data => {
       this.packetParser.execute(data);
     });
-    this.write = buffer => {
-      secureSocket.write(buffer);
-    };
+    this.stream = secureSocket;
     // start TLS communications
     secureSocket._start();
   }


sidorares:
I'm happy to delete anything not required to support node v14+

vobarian:
> I'm happy to delete anything not required to support node v14+

Sorry, my comment may have been confusing. I was just pointing out that the fix I created in this PR will only fix the problem for Node.js versions >= 0.11.4. Since you have stated you're not maintaining for those very old versions, this is fine and the fix can be merged.

It's not related to this PR, but if you want to clean up code that was for old Node versions, the if/else block at the bottom of connection.js can be deleted:

```
if (Tls.TLSSocket) {
  // not supported
} else {
  ...
```

The Node.js docs state that Tls.TLSSocket was added in 0.11.4, therefore the `else` never executes.
https://nodejs.org/dist/latest-v16.x/docs/api/tls.html#class-tlstlssocket

tolgap:
Just to get some attention back to this 🙂 

I am running into this issue. Right now I'm applying the git patch of this PR on each deployment for `mysql2`. Not ideal but it solves any OOM issues for me when using `inlineStreamFactory` with TLS.

sidorares:
@tolgap sorry it didn't get enough attention. Unfortunately needs another rebase after the changes I added for bun compatibility. Would you be able to branch off `vobarian:fix/tls-stream-backpressure`, resolve conflicts, give it some testing and create fresh PR?

vobarian:
Hello! I just revisited this, rebased, and added a couple tests. When using TLS, prior to the fix, these tests will show the problem by failing, and they pass with the fix in connection.js. I'm not entirely sure how the testing infrastructure works for this project, but I am assuming the tests will run with and without the MYSQL_USE_TLS=1 environment variable; they pass in both cases.

vobarian:
Multiple CI jobs were failing due to the test not passing with a compressed connection. Backpressure for "load data infile" is not working with compression. This PR fixes backpressure with TLS but does not address the compression issue.

Also, I may be misunderstanding the CI jobs, but I note there are multiple jobs with "Compression=0" in the name, but if you look at the stack trace for the errors (ERR_STREAM_WRITE_AFTER_END) it is clearly going through the Deflate functions in compressed_protocol.js. I don't know the details of how the tests and CI jobs work for this project but it sounds like compression is being enabled when it should not be.

I briefly looked into the compression backpressure issue and it is not easy to fix. The basic problem is the length of the deflate queue is unbounded, so it can easily exhaust available memory and crash the process. Preventing that is complicated because it requires a mechanism to stop the writer and resume it when appropriate, but the way the compression code is written, it does not participate in the streams' queuing.

test/integration/connection/test-backpressure-load-data-infile.test.cjs
```diff
@@ -0,0 +1,109 @@
...
+const { assert, test, log } = require('poku');
+const common = require('../../common.test.cjs');
+const { Readable, Duplex } = require('stream');
+const Net = require('node:net');
+const driver = require('../../../index.js');
+const { setTimeout } = require('node:timers/promises');
+
+if (common.config.compress) {
+  console.log('skipping test with compression; load data infile backpressure is not working with compression');
+  process.exit(0);
```
wellwelwel, line 12: You cal also use the `skip` method from **poku**, for example:

```ts
import { skip } from 'poku';

if (something) skip('Skipping for some reason');
```

wellwelwel (changes requested):

wellwelwel (approved):

test/integration/connection/test-backpressure-load-data-infile.test.cjs
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+const { assert, log, skip, sleep, test } = require('poku');
+const common = require('../../common.test.cjs');
+const { Readable, Duplex } = require('stream');
```
wellwelwel, line 5:
```suggestion
const { Readable, Duplex } = require('node:stream');
```

test/integration/connection/test-backpressure-load-data-infile.test.cjs
```diff
@@ -0,0 +1,107 @@
...
+    `,
+      infileStreamFactory: () => bigInput,
+    },
+    (err, result) => {
+      if (err) throw err;
+      log('Load complete', result);
+    }
+  );
+
+  await sleep(100); // allow time for backpressure to take effect
```
wellwelwel, line 98:
```suggestion
  await sleep(1000); // allow time for backpressure to take effect
```

Testing locally, I only experienced errors after ~500ms. This increase in time is just a suggestion 🙋🏻‍♂️

wellwelwel:
Thanks, @vobarian! **LGTM!**
I left a comment with a suggestion. Feel free to accept it or not (let me know so I can merge this _PR_) 🤝

vobarian:
Thank you!

Issue #1711, opened 2022-11-24
move mocha and chai to devDependencies, update lru-cache in named-placeholders to allow deduplication


sidorares:

falkenhawk:
I've also noticed it while testing `3.0.0-rc.1` :)
mysql2 brings in mocha with all its sub-dependencies. btw this was introduced in #1606

MasterOdin:
The PR title isn't quite right I think, where it's `named-placeholders` that needs to be updated with regards to `lru-cache`, not `sqlstring`:

```
$ npm ls lru-cache
mysql2@3.0.0-rc.1 /Users/mpeveler/code/github/node-mysql2
├─┬ @typescript-eslint/eslint-plugin@5.48.0
│ └─┬ semver@7.3.8
│   └── lru-cache@6.0.0
├── lru-cache@7.14.1
└─┬ named-placeholders@1.1.2
  └── lru-cache@4.1.5
```

sidorares:
@MasterOdin you are right, I meant "update lru-cache sub dependency in named-placeholders to be the same version as mysql2 uses (or a range that allows deduplication)"

sidorares:
https://github.com/mysqljs/named-placeholders/pull/19 fixes lru-cache dedupe issue

PR #1774, opened 2023-01-08
Move chai and mocha to devDependencies

MasterOdin:
PR completes half of #1711, by moving the `chai` and `mocha` from `dependencies` to `devDependencies` as they're only used for running tests.

For the `package-lock.json` changes, the removal of the `@typescript-eslint` packages is unrelated to the chai/mocha changes, and I get that when I do a fresh `npm install` after cloning this repo. Otherwise, all the other changes to the `package-lock.json` file is moving all the stuff that chai/mocha rely on to be marked as dev dependencies.


sidorares:
Thanks!

PR #1784, opened 2023-01-12
Add geometry typecast test

sidorares:
this is a fork of #1428, investigating mysql 5.7 failures


wellwelwel:
> investigating mysql 5.7 failures

@sidorares, it's because the `SRID 0` in:

```sql
CREATE TEMPORARY TABLE geom_test (p POINT SRID 0, g GEOMETRY NOT NULL SRID 0);
```

---

By removing the `SRID`, it will work:
```sql
CREATE TEMPORARY TABLE geom_test (p POINT, g GEOMETRY NOT NULL);
```

* When we don't specify an `SRID`, **MySQL** will assume the points to be 0 by default.

---

Screenshot 2024-01-23 at 23 16 49

PR #1802, opened 2023-01-19
Simplify query and execute type definitions

### LINKED ISSUE #1329: Problematic overload definitions for Connection.query()
Current definition of Connection.query() methods, with 4 overloads, makes it difficult to write a wrapper on top of Connection class:
```typescript
type ResultVariants = RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader;
type QueryValues = any | any[] | { [param: string]: any };
type SqlResult<T extends ResultVariants> = [T, FieldPacket[]];

query<T extends ResultVariants>(sql: string, values?: QueryValues): Promise<SqlResult<T>>;
query<T extends ResultVariants>(sql: mysql.QueryOptions, values?: QueryValues): Promise<SqlResult<T>>;
query<T extends ResultVariants>(sql: string | mysql.QueryOptions, values?: QueryValues): Promise<SqlResult<T>> {
  return this.connImpl.query(sql, values);
}
```
TSC gives following error:
>  error TS2769: No overload matches this call.
  Overload 1 of 4, '(sql: string, values: any): Promise<[T, FieldPacket[]]>', gave the following error.
    Argument of type 'string | QueryOptions' is not assignable to parameter of type 'string'.
      Type 'QueryOptions' is not assignable to type 'string'.
  Overload 2 of 4, '(options: QueryOptions, values: any): Promise<[T, FieldPacket[]]>', gave the following error.
    Argument of type 'string | QueryOptions' is not assignable to parameter of type 'QueryOptions'.
      Type 'string' is not assignable to type 'QueryOptions'.
   return this.connImpl.query(sql, values);

Same happens even if the wrapper also defines 4 overloads. 

According to [TypeScript Handbook, the better way](https://www.typescriptlang.org/docs/handbook/2/functions.html#writing-good-overloads) to define overloads with the same amount of arguments is to use union types.

Following definitions fixes the problem:
```typescript
  query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
    sql: string | QueryOptions
  ): Promise<[T, FieldPacket[]]>;
  query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
    sql: string | QueryOptions,
    values: any | any[] | { [param: string]: any }
  ): Promise<[T, FieldPacket[]]>;
```

And since `values` is actually optional argument, just single signature should be enough:
```typescript
  query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
    sql: string | QueryOptions,
    values?: any | any[] | { [param: string]: any }
  ): Promise<[T, FieldPacket[]]>;
```

ArtskydJ:
Fixes #1329.

Some type declarations were using overloading which was causing me problems.  The equivalent types can be declared using [Optional Parameters](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-optional-parameters) and [Union Types](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-union-types) which is the recommended method according to the TypeScript Do's and Don'ts (see links).

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -26,41 +26,15 @@ export interface Connection extends EventEmitter {
   changeUser(options: ConnectionOptions): Promise<void>;
 
   query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
-    sql: string
-  ): Promise<[T, FieldPacket[]]>;
-  query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
-    sql: string,
-    values: any | any[] | { [param: string]: any }
-  ): Promise<[T, FieldPacket[]]>;
-  query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
-    options: QueryOptions
-  ): Promise<[T, FieldPacket[]]>;
-  query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
-    options: QueryOptions,
-    values: any | any[] | { [param: string]: any }
+    sql_or_options: string | QueryOptions,
+    values?: any | any[] | { [param: string]: any }
   ): Promise<[T, FieldPacket[]]>;
 
   execute<
     T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader
   >(
-    sql: string
-  ): Promise<[T, FieldPacket[]]>;
-  execute<
-    T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader
-  >(
-    sql: string,
-    values: any | any[] | { [param: string]: any }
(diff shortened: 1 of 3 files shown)


sidorares:
I wonder if we can add some type check smoke test ( should fail without your changes and pass with )

ArtskydJ:
I added `typings/test/promise.ts`, and was able to replicate the issue I was running into.  The test passes only after my changes to `promise.d.ts`.

sidorares:
@wellwelwel do you think this PR is still relevant? We can probably close it as already fixed with recent changes

wellwelwel:
> @wellwelwel do you think this PR is still relevant? We can probably close it as already fixed with recent changes

@sidorares, I have a plan to close most Issues that are questions about how to use `query` and `execute` methods after creating the **TypeScript** documentation by commenting each of them with the related part of the documentation.

The Issue #1329 is one of them.

About this **PR**, I don't see it as an Issue, but yes, this was indirectly fixed in `^3.4.2`.

Issue #1801, opened 2023-01-18
Support bun and deno


sidorares:
Bun 0.5 added support for `net` and `tls` modules
I did a quick tests and there are few issues connecting to the server when started from bun:

- call to `TLSSocket._start()` - this is undocumented method, need to figure out how to start secure connection after tls socket is constructed from unsecure stream
- `crypto.update('string')` calls - bun wants to see a `Buffer` here, probably worth to convert to a buffer explicitly

Haven't tried deno yet

vlasky:
Bun.js's previously stated that ["Support for Node.js APIs continues to be a top priority for Bun"](https://bun.sh/blog/bun-v0.4.0#:~:text=Support%20for%20Node.js%20APIs%20continues%20to%20be%20a%20top%20priority%20for%20Bun), so perhaps it's preferable that this missing functionality is best incorporated into Bun.js?

sidorares:
Yeah, I'll be reporting functionality that is broken when used in bun and not yet deprecated in node, and refactor undocumented or deprecated apis on the driver side

sidorares:
I don't have time to focus on deno ( or even check what works and what not ), but bun can probably be considered "officially" supported: main functionality, tls, compression work out of the box and unit tests run for bun in CI ( small subset for now )

suchislife801:
Just here to express my complete and total frustration with people who insist in supporting Node JS when Deno exists. God how I hate Node JS.... Is it even possible to dislike something this much?

Deno's driver is made by people who don't think strings are important in queries...
mysql2 port of Node JS is never updated...
mysql connector from Oracle is Node JS...

PR #1805, opened 2023-01-23
Bun support

sidorares:
https://github.com/sidorares/node-mysql2/issues/1801

- [x] add Bun to CI
- [x] fix `caching_sha2_password` errors
- [ ] work around missing `net.createServer` by using `Bun.listen`
- [ ] compressed protocol bun fixes
- [ ] TLS bun fixes

( going to track last 3 items separately )

### DIFF
--- a/.github/workflows/ci-bun.yml
+++ b/.github/workflows/ci-bun.yml
@@ -37,6 +37,10 @@ jobs:
         with:
           bun-version: ${{ matrix.bun-version }}
 
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
       - name: Cache dependencies
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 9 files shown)


Djboy08:
Do I have to wait for this to show up in a release to use?

wellwelwel:
@Djboy08, it was merged in Jan, naturally it has already been released. I suggest you open a Issue indicating this problem and, if possible, share a minimal way to reproduce your problem 🙋🏻‍♂️

Djboy08:
> @Djboy08, it was merged in Jan, naturally it has already been released. I suggest you open a Issue indicating this problem and, if possible, share a minimal way to reproduce your problem 🙋🏻‍♂️

Gotcha, the reason i brought it up was because he recently edited it saying he completed the TLS part 2 weeks ago so I was not sure why it took so long for him to update this. 

Sadly I don't really have the time right now to create a issue but Ill see what I can do when I have time. Because at the end of the day this is literally just how planetscale requires `?ssl={"rejectUnauthorized":true}` in the url and for some reason the mysql2 driver with bun just does not work with it.

sidorares:
@Dieken thanks for the report, I'll have a look. I actually only tested with planetscale, not sure what has changed since

PR #1835, opened 2023-02-08
Updated `lru-cache` reset method to clear

daniellockyer:
refs https://github.com/isaacs/node-lru-cache/blob/main/CHANGELOG.md#v7---2022-02

- the `.reset` method became `.clear` in `lru-cache` v7 but the old method was maintained for backward compatibility and given a deprecation warning
- without this change, changing user in `node-mysql2` gives the
  following warning: `(node:27579) [LRU_CACHE_METHOD_reset] DeprecationWarning: The reset method is deprecated. Please use cache.clear() instead.`


sidorares:
Not a problem, can edit in GitHub web ui sith squash and merge

PR #1899, opened 2023-03-15
Create failing test for https://github.com/sidorares/node-mysql2/issues/1847

### LINKED ISSUE #1847: Connection errors are not propagated to the query stream
Sample code:
```js
import { createPool } from "mysql2";

const connection = createPool({
    host: 'localhost',
    user: 'user',
    password: 'pass',
  });
try {
  for await (const row of connection.query("SELECT * FROM big_table").stream()) {
    // handle row here
  }
} catch (e) {
  // connection errors are not caught here
  console.log('caught error', e);
}
```

If any connection errors occur during the command execution, they will not be caught while iterating over the rows.

After investigation I found the issue to be in `Connection.notifyError` here https://github.com/sidorares/node-mysql2/blob/88587c6b64e71a82ca004170eccda3fb1ce81ed1/lib/connection.js#L224-L230

From what I see, the error will be passed to the command only if the command was executed with a callback.

A solution that seems to fix the issue is to also emit the error to the command like this:
```js
if (this._command) {
  this._command.emit('error', err);
}
```

With the above change the error will be emitted to the command, and will also be thrown when iterating over the stream.

Can be reproduced by running a slow query like `SELECT SLEEP(30)` and then issue a `KILL "connection id"`

dsech:
Fixes #1847

This PR contains just the failing test for #1847, without any fixes yet.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -53,9 +53,7 @@ exports.waitDatabaseReady = function(callback) {
   tryConnect();
 };
 
-exports.createConnection = function(args) {
-
-  const driver = require('../index.js');
+function createConnectionFromDriver(driver, args) {
   if (!args?.port && process.env.MYSQL_CONNECTION_URL) {
     return driver.createConnection({ ...args, uri: process.env.MYSQL_CONNECTION_URL })
   }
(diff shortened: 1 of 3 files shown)


wellwelwel:
Closing in favor of #3775.

Issue #1722, opened 2022-12-02
Typing error in v3.0.0-rc1 Interface 'PoolConnection' cannot simultaneously extend types 'PoolConnection' and 'Connection'


abentpole:
Hello, I was giving v3.0.0-rc1 a spin and noticed the following issue:
```
node_modules/mysql2/index.d.ts:86:18 - error TS2320: Interface 'PoolConnection' cannot simultaneously extend types 'PoolConnection' and 'Connection'.
  Named property 'execute' of types 'PoolConnection' and 'Connection' are not identical.

86 export interface PoolConnection extends mysql.PoolConnection, Connection {
                    ~~~~~~~~~~~~~~
```

This was noticed with Typescript version 4.5.5, though I also just tested with 4.9.3 and confirmed it occurs there too.
**Edit:** A note that for this to come up when the library is in use; ensure `skipLibCheck` isn't `true`.

sidorares:
could you post a snippet of typescript code to reproduce the error @abentpole ?

abentpole:
@sidorares Just a matter of importing the promise version of the library in a Typescript project with `"skipLibCheck": false`.

So to replicate it:
1. `npm init`
2. `npm i mysql2@3.0.0-rc.1 typescript`
3. `npx tsc --init`
4. Open `tsconfig.json`, go to the bottom and set "skipLibCheck" to `false`
5. Make a file, `index.ts`, with the content `import * as mysql from 'mysql2/promise';`
6. `npx tsc` and observe the error

CComparon:
Same problem happening to me. Both with `typescript": "4.7.4"` and `typescript": "4.9.4"`.

HofmannZ:
Same issue with version `3.0.1`.

rohan-qp:
Same issue with typescript version `4.9.4`. For now, the only way I found to get it running was by setting `skipLibCheck: true` in `tsconfig`.

khanfx:
So for a concrete example -

When faced with this error:

```
node_modules/mysql2/index.d.ts:86:18 - error TS2320: Interface 'PoolConnection' cannot simultaneously extend types 'PoolConnection' and 'Connection'.
  Named property 'unprepare' of types 'PoolConnection' and 'Connection' are not identical.

86 export interface PoolConnection extends mysql.PoolConnection, Connection {
                    ~~~~~~~~~~~~~~
```

The items indicated are:

From `mysql2\index.d.ts`:

```ts
[...]

Issue #86, opened 2014-03-27
Why does this module exist?


statico:
The README is missing valuable information on why this module should be used instead of node-mysql, node-libmysqlclient and node-mysql-native.

PR #1940, opened 2023-04-09
fix: PoolConnection redundancy when extending Connection interface in TypeScript

wellwelwel:
As I mentioned in [#1722](https://github.com/sidorares/node-mysql2/issues/1722#issuecomment-1500928005) issue:

Once the [`PoolConnection.d.ts`](https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/PoolConnection.d.ts#L4) has already declared extending the `Connection` interface:

```ts
declare class PoolConnection extends Connection {
    connection: Connection;
    release(): void;
}
```

Then it doesn't needs to re extend the `Connection` in [`index.d.ts` `#86`](https://github.com/sidorares/node-mysql2/blob/master/index.d.ts#L86) like:
```ts
export interface PoolConnection extends mysql.PoolConnection, Connection {
  promise(promiseImpl?: PromiseConstructor): PromisePoolConnection;
}
```

> Causing the error:
> ```
> [...] Interface 'PoolConnection' cannot simultaneously extend types 'PoolConnection' and 'Connection'.
>   Named property 'execute' of types 'PoolConnection' and 'Connection' are not identical.
> 
> 86 export interface PoolConnection extends mysql.PoolConnection, Connection {
>                     ~~~~~~~~~~~~~~
> ```

<hr />

So, I just removed the `Connection` from extends in [`index.d.ts` `#86`](https://github.com/sidorares/node-mysql2/blob/master/index.d.ts#L86) and it will work:
```ts
export interface PoolConnection extends mysql.PoolConnection {
  promise(promiseImpl?: PromiseConstructor): PromisePoolConnection;
}
```

<hr />

As you recommended in [#1722](https://github.com/sidorares/node-mysql2/issues/1722#issuecomment-1433942315):
For this `PR`, I cloned the `master` and tested it again in a real use (locally), it works perfectly.

<hr />

Extra tests:

- ```json
  "skipLibCheck": false,
  ```

- ```json
  "skipLibCheck": true,
  ```

- Also, without `skipLibCheck`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -83,7 +83,7 @@ export interface Connection extends mysql.Connection {
   sequenceId: number;
 }
 
-export interface PoolConnection extends mysql.PoolConnection, Connection {
+export interface PoolConnection extends mysql.PoolConnection {
   promise(promiseImpl?: PromiseConstructor): PromisePoolConnection;
 }
 


wellwelwel:
@sidorares Please, let me know if you have any questions or want to change anything 👨🏻‍💻

sidorares:
@wellwelwel thanks for the PR! I'm away with no internet connection for the next 3 days but I'll try to test and merge as soon as I'm back

sidorares:
Actually I think this can be merged right now, LGTM

PR #1949, opened 2023-04-13
Add typings for Connection.promise().

bluefeet:


bluefeet:
If anyone else runs into this issue before this pull request is merged you can do like I did and create a `src/mysql2.d.ts` with:

```ts
import * as mysql2 from 'mysql2';

declare module 'mysql2' {
  export interface Connection {
    promise: () => Promise<Connection>;
  }
}
```

And then import it in `src/index.ts`:

```ts
import './mysql2.d.ts';
```

newtoncodes:
This fix seems wrong.
https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L155
It does not return a Promise instance?

wellwelwel:
This change causes 3 incompatibilities in [index.d.ts](https://github.com/sidorares/node-mysql2/blob/master/index.d.ts) 😕

```sh
> npx tsc;

node_modules/mysql2/index.d.ts:10:18 - error TS2430: Interface 'import("***/node_modules/mysql2/index").Connection' incorrectly extends interface 'import("***/node_modules/mysql2/typings/mysql/index").Connection'.
  The types returned by 'promise(...)' are incompatible between these types.
    Type 'Connection' is missing the following properties from type 'Promise<Connection>': then, catch, finally, [Symbol.toStringTag]

10 export interface Connection extends mysql.Connection {
                    ~~~~~~~~~~

node_modules/mysql2/index.d.ts:86:18 - error TS2430: Interface 'import("***/node_modules/mysql2/index").PoolConnection' incorrectly extends interface 'import("***/node_modules/mysql2/typings/mysql/index").PoolConnection'.
  The types returned by 'promise(...)' are incompatible between these types.
    Type 'PoolConnection' is missing the following properties from type 'Promise<Connection>': then, catch, finally, [Symbol.toStringTag]

86 export interface PoolConnection extends mysql.PoolConnection {
                    ~~~~~~~~~~~~~~

node_modules/mysql2/index.d.ts:90:18 - error TS2430: Interface 'Pool' incorrectly extends interface 'Connection'.
  The types returned by 'promise(...)' are incompatible between these types.
    Type 'Pool' is missing the following properties from type 'Promise<Connection>': then, catch, finally, [Symbol.toStringTag]

90 export interface Pool extends mysql.Connection {
                    ~~~~


Found 3 errors in the same file, starting at: node_modules/mysql2/index.d.ts:10
```

Issue #1936, opened 2023-04-07
Attribute "promise" does not exist on type "Connection"


tomoat:
Attribute "promise" does not exist on type "Connection"

import mysql from 'mysql2'
import util from 'util'

const conn = mysql.createConnection({
  host: '127.0.0.1',
  user: 'root',
  password: '12345678',
  port: 3306,
  database: 'domain',
  rowsAsArray: true,
})

const db = conn.promise()  // This line reported an error
export { db }

wellwelwel:
You need to import by `mysql2/promise`:

```ts
import { createConnection } from 'mysql2/promise';

const db = await createConnection({
   host: '127.0.0.1',
   user: 'root',
   password: '12345678',
   port: 3306,
   database: 'domain',
   rowsAsArray: true,
});

export { db };
```

gtavos:
Even from `mysql2/promise` the` promise()` option is not available in Connection class.  Is it available in 3.2.0 or lower versions?

bluefeet:
I've run into the same issue - while I could import `mysql2/promise` I would prefer to import `mysql2` and then use `conn.promise()` to upgrade it as needed. It would be nice if the typing for Connection included the promise function so that I don't have to create custom typings or explain to developers why it works and to ignore the type errors.

bluefeet:
I figured out how to get `conn.promise()1 types working. Here's my `src/mysql2.d.ts`:

```
import * as mysql2 from 'mysql2';

declare module 'mysql2' {
  export interface Connection {
    promise: () => Promise<Connection>;
  }
}
```

I also have to use it in my `src/index.ts` to make it take affect:

```
import './mysql2.d.ts';
```

A similar edit to the typings for Connection would resolve this for everyone.

sidorares:
Looks like #1949 should fix the issue

newtoncodes:
Doesn't seem right.
Here is the code: https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L155
It should not return a promise. All my implementations require createConnection and .promise() to directly return a connection instance. Seems like the code is right, but the typings are wrong in the new version as well.

wellwelwel:
@tomoat and @gtavos:

In case `PR` #1955 is merged by @sidorares, I suggest you look at the two examples in the comment of this `PR`.

wellwelwel:
@tomoat, @gtavos, can you try it now in `3.2.2` release?

wellwelwel:
[...]

PR #1955, opened 2023-04-14
fix: ConnectionOptions conflict between mysql and mysql/promise

wellwelwel:
@sidorares, this fix is about the `.promise()` option in issue #1936.

I see that in `3.2.1` release you merged the `PR` #1949, but in this `PR` the `Connection` is changed "globally", however in `mysql/promise` types it's causing 3 conflicts because it don't needs the `promise` option, so I rollbacked it before this fix.

<hr />

### In this fix:

1. Removed all the `PR` #1949 changes (sorry @bluefeet).
2. Removed all the `promise` options from "global" `Connection` Interface
3. Imported and added the `promise` option only in `mysql` ([typings/mysql/index.d.ts](https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/index.d.ts))
   - I don't created a new `promise` method, I just imported the original from:
      ```ts
      import { Pool as PromisePool } from '../../promise';
      ```

<hr />

### Then:
- ✅ In `mysql/promise` import, the option `promise` will not appears
- ✅ In `mysql` import, both the `createConnection` and `createPool` has the `promise` option and show all the options within them
- ✅ As mentioned by @newtoncodes, in `mysql` import, the `.promise()` method returns a `Connection` instance and not a promise

<hr />

### A `mysql` test example:
```ts
// non-promise.ts
import mysql from 'mysql2';

const access: mysql.ConnectionOptions = {
   host: '127.0.0.1',
   user: 'root',
   password: '*',
   port: 3306,
   database: 'mydb',
};

(async () => {
   const db = mysql.createConnection(access).promise();
   
   const [pokemons] = await db.execute('SELECT * FROM pokemons');
   
   await db.end();
   
   console.log(pokemons);
})();
```

<hr />

### A `mysql/promise` test example:
```ts
// promise.ts

const access: mysql.ConnectionOptions = {
   host: '127.0.0.1',
   user: 'root',
   password: '*',
   port: 3306,
   database: 'mydb',
};

(async () => {
   const db = await mysql.createConnection(access);
   
   const [pokemons] = await db.execute('SELECT * FROM pokemons');
   
   await db.end();
   
   console.log(pokemons);
})();
```

<hr />

I cloned the latest `master` after `3.2.1` release and tested it in a real use (locally), it works and build the **TypeScript** perfectly in both `mysql` and `mysql/promise`.

<hr />

### Extra tests:

- ```json
  "skipLibCheck": false,
  ```

- ```json
  "skipLibCheck": true,
  ```

- Also, without `skipLibCheck`.

<hr />

I hope this helps 🙋🏻‍♂️

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,5 @@
 import {
   Connection as PromiseConnection,
-  Pool as PromisePool,
   PoolConnection as PromisePoolConnection,
 } from './promise';
 
@@ -71,7 +70,6 @@ export interface Connection extends mysql.Connection {
     ) => any
   ): mysql.Query;
   ping(callback?: (err: mysql.QueryError | null) => any): void;
-  promise(promiseImpl?: PromiseConstructor): PromiseConnection;
   unprepare(sql: string): mysql.PrepareStatementInfo;
   prepare(sql: string, callback?: (err: mysql.QueryError | null, statement: mysql.PrepareStatementInfo) => any): mysql.Prepare;
   serverHandshake(args: any): any;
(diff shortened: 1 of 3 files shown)


wellwelwel:
@sidorares If you don't mind, I'd like to add a `CI` workflow to test the TypeScript build like an external module after this issue (#1936) is solved.

sidorares:
> @sidorares If you don't mind, I'd like to add a `CI` workflow to test the **TypeScript** build like an external module without code (just build) after this issue (#1936) is solved. It would prevent any incompatibility between the types before future `PRs` are approved.

Go for it, I'm very hesitant to merge TS changes exactly because of lack of typing CI

Issue #1924, opened 2023-03-29
Many error symbols not included


njbotkin:
We noticed that a 3140: [ER_INVALID_JSON_TEXT](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_invalid_json_text) isn't included in https://github.com/sidorares/node-mysql2/blob/master/lib/constants/errors.js.  Then we noticed a lot more error symbols aren't included.  Is there a reason for this?  Would you like a PR that includes the rest of the codes and symbols in https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html?

sidorares:
> Would you like a PR that includes the rest of the codes and symbols in https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html?

Yes please. No specific reason, just not in sync with newer error codes added recentlyю First commit to errors.js is from 2013.

elimelt:
Is anyone currently working on this issue? I'd be happy to help!

sidorares:
@elimelt nothing in progress afaik, happy to review your PR

elimelt:
I've thus far been able to update generate-error-constants.js (found in node mysql module) to work with the source code for MySQL 8.0, but a lot of the codes have either been marked as OBSOLETE, or have been changed entirely. It seems like it was already the approach to just remove the OBSOLETE tags on the error codes, so I went ahead and did that, but I am a little more reluctant to push changes that overwrite/change codes entirely. 

Although, since errors.js basically just serves as a lookup table, I doubt this will introduce any compatability issues accross the codebase. For now I will just include the most up-to-date error codes from the MySQL 8.0 source, and if the maintainers want to go a different direction then I'm happy to find an alternative solution upon review of my PR. Perhaps this would be maintaining a different verion of error.js depending on the version of MySQL one is using?

elimelt:
@sidorares PR has been submitted!

wellwelwel:
@njbotkin, I think it was fixed in version `^3.3.0`:
- #1990.

Can you check it again?

PR #1990, opened 2023-05-03
Added updated/new error codes gathered from MySQL 8.0 source code

elimelt:
Per [issue #1924](https://github.com/sidorares/node-mysql2/issues/1924), I added the missing error code for invalid JSON text, as well as many others that were added in MySQL 8.0

Additionally, I wrote a test (test/integration/connection/test-insert-json.js) to verify that the correct error code is returned when error no. 3140 is thrown by MySQL, and also ensured that all existing tests were left unchanged and passing with my changes.

As described in the above issue, there are a few discrepancies between the existing error codes in node-mysql2 and the newest versions of MySQL, and also many errors whcih have been marked as 'OBSOLETE'. To follow what seemed to be the current convention, I went ahead and removed the OBSOLETE prefix from all of these errors, but I'm happy to undo it if needed!


sidorares:
thanks @elimelt !

Issue #1885, opened 2023-03-10
TypeError: LRU is not a constructor


lamiaagebriel:
This error appears each time I install mysql2 using (yarn add mysql2 --save) with Next.js 13.2.
I reinstalled Next.js and mysql2 without any other dependencies to ensure it wasn't another dependency that affects it.

(
  Server Error
  TypeError: LRU is not a constructor
  
  This error happened while generating the page. Any console logs will be displayed in the terminal window.
  Call Stack
  eval
  node_modules\mysql2\lib\parsers\parser_cache.js (5:20)
  (sc_server)/./node_modules/mysql2/lib/parsers/parser_cache.js
  file:///D:/Websites/next-mysql/.next/server/app/api/employees/route.js (1470:1)
  __webpack_require__
  file:///D:/Websites/next-mysql/.next/server/webpack-runtime.js (33:42)
  require
  node_modules\mysql2\lib\parsers\text_parser.js (7:20)
  (sc_server)/./node_modules/mysql2/lib/parsers/text_parser.js
  file:///D:/Websites/next-mysql/.next/server/app/api/employees/route.js (1492:1)
  __webpack_require__
  file:///D:/Websites/next-mysql/.next/server/webpack-runtime.js (33:42)
  require
  node_modules\mysql2\lib\commands\query.js (10:22)
  (sc_server)/./node_modules/mysql2/lib/commands/query.js
  file:///D:/Websites/next-mysql/.next/server/app/api/employees/route.js (986:1)
  __webpack_require__
  file:///D:/Websites/next-mysql/.next/server/webpack-runtime.js (33:42)
  require
  node_modules\mysql2\lib\commands\index.js (5:14)
  (sc_server)/./node_modules/mysql2/lib/commands/index.js
  file:///D:/Websites/next-mysql/.next/server/app/api/employees/route.js (953:1)
  __webpack_require__
  file:///D:/Websites/next-mysql/.next/server/webpack-runtime.js (33:42)
  require
  node_modules\mysql2\lib\connection.js (29:17)
  (sc_server)/./node_modules/mysql2/lib/connection.js
  file:///D:/Websites/next-mysql/.next/server/app/api/employees/route.js (1041:1)
  __webpack_require__
  file:///D:/Websites/next-mysql/.next/server/webpack-runtime.js (33:42)
  require
  node_modules\mysql2\index.js (5:19)
  (sc_server)/./node_modules/mysql2/index.js
[...]

Issue #1965, opened 2023-04-20
require('mysql2') breaks "npm start"


h3llyeah:
If i add to my project
`const mysql = require('mysql2');`
and launch my nodejs project with `npm start`

i get error

```
/nodejs_project/node_modules/lru-cache/dist/cjs/index.js:359
    #initializeTTLTracking() {
                          ^

SyntaxError: Unexpected token '('
    at wrapSafe (internal/modules/cjs/loader.js:915:16)
    at Module._compile (internal/modules/cjs/loader.js:963:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/helar/projects/nodejs/nodejs_udemy/10_sql_intro/node_modules/lru-cache/dist/cjs/index-cjs.js:5:36)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
```

Setup:
OS: Ubuntu 22.04.1 LTS
Nodejs: v12.22.9
mysql2: mysql2@3.2.3

This problem seems to start from verstion mysql2@3.2.1.
mysql2@3.2.0 is ok.

wellwelwel:
@h3llyeah Look at [#1953](https://github.com/sidorares/node-mysql2/issues/1953)

jerrychong25:
Faced similar error on my end.

In the end, I've upgraded my node version and it works.

zoobibackups:
this work fine on the backend , but at cpanel where i have following  node version 

- 10.24.1 
- 11.15.0
- 12.22.9 
- 14.18.3
- 16.17.0 

and i ahve tried mysql2 veriosn 3.2.0, 3.2.1 3.3.0, 3.4.0 ,  but non of the above is working .

Sergiussupa:
Some basic terminal commands are:
...

curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.37.2/install.sh | bash
source ~/.bashrc

node -v
nvm install v20.11.0
node -v
...

Sergiussupa:
Use terminal:
```

[...]

PR #2004, opened 2023-05-11
fix: LRU constructor

wellwelwel:
@sidorares, this time, the fix is in **JavaScript** 🤹🏻‍♀️

<hr />

### This fixes the LRU Constructor:
 - #1885

The problem is a compatibility issue between CommonJS and ECMAScript Modules when importing the `lru-cache`.

<hr />

To fix it, I changed the import from `LRU` in `./lib/parser_cache.js` and `./lib/connection.js`:

- from
```js
  const LRU = require('lru-cache');
```
- to
```js
  const LRU = require('lru-cache').default;
```

Now, it's compatible with both **CJS** and **ESM**.

<hr />

### NextJS

I tested this fix in a latest **NextJS** project and it works perfectly without use the experimental: `serverComponentsExternalPackages: ["mysql2"]`.

- It may be necessary to delete the `./.next` cache folder before the first `npm run` `build` | `dev`.

<hr />

### Notes

- As I didn't tested in older versions of **Node.js**, the CI tests from this PR will be the "final test"
- ~~In case some CI fail, I will close this PR~~ *(not today)* 🥷🏻✨

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -22,7 +22,7 @@ const EventEmitter = require('events').EventEmitter;
 const Readable = require('stream').Readable;
 const Queue = require('denque');
 const SqlString = require('sqlstring');
-const LRU = require('lru-cache');
+const LRU = require('lru-cache').default;
 
 const PacketParser = require('./packet_parser.js');
 const Packets = require('./packets/index.js');
--- a/lib/parsers/parser_cache.js
+++ b/lib/parsers/parser_cache.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const LRU = require('lru-cache');
+const LRU = require('lru-cache').default;
 
 const parserCache = new LRU({
   max: 15000


sidorares:
great work @wellwelwel 

I wonder how difficult would it be to simulate the original problem. Just loading mysql2 from mjs script works fine without this change, and also I believe bundling with webpack also works. Do you think its possible to add some sort of a test case?

Not directly related, but still mentioned few times in relation to LRU:
- lru-cache version 9 has a breaking change, now instead of default export ( or CJS `const LRU = require('lru-cache')` ) its now a named export ( `import { LRUCache } from 'lru-cache'` / `const { LRUCache } = require('lru-cache')` ). Changing to a named export would probably fix this issue as well
- starting from version 8 it uses syntax incompatible with node 14 ( also changed in the package.json engine ) . I should have made upgrade a major version bump and also indicate in mysql2 engine > 14. There are some users who want to be able to continue using mysql2 in node 14, currently except for LRU nothing else requires dropping v14 support. Few options here: 1) admit v14 os past end of life, those who need it would have to stay on a later versions. Possibly upgrade to lru-cache@9.x and add engine > 14, also release mysql2 as 4.x 2) downgrade to version 7, with the fix from your PR to help with various ESM / CJS edge cases, probably add engine >= 14 and release as minor version

sidorares:
actually its node 12 that is incompatible, due to [this syntax](https://github.com/isaacs/node-lru-cache/blob/df2451defce3bfa2d2cc69309b80979cafceb37b/src/index.ts#L1135) 

from https://node.green

image

also AbortController used to be polyfilled and now its not

https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#class-abortcontroller

wellwelwel:
@sidorares 

Before I decide for `.default`, I tried the recommended from `lru-cache` documentation:

```js
import { LRUCache: LRU } = require('lru-cache');
```

So it worked perfectly in **NextJS** and **ESM**, but not in **CJS**.

I understand that the `require` can resolves the missing `.default`, but the `import` doesn't.

So, I do this:

```js
import { default: LRU } = require('lru-cache');
```

And now, it works in everyone 🤹🏻‍♀️

I decided to use `.default` for compatibility reasons.

<hr />

And about the **Node.js 12**, I agree with [this comment](https://github.com/sidorares/node-mysql2/issues/1965#issuecomment-1521329157) about change the `package.json` compatibility to `>=14`.

<hr />

About the tests, in some currently CI (lake `CI - Linux / Node.js 16.x - DB mysql:8.0.22 - SSL=0 Compression=0`) it perform a connection? If yes, then I believe indirectly there is already a test.

wellwelwel:
In addition, to reproduces the error in **NextJS**, I just created a new app from latest version, imported the `mysql2` in `./src/app/page.tsx` file. So I simulated a connection and executed an query in the same file, inside `Home()`:

```tsx
import mysql2 from 'mysql2/promise';

export default async function Home() {
  const db = await mysql2.createConnection({
    host: '',
    user: '',
    password: '',
    database: ''
  });

  const [rows] = await db.execute('SELECT * FROM `table`');
  console.log(rows);

  return (
    <main></main>
  )
}
```

wellwelwel:
In addition, to reproduces the error in **NextJS**, I just created a new app from the latest version, imported the `mysql2` in `./src/app/page.tsx` file. So I performed a connection and executed a query in the same file, inside `Home()`.

<hr />

> `./src/app/page.tsx`

```tsx
import mysql2 from 'mysql2/promise';

export default async function Home() {
   const db = await mysql2.createConnection({
      host: '',
      user: '',
      password: '',
      database: '',
   });

   const [rows] = await db.execute('SELECT * FROM `table`');
   console.log(rows);

   return <main></main>;
}
```

Then:

```sh
npm run build
```

sidorares:
I'm thinking to merge the PR since it solves existing pain while not adding any other problems

We can solve 2 issues I mentioned ( communicating compatible note versions, potentially as a major semver release; lru-cache@9 upgrade ) later

sidorares:
As for the test - I was hoping to see something that can be done at CI time, not a manual repro steps. Ideally with a small number of dependencies. For example, a local webpack config, a script that creates a bundle ( known to expose the original issue ) and then runs a bundle. Again, we can tackle that later

wellwelwel:
@sidorares, thank you so far!

I will be focusing on my projects for now.

Feel free to reach out if you need anything 🙋🏻‍♂️

sidorares:
Thanks a lot for your help @wellwelwel 🤛 !

Issue #2009, opened 2023-05-13
Unsupport encoding "utf8mb3"


cweijan:
Since [connection_config.getCharsetNumber](https://github.com/sidorares/node-mysql2/blob/master/lib/connection_config.js#L239) cannot obtain the corresponding encoding, the encoding finally passed to iconv-lite is undefined, so the following error is reported.

#bytenode/bytenode/issues/165
```
previous Error: set foreign_key_checks = 0; - Encoding not recognized: 'undefined' (searched as: 'undefined')
    at Object.getCodec (/<proj-dir>/node_modules/mysql2/node_modules/iconv-lite/lib/index.js:104:23)
    at Object.getEncoder (/<proj-dir>/node_modules/mysql2/node_modules/iconv-lite/lib/index.js:115:23)
    at Object.encode (/<proj-dir>/node_modules/mysql2/lib/parsers/string.js:23:25)
    at Query.toPacket (/<proj-dir>/node_modules/mysql2/lib/packets/query.js:16:30)
    at Query.start (/<proj-dir>/node_modules/mysql2/lib/commands/query.js:60:38)
    at Query.execute (/<proj-dir>/node_modules/mysql2/lib/commands/command.js:45:22)
    at Connection.handlePacket (/<proj-dir>/node_modules/mysql2/lib/connection.js:456:32)
    at Connection.addCommand (/<proj-dir>/node_modules/mysql2/lib/connection.js:478:12)
    at Connection.query (/<proj-dir>/node_modules/mysql2/lib/connection.js:546:17)
    at /<proj-dir>/node_modules/knex/lib/dialects/mysql/index.js:132:18
    at new Promise (<anonymous>)
    at Client_MySQL2._query (/<proj-dir>/node_modules/knex/lib/dialects/mysql/index.js:126:12)
    at Object.executeQuery (/<proj-dir>/node_modules/knex/lib/execution/internal/query-executioner.js:37:17)
    at Client_MySQL2.<anonymous> (/<proj-dir>/node_modules/@mikro-orm/knex/dist/AbstractSqlConnection.js:169:71)
    at /<proj-dir>/node_modules/knex/lib/execution/transaction.js:363:24
    at new Promise (<anonymous>)
```

I think getCharsetNumber should throw an exception immediately when the encoding cannot be recognized, or when the encoding cannot be recognized, give a default encoding (such as utf8), and now this error is confusing.

cweijan:
[...]

PR #2011, opened 2023-05-15
Fix unrecognized encoding causing loss of default encoding

cweijan:


cweijan:
@sidorares any idea? This is to fix the problem that the data cannot be parsed when the database encoding is utf8mb3, there is a more detailed description in this [issue](https://github.com/sidorares/node-mysql2/issues/2009).

Issue #1586, opened 2022-06-14
keepalive not behaving as documented


mrx8:
according to the commit: https://github.com/sidorares/node-mysql2/commit/6b59a3604a53a7f85c8392fcb9577aabbd1eb4df
keepalive is now always on. It cannot be disabled.
I have a problem with this in combination with the libkeepalive-library.
Is it possible to integrate the possibility to disable keepalive in mysql2 once more ?

andrey-sidorov-iag:
hm, the change looks quite out of sync with the comments, I might need to re-read communication leading to in #1081 

@benbotto can you comment?

sidorares:
hm, the change looks quite out of sync with the comments, I might need to re-read communication leading to in https://github.com/sidorares/node-mysql2/pull/1081

@benbotto can you comment?

benbotto:
I think this is a bug--`true` shouldn't be hard coded in the call to `setKeepAlive`.  Seems like #1081 had the right logic, but in #1086 I erroneously removed the `if (this.config.enableKeepAlive)` check.

#1258 seems to correct the issue, mostly, but there's a comment on `Pool.d.ts` about a bad type.  Also might want a `!!` on https://github.com/sidorares/node-mysql2/pull/1258/files#diff-1978d11697f58179bc32b66c6ac7f21cbdfd6649a8f013248a1f9feee25858feR99

```
    this.enableKeepAlive =
      options.enableKeepAlive === undefined || !!options.enableKeepAlive;
```

Not sure why #1258 was closed.

sidorares:
@kn3ny are you keen to finish that work? The PR looks mostly ready, happy to merge it but if you still plan to make a fresh one let us know

n0v1:
I've just created PR #2016 to fix this.

PR #2016, opened 2023-05-17
Respect enableKeepAlive option

n0v1:
This should make sure that the `enableKeepAlive` option is actually taken into account so that it's possible to disable keepalive. This is related to #1081 and #1586.

Before this change keepalive was always enabled, so I opted to let `enableKeepAlive` default to `true`. If enabled by default is not what we should do I'm happy to change it again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,8 @@ const pool = mysql.createPool({
   connectionLimit: 10,
   maxIdle: 10, // max idle connections, the default value is the same as `connectionLimit`
   idleTimeout: 60000, // idle connections timeout, in milliseconds, the default value 60000
-  queueLimit: 0
+  queueLimit: 0,
+  enableKeepAlive: true
 });
 ```
 The pool does not create all connections upfront but creates them on demand until the connection limit is reached.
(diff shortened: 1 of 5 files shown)


wellwelwel:
Although the description says that it's disabled by default, if it was always enabled until today, I think it's interesting to keep it enabled by "new default", so as not to break projects that already use it with the default setting.

@n0v1
Just a question: Naturally, if someone has set this option manually, it will not be affected by this changes, correct?

n0v1:
As I understand it this PR will not change the behaviour in general. For anyone that did not explicitely set the option it will still be enabled like before. When option was set to true it will still be enabled like before. Only when option was set to false keepalive will now be disabled while it was enabled before.

________________________________
Von: Weslley Araújo ***@***.***>
Gesendet: Mittwoch, 17. Mai 2023, 20:41
An: sidorares/node-mysql2 ***@***.***>
Cc: n0v1 ***@***.***>; Mention ***@***.***>
Betreff: Re: [sidorares/node-mysql2] Respect enableKeepAlive option (PR #2016)


Although the description says that it's disabled by default, if it was always enabled until today, I think it's interesting to keep it enabled by "new default", so as not to break projects that already use it with the default setting.

@n0v1<https://github.com/n0v1>
Just a question: Naturally, if someone has set this option manually, it will not be affected by this changes, correct?

sidorares:
LGTM

Issue #1991, opened 2023-05-04
Typescript: Pool Cluster Typing Mismatch


JacobBanghart:
I am working in a typescript node project and running into a interface mismatch. If you clone https://github.com/JacobBanghart/Mysql2PoolClusterIssue and `npm i` you will see that there are 2 issues with the typings for PoolClusters.    

First: uri is not a valid as a property to pass.  
This can be adjusted by one of 2 ways. Either add the property to PoolOptions or better yet to Connection.ConnectionOptions as an optional property.    

Secondly: PoolCluster.of returns a type of PoolCluster which is actually a PoolNamespace.     

I can open a pr for these items if you would like.

adam-farhi:
if I’m not mistaken this is the correct type
```ts
declare class PromisePoolCluster extends EventEmitter {

    config: PoolCluster.PoolClusterOptions;

    add(config: PoolCluster.PoolClusterOptions): Promise<void>;
    add(group: string, config: PoolCluster.PoolClusterOptions): Promise<void>;

    end(): Promise<void>;

    getConnection(): Promise<PoolConnection>;
    getConnection(group: string): Promise<PoolConnection>;
    getConnection(group: string): Promise<PoolConnection>;

    of(pattern: string, selector?: string): PromisePoolCluster;

    on(event: string, listener: Function): this;
    on(event: 'remove', listener: (nodeId: number) => void): this;
    on(event: 'connection', listener: (connection: PoolConnection) => void): this;
}
```

wellwelwel:
@JacobBanghart, can you check this in version `^3.4.1`?

> I also tested it locally using your [example](https://github.com/JacobBanghart/Mysql2PoolClusterIssue) 🙋🏻‍♂️

wellwelwel:
@adam-farhi, if you could also check the promise-based `PoolCluster`, it would great.

wellwelwel:
- You can follow this examples for callbacks:
https://github.com/sidorares/node-mysql2/blob/00508c38a13f1394fb8a6cf3fd88df1353ba0e77/test/tsc-build/mysql/createPoolCluster/of/of.ts#L4-L16

- And this for promise-based:
https://github.com/sidorares/node-mysql2/blob/00508c38a13f1394fb8a6cf3fd88df1353ba0e77/test/tsc-build/promise/createPoolCluster/of/of.ts#L4-L17

---

I'm closing this **Issue**, but feel free to ask anything 🙋🏻‍♂️

PR #2035, opened 2023-06-01
Add correct Types for PromisePoolCluster

adam-farhi:
fix for https://github.com/sidorares/node-mysql2/issues/1991

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -5,6 +5,8 @@ import {
   FieldPacket,
   QueryOptions,
   ConnectionOptions,
+  PoolCluster,
+  PoolClusterOptions,
   PoolOptions,
   Pool as CorePool
 } from './index';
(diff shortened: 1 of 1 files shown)


JacobBanghart:
Following up here based on the linked ticket. In /lib/pool_cluster.js the return of PoolCluster.of is this._namespaces[key] which is a class PoolNameSpace. PoolNameSpace should be what is returned in a `.of` but instead it returns a PoolCluster. A Promise-based version of this does not solve this issue you can see by cloning the repo linked in the ticket and installing your fork locally.

JacobBanghart:
#2036 Adjusts it in the manner that I was describing. Sorry for leaving the issue out there and not getting back to it life gets hecktic.

wellwelwel:
Closing in favor of #2084.

PR #2036, opened 2023-06-01
fix: update pool cluster of to use pool namespace

JacobBanghart:
pool_cluster.js was out of sync with the typing and returns a pool namespace rather than a self-reference pool cluster

fixes: #1991  
This should fix one of the issues. There are some additional fixes that can be made such as having more dedicated typing for the union types.

### DIFF
--- a/typings/mysql/lib/PoolCluster.d.ts
+++ b/typings/mysql/lib/PoolCluster.d.ts
@@ -2,6 +2,8 @@
 import PoolConnection = require('./PoolConnection');
 import {EventEmitter} from 'events';
 import {PoolOptions} from './Pool';
+import {OkPacket, RowDataPacket, FieldPacket, ResultSetHeader} from './protocol/packets/index';
+import Query = require('./protocol/sequences/Query');
 
 declare namespace PoolCluster {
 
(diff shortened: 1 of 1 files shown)


wellwelwel:
@JacobBanghart, in `JS` everything looks fine, but when I tried this changes in `TS`, it works in `mysql2` import, but not in `mysql2/promise`.

A basic test for both import ways:

<hr />

- `mysql2` ✅
  - [x] JS
  - [x] TS

```ts
import mysql from 'mysql2';

const poolCluster = mysql.createPoolCluster();

poolCluster.add('MY_CLUSTER', {
   host: '',
   user: '',
   password: '',
   database: '',
});

poolCluster.getConnection('MY_CLUSTER', (err, conn) => {
   conn.query('SELECT * FROM `table`;', (err, rows) => {
      poolCluster.end();

      console.log(rows);
   });
});
```

<hr />

- `mysql2/promise` ❌
  - [x] JS
  - [ ] TS

```ts

(async () => {
   const poolCluster = mysql.createPoolCluster();

   poolCluster.add('MY_CLUSTER', {
      host: '',
      user: '',
      password: '',
      database: '',
   });

   const MY_CLUSTER = await poolCluster.getConnection('MY_CLUSTER');
   const [rows] = await MY_CLUSTER.query('SELECT * FROM `table`;');

   await MY_CLUSTER.end();

   console.log(rows);
})();
```

- The errors occur because in both import methods the typings are called from callback (`index.d.ts`).

<hr />

If these changes are compatible with both import methods, I think basic tests could be created to ensure that every new change respects the basic usage of the `createPoolCluster`.  

Currently, this covers the most common cases of the `createConnection` and `createPool`.

> You can see a practical example here: [test:tsc-build](https://github.com/sidorares/node-mysql2/blob/master/test/tsc-build/index.ts)

sidorares:
@wellwelwel would you like to join as a co-maintainer? Happy to give you write permissions

JacobBanghart:
> 

So I had issues getting your second example to work on the latest version. It is having other issues unrelated to my change. But thanks for pointing out the relation to the index.d.ts. It seems more practical to place this as a part of the namespace definition. Then I do not modify anything but raw types and use an interface instead. The biggest disadvantage that I see to this is that you are not directly exporting the PoolNamespace for usage, but I can not think of a given use case for exporting it.

wellwelwel:
> would you like to join as a co-maintainer? Happy to give you write permissions

@sidorares, It would be a pleasure. If you prefer, I can reach out to you via email or a most convenient way for you.

wellwelwel:
@JacobBanghart, when you do it in [PoolCluster.d.ts](https://github.com/sidorares/node-mysql2/pull/2036/files#diff-39c1e0418117826e509dec6c2803a2fc4f9a5273513843ab8c70cbaca5dc34a1):

```ts
export interface PoolNamespace {
  getConnection(callback: (err: NodeJS.ErrnoException | null, connection: PoolConnection) => any): void;
  // ...
}
```

Naturally this will affect both `mysql2` and `mysql2/promise`. The types in `PoolNamespace` are appropriate for `mysql2` (because it expects for callbacks), but not for `mysql2/promise`.

<hr />

You can compare these differences, by example, looking at how it works for `mysql2` ([index.d.ts - L89](https://github.com/sidorares/node-mysql2/blob/master/index.d.ts#L89)) in:

```ts
export interface Pool extends mysql.Connection {
  // ...
}
```

And how it changes for `mysql2/promise` ([promise.d.ts - L91](https://github.com/sidorares/node-mysql2/blob/master/promise.d.ts#L91)) in:

```ts
export interface Pool extends mysql.Connection {
  // ...
}
```

Then, adapting how it would be for the `createPoolCluster`.

<hr />

I understand that you didn't modify the `createPoolCluster` itself, but the `of` option within it.

The errors in my `mysql2/promise` example after the `const MY_CLUSTER = poolCluster.of('MY_CLUSTER');` are the simple tests to ensure that the typings are compatible for both use cases.

<hr />

If you don't mind, I can help you with this 🙋🏻‍♂️

sidorares:
@wellwelwel small suggestion - when you link a code line, please link to a commit, not to a branch head ( press `y` to update url to include commit. ), otherwise change to a line above can make you link to incorrect line

wellwelwel:
Closing in favor of #2084.

PR #2043, opened 2023-06-06
fix: keepAliveInitialDelay not taking effect

hechaoWeb:
TCP keepalive interval time will only take effect if setKeepAlive() is called after the socket object is created.


hechaoWeb:
I accidentally discovered that the keepAliveInitialDelay setting was ineffective while using tcpdump, and I had to examine the source code and conduct tests to find the reason.

sidorares:
@hechaoWeb @Kino-Zh I'm curious - are you the same person or it just happens that you had the same issue / discovered a way to debug and fix simultaneously?

Issue #1761, opened 2022-12-27
Property 'releaseConnection' does not exist on type 'Pool'.


amirhossein-karimi:
Hi,

I use the connection pool and according to the documentation, when I release it (connection), I face the following problem .
`Property 'releaseConnection' does not exist on type 'Pool'.`

notes : i used typescript

and this is my code
`pool.getConnection(function (err, conn) {
    // Do something with the connection
    conn.query("CREATE DATABASE IF NOT EXISTS zxc CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;", (err, rows, fields) => {
        if (err) throw err;
    });
    pool.releaseConnection(conn);
})`

jake-knerr:
I am having the same problem. I am using `conn.release()` instead.

JiaJiaJiang:
The doc [here](https://github.com/sidorares/node-mysql2#using-connection-pools) is wrong, please fix it.
The correct one is `pool.release`, not `pool.releaseConnection`

wellwelwel:
@JiaJiaJiang, the documentation is correct.

The `releaseConnection` is used as a method from the `pool`, like:

```js
pool.getConnection((err, conn) => {
  pool.releaseConnection(conn);
});
```

The `release` is used as a method of the connection from `pool`, like:
```js
pool.getConnection((err, conn) => {
  conn.release();
});
```

<hr />

In fact, the type `releaseConnection` is missing in `Pool`.

wellwelwel:
@amirhossein-karimi and @jake-knerr, can you check this in `^3.3.4` version?

wellwelwel:
After version `^3.3.4`, you can follow this example for `releaseConnection`:
https://github.com/sidorares/node-mysql2/blob/00508c38a13f1394fb8a6cf3fd88df1353ba0e77/test/tsc-build/mysql/createPool/callbacks/releaseConnection.ts#L4-L8

And for `release`:
https://github.com/sidorares/node-mysql2/blob/00508c38a13f1394fb8a6cf3fd88df1353ba0e77/test/tsc-build/mysql/createPool/callbacks/release.ts#L4-L7

---

I'm closing this **Issue**, but feel free to ask anything 🙋🏻‍♂️

Issue #1563, opened 2022-04-20
Promise Pool: Inconsistent types, missing releaseConnection function


perry-mitchell:
Hi,

I'm looking at implementing a connection pool, using the promise entry, but as I can see [here](https://github.com/sidorares/node-mysql2/blob/131c6b852127bc3af33f542f60231bf19cf99509/promise.js#L318) there isn't a `releaseConnection` function.

Additionally, when looking at the standard connection pool **types**, the `releaseConnection` method is not [typed at all](https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/Pool.d.ts#L48).

sidorares:
there is a `PromiseConnection.release` - https://github.com/sidorares/node-mysql2/blob/131c6b852127bc3af33f542f60231bf19cf99509/promise.js#L87-L89 which is typed here https://github.com/sidorares/node-mysql2/blob/dbb344e89a1cc8bb457b24e67b07cdb3013fe844/promise.d.ts#L85

perry-mitchell:
You're right, it is. Couldn't find it in the documentation.

What about the types for `releaseConnection` on the standard pool?

sidorares:
> You're right, it is. Couldn't find it in the documentation.

I think the promise wrapper is increasingly becoming the most popular api and deserve place at the beginning of readme, while callback style api can reuse documentation from `mysqljs/mysql`

> What about the types for `releaseConnection` on the standard pool?

looks like it's outdated, now deprecated `PoolConnection.end()` used to be a way to release connection back into the pool

sidorares:
@perry-mitchell would you like to contribute a fix to missing typings? If you can think of some test harness that would be awesome ( have a unit test in TS that can both run without runtime errors and pass static type checks )

sidorares:
I'll close this one as we also discuss same issue in #1761

PR #2053, opened 2023-06-11
fix: releaseConnection types and promise

wellwelwel:
This fixes the #1761 and #1563.

<hr />

### JS

The [promise.js](https://github.com/sidorares/node-mysql2/blob/f930f6e0453f6ffb62f46dec40825f0bff2307a1/promise.js) didn't have the method `releaseConnection`.

Instead of creating from scratch, I just "extended" it from the `release` method:

```js
releaseConnection(connection) {
  if (connection instanceof PromisePoolConnection) connection.release();
}
```

Then, I added the test to [test/integration/promise-wrappers/test-promise-wrappers.js](https://github.com/sidorares/node-mysql2/compare/master...wellwelwel:node-mysql2:release-connection#diff-5b7f569fa112eba31069c6d57e2a1958dd38979a02880434a4f0b02057dac748).

<hr />

### TS

I had to fix a few wrong types before to fix the `releaseConnection`.

Before this **PR**, according by typings, this should be possible:

```js
await (await (await pool.getConnection()).getConnection()).getConnection();
// To infinity and beyond... 😵‍💫
```

So, I fixed it to only `Pool` have the method `getConnection`.

- At this time, I added "negative" tests in `tsc-build` to ensure that the above code doesn't occur again.

<hr />

Lastly, I added some tests for each method that I changed something in types:

- `releaseConnection`
- `release`
- `getConnection`
- `connection`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,7 @@
 import {
   Connection as PromiseConnection,
-  PoolConnection as PromisePoolConnection,
+  Pool as PromisePool,
+  PoolConnection as PromisePoolConnection2,
 } from './promise';
 
 import * as mysql from './typings/mysql';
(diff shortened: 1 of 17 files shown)


wellwelwel:
@sidorares, this **PR** demands a patch release, can I do it or in these cases I wait for you?

Issue #2052, opened 2023-06-11
Packet.readLengthCodedNumberExt exception with mysql 8


wsporto:
I am getting the exception **Packet.readLengthCodedNumberExt** when I try to execute the prepare statement with a mysql 8 database. This works fine when I run with mysql 5.

```
Trace:
at Packet.readLengthCodedNumberExt (C:\dev\mysql2-error\node_modules\mysql2\lib\packets\packet.js:231:13)
    at Packet.readLengthCodedNumber (C:\dev\mysql2-error\node_modules\mysql2\lib\packets\packet.js:193:17)
    at new ColumnDefinition (C:\dev\mysql2-error\node_modules\mysql2\lib\packets\column_definition.js:28:33)
    at Prepare.readParameter (C:\dev\mysql2-error\node_modules\mysql2\lib\commands\prepare.js:80:17)
    at Prepare.execute (C:\dev\mysql2-error\node_modules\mysql2\lib\commands\command.js:45:22)
    at Connection.handlePacket (C:\dev\mysql2-error\node_modules\mysql2\lib\connection.js:490:32)
    at PacketParser.onPacket (C:\dev\mysql2-error\node_modules\mysql2\lib\connection.js:95:12)
    at PacketParser.executeStart (C:\dev\mysql2-error\node_modules\mysql2\lib\packet_parser.js:75:16)
    at Socket.<anonymous> (C:\dev\mysql2-error\node_modules\mysql2\lib\connection.js:102:25)
    at Socket.emit (node:events:527:28)
  ```

Code to reproduce:

```ts
const mysql = require('mysql2/promise');

/* 
docker run --name mysql-dev -e MYSQL_ROOT_PASSWORD=password -e MYSQL_DATABASE=mydb 
-p 3306:3306 mysql:8.0.33 --default-authentication-plugin=mysql_native_password
*/
async function run() {
    const conn = await mysql.createConnection({
        host: 'localhost',
        user: 'root',
        password: 'password',
        database: 'mydb'
    });
    await conn.query(`
        CREATE TABLE IF NOT EXISTS mytable1 (
            id int NOT NULL AUTO_INCREMENT,
            value int,
            PRIMARY KEY (id)
        ) ENGINE=InnoDB
    `);
    await conn.prepare('select value from mytable1 order by ?');
}

run().then(() => {
    console.log("ok");
});
```
Using mysql2 v3.3.3

wellwelwel:
[...]

PR #2056, opened 2023-06-12
fix: handle prepare response when number of fields in the prepare header does not match the actual number in the response

sidorares:
ref: #2052 

*Update*: main reason for the #2052 is not the medatdata_follows flag and/or missing parameters. The issue is caused by prepare header returning number of placeholders in the query ( 1 for `select * from user order by ?` ) but the actual number of parameters following the header is 0 - see https://github.com/sidorares/node-mysql2/issues/2052#issuecomment-1620318928. Changing the scope of this PR to handle "ignored" parameters / columns

------

Context: we currently expect that prepare response has exactly num fields + num parameters ColumnnDefinition packets in the response. However ether of them could be skipped by the server even when num fields is > 0 and/or num parameters > 0.

See https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_prepare.html

> if num_params > 0 and CLIENT_OPTIONAL_RESULTSET_METADATA is not set or if medatdata_follows is RESULTSET_METADATA_FULL num_params packets will follow. Then a [EOF_Packet](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_eof_packet.html) will be transmitted, provided that CLIENT_DEPRECATE_EOF is not set.

 - [x] allow to handle exception from test example in #2043
 - [x] I plan to gradually switch to a built in node test runner. Add a setup to run new tests in node v20 using [built in runner](https://nodejs.org/api/test.html)
 - [ ] add a unit test (initially failing)

Descoping:
 - [ ] parse `metadata_follows` in COM_STMT_PREPARE_OK packet deserealisation
 - [ ] expect possible EOF packet where we currently read ColumnDefinition based on `OPTIONAL_RESULTSET_METADATA ` / `DEPRECATE_EOF` connection flags and metadata_follows COM_STMT_PREPARE_OK packet
 - [ ] verify that empty parameters / fields array does not break rest of the "execute PS" logic. For example, we currently might fail `.execute()` call if expected and actual number of parameters do not match. In that case 2 options: store them separately ( reported number of parameters and array of deserialised parameter metadata ( could be empty ) OR fill array with default / unknown metadata.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -38,12 +38,12 @@ jobs:
         include:
           # 20.x 
           - node-version: "20.x"
-            mysql-version: "mysql:8.0.18"
+            mysql-version: "mysql:8.0.33"
             use-compression: 1
             use-tls: 0
             use-builtin-test-runner: 1
           - node-version: "20.x"
-            mysql-version: "mysql:8.0.18"
+            mysql-version: "mysql:8.0.33"
             use-compression: 0
             use-tls: 1
             use-builtin-test-runner: 1
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -41,10 +41,12 @@ jobs:
             mysql-version: "mysql:8.0.18"
             use-compression: 1
             use-tls: 0
+            use-builtin-test-runner: 1
           - node-version: "20.x"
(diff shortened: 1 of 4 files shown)


sidorares:
I messed up something with cherry picking, recreating this in a new PR

Issue #2059, opened 2023-06-12
PoolConnection type error after update v3.3.4


matvejs16:
Getting error: TS2551: Property 'getConnection' does not exist on type 'PoolConnection'. Did you mean 'connection'?
```
const connection = await DB.Handle_Promise.getConnection()
```

wellwelwel:
Hi @matvejs16, I'm gradually looking for typings.

The `PoolConnection` is the pool connection itself. After a lot tests in **JS** to get the conflicts between the types, always when I tried to get a connection from a Pool Connection, it gives me an error. Then, in **PR** #2053, I changed it to only the Pool can get a connection, as seen in the [documentation](https://github.com/sidorares/node-mysql2/blob/1a9972ddeb6561b0848eaedfcaf05f667ea180b4/README.md?plain=1#L157) and according with [./lib/pool_connection.js](https://github.com/sidorares/node-mysql2/blob/1a9972ddeb6561b0848eaedfcaf05f667ea180b4/lib/pool_connection.js), you can see that it doesn't have the the `getConnection` method.

If your code works in **JS** (ignoring the types), can you share a *basic repro* example?

matvejs16:
Yep, JS is working ok
```
import mysql from 'mysql2'
import mysqlPromise from 'mysql2/promise'

interface DB {
    Handle: mysql.Pool | null,
    Handle_Promise: mysqlPromise.PoolConnection | null,
    Connect(callback: Function): void,
    escape(string: string | number | object | Array<any>): string | number | null,
    getPromiseConnection(): mysqlPromise.PoolConnection | undefined
}

const DB: DB = {
    Handle: null,
    Handle_Promise: null,
    Connect: function(callback: Function) {
        this.Handle = mysql.createPool({
            connectionLimit: 1000,
            host: 'IP',
            user: 'USER',
            password: 'PASS',
            database: 'DATABASE',
            debug: false,
            connectTimeout: 10000,
            multipleStatements: true
        })
        this.Handle.query('SET NAMES utf8mb4;');
        this.Handle.query('SET CHARACTER SET utf8mb4');
        this.Handle.query('SET COLLATION_CONNECTION="utf8mb4_general_ci"');
        this.Handle_Promise = this.Handle.promise()
        callback()
    },
    escape: function(string) {
        if (string == 'NULL') return string
        return mysql.escape(string)
    }
}

DB.Connect(function () {
    // @ts-ignore
[...]

PR #2060, opened 2023-06-12
fix: createPool promise as PromisePool

wellwelwel:
Fixes the issue #2059.

<hr />

Before this **PR**, the `.promise()` typing tests started from `.promise()['any_tested_method']`, but the `.promise()` itself was not tested.

#### Credit:
- This test was based on the the code from https://github.com/sidorares/node-mysql2/issues/2059#issuecomment-1587924922 by @matvejs16.


sidorares:
LGTM

PR #2067, opened 2023-06-15
refactor: unifies spread types in their proper places

wellwelwel:
It's a large **TypeScript** refactoring that includes every single `*.ts` file in the entire project.

> **It'd be great if someone could review it** 🚀

<hr />

### Why?

Many types are confusing because they are redefined several times, for example:

- `Pool` types are spread  and defined in `index.d.ts`, `typings/mysql/index.d.ts` and `typings/mysql/lib/Pool.d.ts` at the same time
  - The same for `Connection`, `ConnectionOptions`, `execute`, `query`, etc.
- `execute` and `query` are complex and have to do several <kbd>ctrl</kbd> + <kbd>c</kbd> and <kbd>v</kbd> around the types
- Sometimes a type is only present in one of them, in other as overwritten and another time it's just repeated

<hr />

### To consider

- **It makes new TS contributions much easier:**
  - Do you want to change some type in `Connection`? It's simple: go to `Connection.d.ts` file 👨🏻‍💻
- It makes creating new **TS** tests more practical
- By certain way, many past changes that have caused **TS** errors is because the types have been spread around

<hr />

### CI

- `npm run lint:code` now checks for every `*.ts` file (including the tests)

<hr />

### What does this refactoring change?

> In final usage, nothing. It's just a refactoring, but... 🤹🏻‍♀️

- ✅ These changes resolves **100%** of `eslint` errors in `*.ts`

- ✅ Unifies the types in their proper places:
  - `Pool` as example: now, all typings are defined in `typings/mysql/lib/Pool.d.ts`
  - `execute` and `query` overloads too. Now both can be reused:
    - If you change something in type `execute` or `query`, it will automatically affect all types that need it (`Connection`, `PoolConnection`, etc.) 🤹🏻‍♀️

- ✅ Increments `NodeNext` TypeScript compatibility by requires the `.js` in local imports for `*.ts` files

- ✅ Resolves circular imports in every `*.ts` file

- ✅ Patterns all imports as `import any from './'`
  - Before, it was mixed with `import any = require('./')`

- ℹ️ Removes older **TypeScript** tests that are unused
  - Then, removes `@types/mocha`, `@types/chai`, `mocha` and `chai` dependencies

<hr />

### Lastly

It's the new **index.d.ts**:
```ts
export * from './typings/mysql/index.js';
```
>  Yes, it's everything, because each type is in its proper place and the `index.d.ts` just calls them 🥷🏻✨


wellwelwel:
I would appreciate if someone could review it 🙋🏻‍♂️

wellwelwel:
Just showing in practice that no exports / imports was changed in their final usage:

<hr />

#### From **npm**
  - from-npm

#### From these changes
  - from-pr

<hr />

> The same for `mysql2/promise` ✨

sidorares:
great work @wellwelwel 
I just glanced over the changes I think it looks good but would be really good if someone else could take another look

wellwelwel:
> [...] but would be really good if someone else could take another look

@sidorares

For a moment, I considered tag **#jmordica**. He does the most of the *original* types and created a [license file](https://github.com/sidorares/node-mysql2/blob/fefe036512624b384d48c926b11e7a986f7ae57c/typings/mysql/LICENSE.txt) on his changes, but that was in 2016.

wellwelwel:
@sidorares, I want to bring up another point:

<hr />

The code bellow shows a type error that it's happens in the current release, but it was fixed in merged PR #2053.

  - Screenshot 2023-06-16 at 16 51 54

Using `tsconfig.json` like:
```json
{
  "include": ["index.ts"],
  "compilerOptions": {
    "esModuleInterop": true,
    "moduleResolution": "node"
  }
}
```

<hr />

But everything will work fine if I use the `tsconfig.json` like:
```json
{
  "include": ["index.ts"],
  "compilerOptions": {
    "esModuleInterop": true,
    "moduleResolution": "nodenext"
  }
}
```

<hr />

It occurs because in `index.d.ts`, it's correct:
- https://github.com/sidorares/node-mysql2/blob/fefe036512624b384d48c926b11e7a986f7ae57c/index.d.ts#L156

But in `Pool.d.ts`, it doesn't have the Promise "version":
- https://github.com/sidorares/node-mysql2/blob/fefe036512624b384d48c926b11e7a986f7ae57c/typings/mysql/lib/Pool.d.ts#L64

<hr />

### From this PR

In this refactoring, since all the `Pool` types are in `Pool.d.ts`, naturally this code will work for both `tsconfig.json` and for example, the issue #2059 would probably not exist.

✅ Checking:
- Screenshot 2023-06-16 at 17 35 31

wellwelwel:
@sidorares, I'm thinking of closing this **PR**.
A large change like that it's hard to get a good review.

Instead, I'm considering bringing here a **Milestone** that I have into a private repository where I test everything before bringing it here.

Then, reach exactly at the same place as this **PR**, but in stages.

> Independently of closing or not, I think it's good to bring it here.
> By this way, my PRs would be more easily planned and everyone would be aware of everything

wellwelwel:
See #2072

PR #2069, opened 2023-06-16
Support parsing gtids from OK package

sanglt:


lib/packets/resultset_header.js
```diff
@@ -72,6 +73,10 @@ class ResultSetHeader {
             stateChanges.trackStateChange = packet.readLengthCodedString(
               encoding
             );
+          } else if (type === sessionInfoTypes.STATE_GTIDS) {
+            const something = packet.readLengthCodedString(encoding);
```
sidorares, line 77: maybe better to be a bit more verbose:

```suggestion
            // TODO: find if the first length coded string means anything. Usually comes as empty
            const _unknownString = packet.readLengthCodedString(encoding);
```

sanglt: Thanks, I updated the code and also add the skip eslint check for this line.

wellwelwel (changes requested):
```diff
- // eslint-disable-next-line @typescript-eslint/no-unused-vars
+ // eslint-disable-next-line no-unused-vars
```

You can test the `lint` locally by perform: `npm run lint` 🙋🏻‍♂️

<hr />

> I noticed that you have already fixed this 🤹🏻‍♀️

wellwelwel (approved):

sidorares:
@sanglt you initialise resulting object with `gtids: []` and then assign a value to `stateChanges.gtid = ...` ( non-plural name )
I think it should be `stateChanges.gtids`

sidorares:
before this is merged: I'd like to do a bit of research to confirm if list comes as one SessionInfo object, comma-separated values in a string or potentially multiple type 3 SessionInfo objects ( or mix of both )

sidorares:
I think this this part of the server that serialises gtids SessionInfo: https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/session_tracker.cc#L269

sidorares:
Looks like the format ls comma-separated list:

`SID:GNO:GNO,SID:GNO,SID:GNO,...`

https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/rpl_gtid.h#L1762-L1785
https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/rpl_gtid_set.cc#L760-L824

Not sure yet if it's possible to have more than one gtid session info and how should we handle that ( this PR implementation: last one 'wins' ). I'm happy to merge with previous suggestions fixed ( use `gtids` property consistently )

Also from the mysql server code it looks like "zero length first string" is encoding, but at the same time looks like its never set. We can keep `_unknownString` name for now - maybe add a link to the source code to help clarifying later - https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/session_tracker.cc#L274

sanglt:
> Looks like the format ls comma-separated list:
> 
> `SID:GNO:GNO,SID:GNO,SID:GNO,...`
> 
> https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/rpl_gtid.h#L1762-L1785 https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/rpl_gtid_set.cc#L760-L824
> 
> Not sure yet if it's possible to have more than one gtid session info and how should we handle that ( this PR implementation: last one 'wins' ). I'm happy to merge with previous suggestions fixed ( use `gtids` property consistently )
> 
> Also from the mysql server code it looks like "zero length first string" is encoding, but at the same time looks like its never set. We can keep `_unknownString` name for now - maybe add a link to the source code to help clarifying later - https://github.com/mysql/mysql-server/blob/ea7087d885006918ad54458e7aad215b1650312c/sql/session_tracker.cc#L274

Yes, the definition of GTID set from MySQL document is:

```
A GTID set is a set comprising one or more single GTIDs or ranges of GTIDs.
...
A GTID set can include any combination of single GTIDs and ranges of GTIDs, and it can include GTIDs originating from different servers. This example shows the GTID set stored in the [gtid_executed](https://dev.mysql.com/doc/refman/5.7/en/replication-options-gtids.html#sysvar_gtid_executed) system variable (@@GLOBAL.gtid_executed) of a replica that has applied transactions from more than one source:

2174B383-5441-11E8-B90A-C80AA9429562:1-3, 24DA167-0C0C-11E8-8442-00059A3C7B00:1-19
```

On my test, depend on the value of `[session_track_gtids](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_session_track_gtids)` `OWN_GTID ` or `ALL_GTIDS ` the return values will be single set or multiple set respectively.

sidorares:
Would be really good to add a documentation section covering all supported session track types

sidorares:
@sanglt should be in npm shortly as v3.4.0

Issue #2072, opened 2023-06-18
TypeScript: missing, wrong and refactoring types


wellwelwel:
> Please, don't use this area to indicate new Issues, give preference to [opening a new Issue](https://github.com/sidorares/node-mysql2/issues/new).

### Pool
- [x] Missing Connection URI overload in `mysql` import, like `createPool('uri')`
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)
- [x] Confusing `maxIdle` option description
  - [#2070](https://github.com/sidorares/node-mysql2/issues/2070)
  - [#1858](https://github.com/sidorares/node-mysql2/pull/1858)
  - [#2073](https://github.com/sidorares/node-mysql2/pull/2073)

### Pool Cluster
> It doesn't have any tests yet ⚠️
- [x] Missing Connection URI overload, like `poolCluster.add('conn', 'uri')`
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)
- [x] Missing config option `url` in `add` method
  - [#1991](https://github.com/sidorares/node-mysql2/issues/1991)
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)
- [x] `end`, `of` and `add` methods can't be in a `PoolCluster` connection
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)
- [x] `of` type can't be a `PoolCluster`
  - [#1991](https://github.com/sidorares/node-mysql2/issues/1991)
  - [#2036](https://github.com/sidorares/node-mysql2/pull/2036)
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)
- [x] Pool Cluster is missing in `promise.d.ts`
  - [#2035](https://github.com/sidorares/node-mysql2/pull/2035)
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)

### Constants
- [x] Missing export for `Types` (it's necessary to fix both **JS** and **TS**)
  - [#1938](https://github.com/sidorares/node-mysql2/issues/1938)
  - [#2084](https://github.com/sidorares/node-mysql2/pull/2084)

<hr>

### Notes

- All fixes need to be covered by tests ([see practical examples here](https://github.com/sidorares/node-mysql2/blob/fefe036512624b384d48c926b11e7a986f7ae57c/test/tsc-build/index.ts))
- If new issues are identified, they will be added here

<hr>

<details open>

<summary>
  <b>🔬 About TypeScript Tests</b>
[...]

PR #2075, opened 2023-06-19
refactor: unifies spread types in their proper places

wellwelwel:
This **PR** is the last part of the TypeScript refactoring (#2072).
As before in #2067, the final usage is unaffected by these changes.
> By separating the **PRs**, I believe it's now much better to review the changes 🔬

<hr>

### Unifying the types
- I'll use the `Pool` as example
- The same idea goes for all the other types that have been spread out

All `Pool` types spread across the [`index.d.ts`](https://github.com/sidorares/node-mysql2/blob/37161a2a0b11e9ca916d38e64709f4bbc32f8af0/index.d.ts), [`typings/mysql/index.d.ts`](https://github.com/sidorares/node-mysql2/blob/37161a2a0b11e9ca916d38e64709f4bbc32f8af0/typings/mysql/index.d.ts) and [`typings/mysql/lib/Pool.d.ts`](https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/Pool.d.ts) have been unified in *"typings/mysql/lib/Pool.d.ts"*.

<hr>

### About the approach of reusable `query` and `execute` overloads
- I'll use the `execute` as example
- The same idea goes for the `query` overloads

I moved all variations (overloads) of the `execute` around the types into *"typings/mysql/lib/protocol/sequences/ExecutableBase.ts"*.

`Pool` as example is a *class*, not an interface, so I did the `ExecutableBase` as an *Anonymous Class*.
> I chose this method because it allows to dynamic extending the `QueryableBase` and `Emitter` like a *Function*.

For example:
```ts
declare class Pool extends QueryableBase(ExecutableBase(EventEmitter)) { /* ... */ }
```

> It doesn't changes the `execute` or `query` types, just where they come from.

<details>
<summary>
<h4>How this works in practice:</h4>
</summary>

*Before*
Screenshot 2023-06-19 at 08 41 25

*After*
Screenshot 2023-06-19 at 08 42 42

</details>

<hr>

### Notes

- No types were added or removed, only unified 🙋🏻‍♂️


wellwelwel:
@sidorares, if you think it's better, I can do a test to **strictly check** the returned types for each overload of `execute` and `query`.

Since it's just a refactoring, I didn't focus on tests, but I believe this could be a safer approach.

<hr>

Something like:

```ts
import mysql from 'mysql2';

const conn = mysql.createConnection({});

conn.execute<mysql.RowDataPacket[]>('', (_err, result, fields) => {
   const _r: mysql.RowDataPacket[] = result;
   const _f: mysql.FieldPacket[] = fields;
});

conn.execute<mysql.RowDataPacket[][]>('', (_err, result, fields) => {
   const _r: mysql.RowDataPacket[][] = result;
   const _f: mysql.FieldPacket[] = fields;
});

conn.execute<mysql.OkPacket>('', (_err, result, fields) => {
   const _r: mysql.OkPacket = result;
   const _f: mysql.FieldPacket[] = fields;
});

conn.execute<mysql.OkPacket[]>('', (_err, result, fields) => {
   const _r: mysql.OkPacket[] = result;
   const _f: mysql.FieldPacket[] = fields;
});

conn.execute<mysql.ResultSetHeader>('', (_err, result, fields) => {
   const _r: mysql.ResultSetHeader = result;
   const _f: mysql.FieldPacket[] = fields;
});
```

```ts

const conn = await mysql.createConnection({});

const [_r0, _f0]: [mysql.RowDataPacket[], mysql.FieldPacket[]] = await conn.execute<mysql.RowDataPacket[]>('');
const [_r1, _f1]: [mysql.RowDataPacket[][], mysql.FieldPacket[]] = await conn.execute<mysql.RowDataPacket[][]>('');
const [_r2, _f2]: [mysql.OkPacket, mysql.FieldPacket[]] = await conn.execute<mysql.OkPacket>('');
const [_r3, _f3]: [mysql.OkPacket[], mysql.FieldPacket[]] = await conn.execute<mysql.OkPacket[]>('');
const [_r4, _f4]: [mysql.ResultSetHeader, mysql.FieldPacket[]] = await conn.execute<mysql.ResultSetHeader>('');
```

> The same for `query`.

<hr>

These tests aren't the same as the existing tests for the `execute` and `query`.

wellwelwel:
After thinking for a while, I decided to create a **strict** test to check each overload and its return types of `execute` and `query`.

This will ensure that both this refactoring and any future changes to the `execute` and `query` methods are tested for variations of usage and their corresponding returns in each variation of usage 🤹🏻‍♀️

Now I can say that this part of the #2072 is completed. That will make the fix and the creation of the promise-based `PoolCluster` much easier 🎉

> @sidorares, in that case, if approved, I would like you to handle the merge.

PR #2095, opened 2023-06-29
fix:  typings

mzstic:
`acquireTimeout` exists only on this one place in repo and when I tried to use it, it was only complaining with deprecation warning that such parameter does not exist and in future versions it will be an error.

As it's not being used anywhere else in the actual code I think it can be safely removed from the TS types as well.

### DIFF
--- a/typings/mysql/lib/Pool.d.ts
+++ b/typings/mysql/lib/Pool.d.ts
@@ -11,12 +11,6 @@ import { ExecutableBase } from './protocol/sequences/ExecutableBase.js';
 
 export interface PoolOptions extends ConnectionOptions {
   /**
-   * The milliseconds before a timeout occurs during the connection acquisition. This is slightly different from connectTimeout,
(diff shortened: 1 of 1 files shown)


wellwelwel:
LGMT.

Just complementing the warning:
```
Ignoring invalid configuration option passed to Connection: acquireTimeout. This is currently a warning, but in future versions of MySQL2, an error will be thrown if you pass an invalid configuration option to a Connection
```

wellwelwel:
For better *changelog*, I think a commit message like "fix: remove acquireTimeout invalid option" would be good.

What you think, @mzstic?

sidorares:
@mzstic I'll leave it to you to handle. You can tweak commit message that goes to changelog when you click "squash and merge" button, up to you if you want to communicate this to contributors for the sake of knowledge sharing or do it yourself

wellwelwel:
Thanks @sidorares, I didn't know that.

Issue #1934, opened 2023-04-04
Typing for stored procedure results wrong


joshuajaharwood:
When executing a stored procedure, results are returned in the following shape:

`[[[RowDataPacket], ResultSetHeader], ColumnDefinition[]]`

This can't be typed by passing in a RowDataPacket overload to the..

```typescript
query<T extends RowDataPacket[][] | RowDataPacket[] | OkPacket | OkPacket[] | ResultSetHeader>(
    options: QueryOptions
  ): Promise<[T, FieldPacket[]]>;
```

...query method as T's bounds don't allow for the actual result format.

Here's an example of the returned result:

```json
[
  [
    [
      {
        "rowResult1": true
      },
      {
        "row2Result2": true
      }
    ],
    {
      "fieldCount": 0,
      "affectedRows": 0,
      "insertId": 0,
      "info": "",
      "serverStatus": 34,
      "warningStatus": 0
    }
  ],
  [
    [
      {
        "_buf": {
          "type": "Buffer",
          "data": ["data was removed"]
        },
        "_clientEncoding": "utf8",
        "_catalogLength": 3,
        "_catalogStart": 10,
        "_schemaLength": 0,
        "_schemaStart": 14,
        "_tableLength": 0,
        "_tableStart": 15,
        "_orgTableLength": 0,
        "_orgTableStart": 16,
        "_orgNameLength": 15,
        "_orgNameStart": 33,
        "characterSet": 63,
        "encoding": "binary",
        "name": "column_name",
        "columnLength": 10,
        "columnType": 10,
        "type": 10,
        "flags": 128,
        "decimals": 0
      },
      {"others":  "removed"}
    ],
    null
  ]
]
```

I'm using mysql2 v3.2.0.

adammflax:
Hi Sidorares,
I'm struggiling to visualise all these situations if you give a couple of examples of output we can try and make a better type

sidorares:
some examples are in the unit tests:

https://github.com/sidorares/node-mysql2/blob/ef7bac8a994b9bd8073b93212f303ad227ba4c61/test/integration/connection/test-binary-multiple-results.js#L27-L106

https://github.com/sidorares/node-mysql2/blob/ef7bac8a994b9bd8073b93212f303ad227ba4c61/test/integration/connection/test-multiple-results.js#L27-L105

wellwelwel:
@joshuajaharwood, thanks for report it.

[...]

PR #2100, opened 2023-06-30
fix: create ProcedureCallPacket typings

wellwelwel:
This **PR** looks to fix the Issue #1934.

---

<details>
<summary>
<b>Debugging Process</b>
</summary>

- [x] **RowDataPacket[]**
`[ ...RowDataPacket, ResultSetHeader ]`
- [x] **RowDataPacket[][]**
`[ ...[ ...RowDataPacket ], ResultSetHeader ]`
- [x] **OkPacket**
`ResultSetHeader`
- [x] **OkPacket[]**
`ResultSetHeader`
- [x] **ResultSetHeader**
`ResultSetHeader`
- [x] *Executing an empty Procedure*
`ResultSetHeader`
- [x] *an empty result of* **RowDataPacket[]**
`[ [], ResultSetHeader ]`
- [x] *an empty result of* **RowDataPacket[][]**
`[ [], ResultSetHeader ]`
- [x] **Multiple Statements**
  - [x] As the same
    - [x] OkPacket, ResultSetHeader
    `ResultSetHeader`
    - [x] RowDataPacket (two or more queries)
    `[ ...[ ...RowDataPacket ], ...[ ...RowDataPacket ], ResultSetHeader ]` 
  - [x] By combining
    - [x] **INSERT**, **UPDATE** and **DELETE**
    `ResultSetHeader`
    - [x] **INSERT**, **UPDATE**, **SELECT** and **DELETE**
    `[ ...[ ...RowDataPacket ], ResultSetHeader ]`
    - [x] **INSERT**, **SELECT**, **UPDATE**, **SELECT** and **DELETE**
    `[ ...[ ...RowDataPacket ], ...[ ...RowDataPacket ], ResultSetHeader ]` 

</details>

---

### Debugging Conclusion
- Always when a **SELECT** is executed (by itself, multiple queries, with or without results and including together with **INSERT**, **UPDATE**, etc.), the result will be an array with the returned rows (or `rowsAsArray`) and the last item from this array always be a `ResultSetHeader`.
- For **INSERT**, **UPDATE**, **DELETE**, **TRUNCATE**, **DROP**, empty Procedures, etc., the result always be a `ResultSetHeader` object (not an array).
- In any case, the `ResultSetHeader` (as a direct object or as the last array item) calculates the affected rows of all queries and it don't returns the `insertId` (even if it's a single INSERT query inside the Procedure).

---

In case no **SELECT** query is used, it's more convenient to just perform `query<ResultSetHeader>`.

But when it has one or more **SELECT** queries (even with empty results), after debugging, it's safe to say that the **Procedure Calls** have different behaviors. These behaviors aren't included in the current typings.

> The way I see it currently there is nothing wrong and added `ProcedureCallPacket` is just a convenience for when you know returned shape is `[Resultset, OkPacket]`

When using **TypeScript**, it's has some rules that can be bypassed by using `as`.

For example, I can use the `as` to "force" a type, but I can't do that:
```ts
const n = 'a' as number; // ❌
```

But, I can perform a redundant `as` by using it together with `unknown`
```ts
const number = 'a' as unknown as number; // ✅ (maybe?)
```

Going by this approach, we would need to do this:
```ts
(await conn.query(sql)) as unknown as [[...RowDataPacket[], ResultSetHeader], FieldPacket];
```

But that is not the only problem.

The current types can be misleading by believing that every Procedure Call will be a direct `ResultSetHeader` object.

It can also be confusing when using as `RowDataPacket[]` or `RowDataPacket[][]`, especially since the `ResultSetHeader` isn't detached, but the last item in the array (together with the rows).

By explicitly indicating these possibilities, it ensure that the Procedure Call return can be handled properly.

---

### Conclusion
Defining the proper types for Procedure Calls is not only a practical solution, but also a safer one.

That said, this **PR** proposes to make possible catch the proper typings returned from **Procedure Calls**.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,6 +46,7 @@
       "rules": {
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "off",
+        "arrow-parens": "off",
         "no-restricted-syntax": [
           "error",
           {
--- a/test/tsc-build/helpers.ts
+++ b/test/tsc-build/helpers.ts
@@ -1,0 +1,37 @@
+import { mysql, mysqlp } from './index.js';
+
+export const isResultSetHeader = (
+  data: unknown
+): data is mysql.ResultSetHeader | mysqlp.ResultSetHeader => {
+  if (!data || typeof data !== 'object') return false;
+
+  const keys = [
+    'fieldCount',
+    'affectedRows',
+    'insertId',
+    'info',
+    'serverStatus',
+    'warningStatus',
+  ];
+
+  return keys.every((key) => key in data);
+};
+
+export const isOkPacket = (
+  data: unknown
+): data is mysql.OkPacket | mysqlp.OkPacket => {
+  if (!data || typeof data !== 'object') return false;
+
+  const keys = [
+    'fieldCount',
+    'affectedRows',
+    'changedRows',
+    'insertId',
+    'serverStatus',
+    'warningCount',
+    'message',
+    'procotol41',
+  ];
+
+  return keys.every((key) => key in data);
+};
--- a/test/tsc-build/strict-checks/ProcedureCallPacket.ts
+++ b/test/tsc-build/strict-checks/ProcedureCallPacket.ts
@@ -173,14 +173,14 @@ const procedureCall = {
 
   // Checking `ResultSetHeader | OkPacket | OkPacket[]` Procedure Calls
   {
-    await conn.query(dropProcedure.select);
-    await conn.query(createProcedure.select);
+    await conn.query(dropProcedure.update);
+    await conn.query(createProcedure.update);
 
     const [procedureResult] = await conn.query<
       mysqlp.ProcedureCallPacket<
         mysqlp.ResultSetHeader | mysqlp.OkPacket | mysqlp.OkPacket[]
       >
-    >(procedureCall.select, []);
+    >(procedureCall.update, []);
 
     // Strict checking the `ResultSetHeader`
     const resultSetHeader: mysqlp.ResultSetHeader = procedureResult;
--- a/test/tsc-build/strict-checks/ProcedureCallPacket.ts
+++ b/test/tsc-build/strict-checks/ProcedureCallPacket.ts
@@ -1,0 +1,189 @@
+import { mysql, mysqlp } from '../index.js';
+import { access } from '../mysql/baseConnection.js';
+import { isResultSetHeader } from '../helpers.js';
+
+const dropProcedure = {
+  select: 'DROP PROCEDURE IF EXISTS selectProcedure',
+  update: 'DROP PROCEDURE IF EXISTS updateProcedure',
+};
+
+const createProcedure = {
+  select: `
+    CREATE PROCEDURE selectProcedure()
+    BEGIN
+      SELECT 1 as id;
+    END
+  `,
+  update: `
+    CREATE PROCEDURE updateProcedure()
+    BEGIN
+      SET @a = 1;
+    END
+  `,
+};
+
+const procedureCall = {
+  select: 'CALL selectProcedure()',
+  update: 'CALL updateProcedure()',
+};
+
+// Callback
+(() => {
+  interface User extends mysql.RowDataPacket {
+    id: number;
+  }
+
+  const conn = mysql.createConnection(access);
+  const connAsRow = mysql.createConnection({ ...access, rowsAsArray: true });
+
+  // Checking `RowDataPacket[]` Procedure Calls
+  conn.query(dropProcedure.select, () => {
+    conn.query(createProcedure.select, () => {
+      conn.query<mysql.ProcedureCallPacket<User[]>>(
+        procedureCall.select,
+        [],
+        (_err, procedureResult) => {
+          procedureResult.forEach((results) => {
+            if (isResultSetHeader(results)) {
+              console.log(results);
+
+              return;
+            }
+
+            // Strict checking the `RowDataPacket[]`
+            const user: User = results;
+            const id: number = user.id;
+
+            console.log(id);
+          });
+        }
+      );
+    });
+  });
+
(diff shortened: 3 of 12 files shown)


sidorares:
From the protocol point of view result of calling a procedure is indistinguishable from 2 statements ( where last results in ok packet, for example `select 1; set @a = 1;` ).

@wellwelwel can you help me understanding "Typing for stored procedure results wrong" problem? The way I see it currently there is nothing wrong and added `ProcedureCallPacket` is just a convenience for when you know returned shape is `[Resultset, OkPacket]`

wellwelwel:
Hey @sidorares, of course 🙋🏻‍♂️

> In https://github.com/sidorares/node-mysql2/issues/1934#issuecomment-1606546332, I explain my "*debugging*" process.

The `ProcedureCallPacket` just reproduces all the variations of current types from `execute` and `query`.

For example:

#### When I don't know what will be returned, I can use it as:
```ts
 /**
  * The `ProcedureCallPacket` will be:
  * `OkPacket | ResultSetHeader | RowDataPacket[] | RowDataPacket[][] | OkPacket[]`
  */
connection.execute<ProcedureCallPacket>('...');
```
- It would be like using `connection.execute('...');`

#### Or else
```ts
 connection.execute<ProcedureCallPacket<OkPacket>>('...');
 connection.execute<ProcedureCallPacket<OkPacket[]>>('...');
 connection.execute<ProcedureCallPacket<ResultSetHeader>>('...');
 connection.execute<ProcedureCallPacket<RowDataPacket[]>>('...');
 connection.execute<ProcedureCallPacket<RowDataPacket[][]>>('...');
```

---

In these examples I used the promise-based version, but the same idea applies for callbacks.

wellwelwel:
Important to say, I'm still testing the real possibilities of return and mapping them out before, to make sure that all these returns really happen.

> I plan to document (by commenting here) the final mapped results.

wellwelwel:
@sidorares, imagine this situation:
> Demonstrating in practice, this is the problem of Issue #1934

```ts

interface User extends mysql.RowDataPacket {
   id: number;
   name: string;
}

const conn = await mysql.createConnection({
   /* ... */
});

const sql = 'CALL showUsers()';
```

---

### 1) Realizing the error
> `RowDataPacket[]`
```ts
const [procedureResult] = await conn.query<User[]>(sql);

// should be: `User[]`, but it's `User` ❌
const Users = procedureResult[0];

// should be: `ResultSetHeader`, but it's `User` ❌
const { info } = procedureResult[1];

```

---

### 2) Trying for an alternative
> `RowDataPacket[][]`
```ts
import mysql, { RowDataPacket } from 'mysql2/promise';

interface User extends RowDataPacket {
   id: number;
   name: string;
}

const conn = await mysql.createConnection({
   /* ... */
});

const sql = 'CALL showUsers()';
const [procedureResult] = await conn.query<User[][]>(sql);

// now, it's `User[]` ✅
const Users = procedureResult[0];

// should be: `ResultSetHeader`, but it's `User[]` ❌
const { info } = procedureResult[1];

```

---

### From this PR
```ts
const [procedureResult] = await conn.query<mysql.ProcedureCallPacket<User[]>>(sql);

// `User[]` ✅
const Users = procedureResult[0];

// `ResultSetHeader` ✅
const { info } = procedureResult[1];

```

---

But, it has some conflicts when the **Procedure Call** performs queries that result in `OkPacket` (I'm investigating this).

wellwelwel:
> The way I see it currently there is nothing wrong and added `ProcedureCallPacket` is just a convenience for when you know returned shape is `[Resultset, OkPacket]`

I don't just see it as a convenience, but both a convenience and also a solution (thinking about the example situation).

- This can be convenient because I can indicate what it is and use the suggestion types easier.
- This can be a solution if I define the typings beforehand (as in the example).

Just another point, considering all existing variations, I don't see why we shouldn't have this one too, but I prefer to finish some tests before being able to provide any certainty.

wellwelwel:
@sidorares, I have completely changed the description of this **PR** after the tests conclusions.
Can you check it?

I also explained your questions and left my opinion about it.
Feel free to ask anything 🙋🏻‍♂️

wellwelwel:
For now, the most accurate typings based on debug process are:

```ts
declare type ProcedureCallPacket<
  T = RowDataPacket[] | RowDataPacket[][] | ResultSetHeader
> = T extends RowDataPacket[]
  ? [...T, ResultSetHeader]
  : T extends RowDataPacket[][]
  ? [[...T], ResultSetHeader]
  : T extends ResultSetHeader | OkPacket | OkPacket[]
  ? ResultSetHeader
  : never;
```

> *Still developing* 💭

It allows to use as:

```ts
ProcedureCallPacket;
ProcedureCallPacket<ResultSetHeader>;
ProcedureCallPacket<RowDataPacket[]>;
ProcedureCallPacket<RowDataPacket[][]>;
```

sidorares:
I feel like this is starting to be nuanced enough that without documentation almost no one will be benefitting from this feature.

I'm happy for the `ProcedureCallPacket` ( this PR ) to be merged, but maybe we need to prioritise "TS examples and best practises" page after

wellwelwel:
> I feel like this is starting to be nuanced enough that without documentation almost no one will be benefitting from this feature.

Yes, I agree.

---

> but maybe we need to prioritise "TS examples and best practises" page after

It can be done by documenting up to the moment before to this **PR**.

Then, this **PR** can be merged later, already together with **Procedure Call** examples documentation.

wellwelwel:
@sidorares, it's done 🙋🏻‍♂️

I skipped the #1934 from **TypeScript Milestone**  because I thought it was about a question/discussion.
Now it's possible to say that the open TypeScript Issues, for the most part, are about questions and discussions.

The next step will be to create a **TypeScript** documentation in `documentation/en/TypeScript-Examples.md`.

I remember the first time I used **TypeScript** with **mysql2** and didn't understand why every query was `any`. Finally, I noticed that it needs `npm i -D @types/node` to work 😅
I plan to start from here 🤹🏻‍♀️

PR #2114, opened 2023-07-06
Update events that are propegated from pool cluster to include remove

Michael03:
Removed events from promise wrapper that aren't fired by pool cluster.
I've seperated this into 2 commits. The first adds a failing test the second fixes that test

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -117,7 +117,7 @@ export interface PoolCluster extends EventEmitter {
 
   on(event: string, listener: (args: any[]) => void): this;
   on(event: 'remove', listener: (nodeId: number) => void): this;
-  on(event: 'connection', listener: (connection: PoolConnection) => void): this;
+  on(event: 'warn', listener: (err: Error) => void): this;
 }
 
 export function createConnection(connectionUri: string): Promise<Connection>;
--- a/promise.js
+++ b/promise.js
@@ -448,7 +448,7 @@ class PromisePoolCluster extends EventEmitter {
     super();
     this.poolCluster = poolCluster;
     this.Promise = thePromise || Promise;
-    inheritEvents(poolCluster, this, ['acquire', 'connection', 'enqueue', 'release']);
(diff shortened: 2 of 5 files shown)


promise.d.ts
wellwelwel, line 1: @Michael03, a question:

Should this change be the same in non promise-based `Pool Cluster`?

https://github.com/sidorares/node-mysql2/blob/facc68ce04947782997d9408c96736f9fe706497/typings/mysql/lib/PoolCluster.d.ts#L83

sidorares:
Thanks @Michael03 
I personally never used PoolCluster so maybe worth someone else to have a look but overall happy to merge.

One suggestion: could you move the affected test to under `test/builtin-runner/*` and rewrite it to use [describe/it](https://nodejs.org/api/test.html)?

sidorares:
actually that test file is way too big. Maybe worth factoring out PoolCluster test into separate file and add it to tests managed with a new runner

Michael03:
Ill get that refactored 👍

wellwelwel:
Looking into `lib/pool_cluster.js`, it contains only these events from `PoolCluster` class:

- `_increaseErrorCount`
https://github.com/sidorares/node-mysql2/blob/9051caf277990c110bb54cb4a8a23c497afb973c/lib/pool_cluster.js#L247

- `_getConnection`
https://github.com/sidorares/node-mysql2/blob/9051caf277990c110bb54cb4a8a23c497afb973c/lib/pool_cluster.js#L264

---

I think it's good, just pattern the change from `promise.d.ts` to `typings/mysql/lib/PoolCluster.d.ts` 🙋🏻‍♂️

```diff
- on(event: 'connection', listener: (connection: PoolConnection) => void): this;
+ on(event: 'warn', listener: (err: Error) => void): this;
```

https://github.com/sidorares/node-mysql2/blob/facc68ce04947782997d9408c96736f9fe706497/typings/mysql/lib/PoolCluster.d.ts#L83

wellwelwel (approved):

PR #2125, opened 2023-07-10
chore: deprecate `OkPacket` type and `changedRows` property

wellwelwel:
Just showing how this appears for final user:

---

Screenshot 2023-07-10 at 03 33 50

---

Screenshot 2023-07-10 at 03 32 43


wellwelwel:
@sidorares, I think it's good, what do you think?

In a major version, the `OkPacket` can be removed from `query` and `execute` types and the `changedRows` from `ResultSetHeader`.

sidorares:
lgtm

Issue #1983, opened 2023-04-26
Add a function to update pool default configurations


feeeeliipe:
Hello,

I'm working in a project that uses a MySQL instance running on AWS RDS and to improve security we're using [IAM based authentication](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html), which means that the password to create new connections with the database expires 15 minutes after its creation time and we need to generate a new one in order to create new connections.

When we use a simple connections, it works fine...

```
import { Signer } from "@aws-sdk/rds-signer";

// generate a new RDS token
const signer = new Signer();
const token = await signer.getAuthToken();

const connection = mysql.createConnection({
  host: 'aws.address...,
  user: 'iam_based_authentication_user',
  database: 'my_db',
});
```

The problem happens when we create a new pool using the RDS token. After 15 minutes (when the token is not valid anymore) we're not able to update pool configurations with a new valid token and that's preventing us to create new connections.

```
import { Signer } from "@aws-sdk/rds-signer";

// generate a new RDS token
const signer = new Signer();
const token = await signer.getAuthToken();

const pool = mysql.createPool({
  host: 'aws.address...,
  user: 'iam_based_authentication_user',
  database: 'my_db',
  password: token,
});

// Works fine
const connection = await pool.getConnection();

// Wait for 15 minutes and run it again...
// This code is going to fail, once the RDS token that is stored at the pool is not valid anymore.

// At this point I could generate a new RDS token, but I'm not able to update the password for pool's instance
const connection2 = await pool.getConnection();

```

Do you folks think that is possible to add a function to update the pool default configurations in order to change new connection attributes?

sidorares:
yes, I think we should allow to set connection config per connection on demand

There might be a workaround to your issue though. What auth plugin is used for connection? If its a (built in) cleartext plugin auth you might be able to override it to use dynamic password instead of the one passed to a pool config

See example in https://github.com/sidorares/node-mysql2/issues/1017#issuecomment-632492434

sidorares:
I'm not a big fan of polymorphic arguments but we could probably add to `createPool(config: PoolConfig)` another version
`createPool(createConfigOptionallyAsyncFunction)`, it'll get all config parameters at the time a new connection is spawned

feeeeliipe:
> yes, I think we should allow to set connection config per connection on demand
> 
[...]

Issue #1617, opened 2022-08-29
Do not enable mysql_clear_password by default


MasterOdin:
#1552 added support for the `mysql_clear_password` auth plugin, but made it enabled by default. Per the [manual docs on it](https://dev.mysql.com/doc/refman/5.7/en/cleartext-pluggable-authentication.html):

>> Sending passwords as cleartext may be a security problem in some configurations. To avoid problems if there is any possibility that the password would be intercepted, clients should connect to MySQL Server using a method that protects the password. Possibilities include SSL (see [Section 6.3, “Using Encrypted Connections”](https://dev.mysql.com/doc/refman/5.7/en/encrypted-connections.html)), IPsec, or a private network.
>
> To make inadvertent use of the mysql_clear_password plugin less likely, MySQL clients must explicitly enable it.

I would propose to make it disabled by default, and only enable it implicitly if the `LIBMYSQL_ENABLE_CLEARTEXT_PLUGIN` environment variable is set to `1`, `Y` or `y` (per the docs), or that a library consumer can add it themselves via `config.authPlugins` (following #1497 being merged to make it easy to reference).

MasterOdin:
With the v3.0.0-rc.1 release that includes the support for `mysql_clear_password`, I'd like to bump up this issue before a final v3.0.0 goes out such that doing this would then constitute a breaking change.

cc @sidorares

sidorares:
Looping in plugin users - @rathboma @silverbullettruck2001 - wdyt?

sidorares:
What I see the main attack vector is dns poisoning and mysql honeypot server stealing passwords.

What we discuss I guess is
- if server replies with "hey client, let's use cleartext auth" auth switch only agree if 1) `enableCleartextPlugin: true` is passed to a config or potentially `LIBMYSQL_ENABLE_CLEARTEXT_PLUGIN` is set
- if not explicitly enabled, throw connection error with message explaining how to enable

sounds reasonable to me

MasterOdin:
I agree with the above. Happy to put up a PR to tackle that, though would appreciate a pointer for (2) on where you'd like to see that in the code.

For anything dealing with databases, given how sensitive the data that might live in it, having the mechanisms to connect start as secure as possible and then only weaken security at the user's request is important to me.

sidorares:
> I agree with the above. Happy to put up a PR to tackle that, though would appreciate a pointer for (2) on where you'd like to see that in the code.

Probably here:

[...]

PR #2143, opened 2023-07-20
Allow to connect with arbitrary plugin

sidorares:
**WIP**

Currently mysql2 driver always tries to connect with mysql_native_password plugin. For the servers supporting PLUGIN_AUTH and configured to use plugins other than mysql_native_password initial connection is usually followed by AUTH_SWITCH_REQUEST packet and additional handshake with another plugin. This PR will allow to use plugins other than `mysql_native_password` in the initial client hello response. The order of preference on the plugin used during connection is:

  - if the server does not support plugin authorisation and secure auth server capability flag is set - mysql_native_password ( aka auth4.1 )
  - if the server does not support plugin authorisation and secure auth server capability flag is not set- pre auth4.1 token
  - plugin indicated by `defaultAuthenticationPlugin` config parameter name, if set ( error if the parameter is set but no standard or user provided plugin under this name )
  - plugin returned in `auth_plugin_name` [server hello packet field](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake_v10.html) ( error if plugin not configured )
  - `"mysql_native_password"` if no `defaultAuthenticationPlugin` or serverHello.auth_plugin_name

---------------

related issues: 
- https://github.com/sidorares/node-mysql2/issues/560
- https://github.com/sidorares/node-mysql2/issues/1617

- [ ] fix dead link https://mysqlserverteam.com/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/ to point to https://dev.mysql.com/blog-archive/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/
- [x] move standardAuthPlugins from auth_switch.js to connection. Make fields lazily loaded via getters. Also have a logic "given plugin name, return plugin instance" in the connection. Make sure it is possible to override standard plugins with custom ones
- [x] remove authToken calculation from handshakeResponse. Calculate it in the client_handshake command and pass the token to handshake_response packet
- [x] use plugin name from initial server hello packet to initialise plugin. Make it possible to override if defaultAuthenticationPlugin config option set ( mysql server uses default_authentication_plugin name in its config and sys variable - https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_default_authentication_plugin )
  Potentially allow defaultAuthenticationPlugin to be a function connection => Promise<string>, but also maybe better to have connection => Promise<connection config> instead to cover everything - see https://github.com/sidorares/node-mysql2/issues/1983#issuecomment-1524551716
- [x] initialize _authPlugin in the client_handshake command ( also - make sure it is removed on auth / auth switch / change user success )

- [ ] refactor change_user packet to use code from handshake_response packet

- [ ] ~~MAJOR VERSION: delete all references to authSwitchHandler and related code ( documentation/en/Authentication-Switch.md, auth_switch.js: warnLegacyAuthSwitch etc )~~ later
- [ ] tests for all 3 scenarios ( auth, auth switch, change user )
- [ ] tests for explicit default plugin name

### DIFF
--- a/lib/auth_41.js
+++ b/lib/auth_41.js
@@ -39,6 +39,7 @@ function sha1(msg, msg1, msg2) {
 
   return hash.digest();
 }
+exports.sha1 = sha1;
 
 function xor(a, b) {
   const result = Buffer.allocUnsafe(a.length);
--- a/lib/commands/client_handshake.js
+++ b/lib/commands/client_handshake.js
@@ -15,6 +15,9 @@ const Packets = require('../packets/index.js');
 const ClientConstants = require('../constants/client.js');
 const CharsetToEncoding = require('../constants/charset_encodings.js');
 const auth41 = require('../auth_41.js');
+const auth323 = require('../auth_323.js');
+const caching_sha2_password = require('../auth_plugins/caching_sha2_password.js');
+const native_password = require('../auth_plugins/mysql_native_password.js');
 
 function flagNames(flags) {
   const res = [];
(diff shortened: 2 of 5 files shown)


123FLO321:
Any update on this MR?
Currently ProxySQL does not support Auth Switch, making it fail to connect when not using mysql_native_password.
The possibility to specify the inital auth method, added in this mr, should solve this.

sidorares:
@wellwelwel I'm going to resurrect work on this, hope to make some progress in the coming days.
Main benefit is that it can dramatically improve performance of short lived connections as there is no need to re-negotiate `caching_sha2_password` method.

Since mysql v8 and above is widely adopted now, what happens for the majority of clients is this:

S: hey, I'm mysql v9.2. I support authentication plugins, prefer connection using caching_sha2_password. Here is my random data to start handshake
C: Hello, I'm going to connect using 'mysql_native_password'!
S: No, please use caching_sha2_password.  Here is my random data again to start handshake
C: Ok then ( continues caching_sha2_password sequence )

At least 1 full roundtrip can be easily avoided

dillen-roggensinger:
we're looking to upgrade our MySQL databases and would love to see this work completed. thanks for your contributions!

alumni:
Would be great to have an option that doesn't require implementing `authPlugins`. E.g.:
```json
{
  "defaultAuthenticationPlugin": "caching_sha2_password",
  "serverPublicKey": "xxxyyy",
  "overrideIsSecure": true
}
```

dannyreg:
@sidorares wondering if you need any assistance with this PR, I can possibly help and get this moved forward with whatever work you've done on it so far, or maybe just start a new PR? Either case lmk if you need some help :)

sidorares:
I'm going to postpone "Allow to connect with arbitrary plugin" since we already have "use initially auth method that server prefers". I'm not sure if there are real users that need to override that. We can resurrect this if there are

PR #2159, opened 2023-08-02
feat: add conn-level `infileStreamFactory` option

### LINKED ISSUE #2151: Handling `infileStreamFactory` at connection-level
The `infileStreamFactory` option only works when calling `connection.query()` or `connection.execute()`, which is [as documented](https://github.com/sidorares/node-mysql2/blob/master/documentation/en/Extras.md#sending-tabular-data-with-load-infile-and-local-stream).

However, if I set it on a connection-level [as suggested in #1080](https://github.com/sidorares/node-mysql2/issues/1080#issuecomment-588032567), the configuration is currently not propagated when calling `connection.query()` or `connection.execute()`. Since I am using Sequelize, which only exposes the connection configuration via `dialectOptions`, there is no other way I can set the `infileStreamFactory` option.

To help move this forward, I am keen to know where the change should be made:

1. To support the `infileStreamFactory` connection option in `mysql2` and pass it when calling `Commands.Execute()`/`Commands.Query()`; or
1. To make the change in Sequelize instead to expose these options at a query-level.

Any comments/feedback is deeply appreciated, thanks!

loozhengyuan:
This PR provides support for the `infileStreamFactory` option at the connection level. This will be useful for users who are looking to validate the file path at the connection/pool level. For downstream libraries like [sequelize](https://github.com/sequelize/sequelize) and [typeorm](https://github.com/typeorm/typeorm), exposing this option at the connection-level will also allow users to set `infileStreamFactory` option through the [`dialectOptions`](https://sequelize.org/docs/v6/other-topics/dialect-specific-things/#mysql) and [`extra`](https://typeorm.io/data-source-options#mysql--mariadb-data-source-options) parameters respectively.

With the change, the `infileStreamFactory` option can be supplied when instantiating a new connection:

```js
const fs = require("fs");

  user: 'test',
  database: 'test',
  infileStreamFactory: (path) => {
    // Validate file path
    const validPaths = ['/tmp/data.csv'];
    if (!validPaths.includes(path)) {
      throw new Error(`invalid file path: ${path}: expected to be one of ${validPaths.join(',')}`);
    }
    return fs.createReadStream(path);
  }
});

connection.query('LOAD DATA LOCAL INFILE "/tmp/data.csv" INTO TABLE test', onInserted);
```

Fixes #2151

### DIFF
--- a/documentation/en/Extras.md
+++ b/documentation/en/Extras.md
@@ -68,6 +68,28 @@ connection.query(
 );
 ```
 
+The `infileStreamFactory` option may also be set at a connection-level:
+
+```js
+const fs = require("fs");
+const mysql = require('mysql2');
+
+const connection = mysql.createConnection({
+  user: 'test',
+  database: 'test',
+  infileStreamFactory: path => {
+    // Validate file path
+    const validPaths = ['/tmp/data.csv'];
+    if (!validPaths.includes(path)) {
+      throw new Error(`invalid file path: ${path}: expected to be one of ${validPaths.join(',')}`);
+    }
+    return fs.createReadStream(path);
+  }
+});
+
+connection.query('LOAD DATA LOCAL INFILE "/tmp/data.csv" INTO TABLE test', onInserted);
+```
+
 ## Connecting using custom stream:
 
 ```js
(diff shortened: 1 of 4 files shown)


sidorares:
lgtm!

sidorares:
@loozhengyuan published as v3.6.0

PR #2131, opened 2023-07-13
Update Amazon RDS SSL CA cert

alexjurkiewicz:
Fixes #2130.

The new cert chain is significantly longer, there are 115 certs now. This reflects the fact there are four CA root types with per-region unique certificates.

PR #2163, opened 2023-08-07
Giltohar update ssl profiles with global pem 2024 plus

giltohar:


sidorares:
Hi @giltohar 
That would be a duplicate of #2131 , right? If not could you highlight how its different?

The reason I'm not merging #2131 yet is mainly size impact ( +130kb increase ) for a feature that only small subset of users need. I'm thinking of a possible solution to this

giltohar:
@sidorares my bad. i didn't see that PR. yes the pr sould be essentially duplicates. I was looking to update it with the newer global bundle for both regular AWS and AWS China

PR #2200, opened 2023-09-06
Fix EventEmitter on method sigatures to use spread syntax

KunZhou-at:
The listener parameter of `emitter.on` takes a variable list of parameters instead of an array. See https://nodejs.org/api/events.html#passing-arguments-and-this-to-listeners for an example.


wellwelwel:
Thanks @KunZhou-at 🙋🏻‍♂️

PR #2219, opened 2023-09-21
Add warnings count for non-DML/DLL queries like SELECT

KunZhou-at:
Turns out SELECT queries can also produce warnings, for example:
```
mysql> SELECT STR_TO_DATE("asdfasdfasdf", "%m-%d-%Y %H:%i:%s");
+--------------------------------------------------+
| STR_TO_DATE("asdfasdfasdf", "%m-%d-%Y %H:%i:%s") |
+--------------------------------------------------+
| NULL                                             |
+--------------------------------------------------+
1 row in set, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+-------------------------------------------------------------------+
| Level   | Code | Message                                                           |
+---------+------+-------------------------------------------------------------------+
| Warning | 1411 | Incorrect datetime value: 'asdfasdfasdf' for function str_to_date |
+---------+------+-------------------------------------------------------------------+
1 row in set (0.00 sec)
```

This change adds the warnings count from the second EOF packet for a SELECT query ([reference here](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query_response.html)) because I don't think it's possible for the first EOF packet, the one following `FIELDS`, to have warnings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
         "eslint-plugin-async-await": "0.0.0",
         "eslint-plugin-markdown": "^3.0.0",
         "husky": "^8.0.2",
-        "lint-staged": "^14.0.1",
+        "lint-staged": "^15.0.1",
         "portfinder": "^1.0.28",
         "prettier": "^3.0.0",
         "progress": "^2.0.3",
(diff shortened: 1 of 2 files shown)


wellwelwel:
@KunZhou-at, using this approach in practice, would it be something like `[rows, fields, warnings] = await conn.query(...)`?

KunZhou-at:
@wellwelwel Yeah, in terms of the API it might be a little confusing that DMLs have their warning counts as the first returned value but it's the last returned value for SELECTs.

wellwelwel:
Thanks, @KunZhou-at. In this case, I'm only focusing on typings 🙋🏻‍♂️

You've changed the `query` callback types:
```diff
- | ((err: QueryError | null, result: T, fields: FieldPacket[]) => any)
+ | ((err: QueryError | null, result: T, fields: FieldPacket[], warningCount: number | number[]) => any)
```

So I wonder if it can also be done in `execute`.

Also, for promises:
```diff
- Promise<[T, FieldPacket[]]>;
+ Promise<[T, FieldPacket[], warningCount: number | number[]]>;
```

PR #2220, opened 2023-09-22
Correcting typo from "paramaters" to "parameters"

atlc:


wellwelwel:
Thanks @atlc 🙋🏻‍♂️

Looking into code, this same typo occurs in:
https://github.com/sidorares/node-mysql2/blob/001a18370462ab6c531995d576ddc3d9ae4ef358/promise.d.ts#L27

Issue #2192, opened 2023-08-26
Problems with promise based pools after updating to 3.6.0


BauerPh:
I get this Error after updating to 3.6.0:

```js
/home/phx/phxAssistant/api/node_modules/mysql2/promise.js:356
    const localErr = new Error();
                     ^
Error: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined
```

it works fine with the previous version 3.5.2

Here is my call:

```js
const pool = require('mysql2')
  .createPool({
    host: process.env.DB_HOST,
    user: process.env.DB_USER,
    password: process.env.DB_PASS,
    database: process.env.DB_DATABASE,
    connectionLimit: 5,
    multipleStatements: true,
    dateStrings: true,
    typeCast: function (field, next) {
      if (field.type === 'TINY' && field.length == 1) {
        return field.string() === '1'; // true / false
      } else if (field.type === 'BLOB' && (field.length == 4294967295 || field.db === '')) {
        let value = field.string();
        if (/^-?\d+$/.test(value)) return value;
        try {
          let res = JSON.parse(value);
          return res;
        } catch (e) {
          return value;
        }
      }
      return next();
    },
  })
  .promise();

exports.query = async function (...args) {
  return (await pool.query(...args))[0];
};
```

sidorares:
Hi @BauerPh could you add some code to your example that can replicate the issue on a fresh database so I can debug it locally? Thanks

jgonera:
@sidorares I'm running into the same issue when using `mysql2` with [`sql-template-tag`](https://github.com/blakeembrey/sql-template-tag) package. `mysql2` 3.5.2 works, 3.6.1 results in the error mentioned above. Example of problematic code:

```ts
import mysql from 'mysql2/promise'
import sql from 'sql-template-tag'

const db = mysql.createPool(...)
db.query(sql`SELECT 1`)
```

If I do the following, 3.6.2 works though:

```ts
import mysql from 'mysql2/promise'
import sql from 'sql-template-tag'

const db = mysql.createPool(...)
const query = sql`SELECT 1`
db.query({ sql: query.sql, values: query.values })
```

sidorares:
@jgonera looks like a side effect of https://github.com/sidorares/node-mysql2/pull/2159/files#diff-f35455f30798a56117f979b6ad64a99cbc915318f08795935d68c80ab2223743R623-R626

[...]

PR #2238, opened 2023-10-11
fix: sql-template-strings compatibility

BauerPh:
this should fix issue #2192.

tested with Promise wrapper and Pools only

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -622,7 +622,8 @@ class Connection extends EventEmitter {
       // execute(options, cb)
       options = {
         ...options,
-        ...sql
+        ...sql,
+        sql: sql.sql
       };
       if (typeof values === 'function') {
         cb = values;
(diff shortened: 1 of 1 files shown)


BauerPh:
> thanks @BauerPh ! Could you check unit tests locally?

I'll try, but im not a pro xD

BauerPh:
okay: tests should pass now.
But i don't know if that is a good solution but it works now with the sql-template-tag & sql-template-strings library.

sidorares:
why do you need to detect that? Can this

```js
{
          ...options,
          sql: sql.sql,
          values: sql.values,
        };
```
work for both scenarios?

BauerPh:
yeah, it was stupid :D

Problem:

```JS
{
  ...options,
  sql: sql.sql,
  values: sql.values
};
```
omits other options passed through the `sql` parameter, like `namedPlaceholders` => tests fail.

```
options = {
  ...options,
  ...sql,
  sql: sql.sql
};
```
should finally do it.

sidorares:
lgtm, thanks @BauerPh !

sidorares:
should be released to npm as v3.6.2 in a moment. When you have time could you try with the version from npm and close original issue if everything works well for you @BauerPh ?

PR #2242, opened 2023-10-15
fix: remove stream.emit('close') on stream end

Ivan-Baranov:
On node v18+ stream.emit('close') produce Error: Premature close

    at new NodeError (node:internal/errors:405:5)
    at Readable.onclose (node:internal/streams/end-of-stream:154:30)
    at Readable.emit (node:events:514:28)

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -227,7 +227,7 @@ class Query extends Command {
   }
 
   /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }] */
-  row(packet, _connection) { 
+  row(packet, _connection) {
     if (packet.isEOF()) {
       const status = packet.eofStatusFlags();
       const moreResults = status & ServerStatus.SERVER_MORE_RESULTS_EXISTS;
@@ -279,7 +279,6 @@ class Query extends Command {
     });
     this.on('end', () => {
       stream.push(null); // pushing null, indicating EOF
-      stream.emit('close'); // notify readers that query has completed
     });
     this.on('fields', fields => {
       stream.emit('fields', fields); // replicate old emitter
@@ -302,7 +301,7 @@ class Query extends Command {
       Timers.clearTimeout(this.queryTimeout);
       this.queryTimeout = null;
     }
-    
+
     const err = new Error('Query inactivity timeout');
     err.errorno = 'PROTOCOL_SEQUENCE_TIMEOUT';
     err.code = 'PROTOCOL_SEQUENCE_TIMEOUT';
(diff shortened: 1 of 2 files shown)


wellwelwel:
Hi, @Ivan-Baranov 🙋🏻‍♂️

The CI failure comes from `stream/promises` that is missing on **Node 14**.

---

Instead:
```js
if (process.versions.node.split(".", 1)[0] <= 14) {
  stream.emit('close'); // notify readers that query has completed
}
```

An alternative could be to use `stream/promises` as callback, for example:

```js
const { finished } = require('stream');

const rowsStreamForAwait = [];
function testStreamForAwait(callback) {
  const stream = connection.execute('SELECT * FROM announcements').stream();
  stream.on('data', row => {
    rowsStreamForAwait.push(row);
  });
  stream.on('end', () => {
    callback(null);
  });
  stream.on('error', err => {
    callback(err);
  });
}
testStreamForAwait(err => {
  if (err) {
    throw err;
  }
});


const rowsStreamFinished = [];
function testStreamFinished(callback) {
  const stream = connection.execute('SELECT * FROM announcements').stream();
  stream.on('result', row => {
    rowsStreamFinished.push(row);
  });
  finished(stream, err => {
    callback(err);
  });
}
testStreamFinished(err => {
  if (err) {
    throw err;
  }
});

process.on('exit', () => {
  assert.deepEqual(rows, rowsStreamForAwait);
  assert.deepEqual(rows, rowsStreamFinished);
});
```

---

Also (just a small tip), if you're familiar with Docker, you can test locally any version of **MySQL** and **NodeJS** with a simple `docker-compose.yml` 🐳

```yml
version: '3.9'

services:
  mysql2:
    image: node:14
    working_dir: /usr/app
    volumes:
      - ./path-to-mysql2:/usr/app:ro
    command: npm run test
    environment:
      MYSQL_HOST: db
      use-compression: 0
      use-tls: 0
    depends_on:
      db:
        condition: service_healthy
  db:
    image: mysql:5.7
    platform: linux/amd64
    restart: always
    environment:
      MYSQL_DATABASE: 'test'
      MYSQL_ALLOW_EMPTY_PASSWORD: 'yes'
    healthcheck:
      test: ['CMD', 'mysqladmin', 'ping', '-h', 'localhost', '-u', 'root']
      interval: 10s
      timeout: 5s
      retries: 5
```

Then, run `docker compose up` and wait for the last log as "*exited with code 0*" 🧙🏻✨

Ivan-Baranov:
I want to show error in node 18+, then used `await finished(stream)` or `for await(let row of stream)`

Issue #2172, opened 2023-08-14
ERR_TLS_CERT_ALTNAME_INVALID error when using CNAME as host name for RDS connection


pawelusfr:
We have a CNAME set in Route53 for our RDS cluster. When using that CNAME as host for MySQL connection we are getting `ERR_TLS_CERT_ALTNAME_INVALID` error (Hostname/IP does not match certificate's altnames).

SSL option in configuration is set to `ssl: 'Amazon RDS'`.

This is because the certificate from RDS doesn't have the CNAME listed as altname and AFAIK there is no way to do that in RDS.

This was working fine in mysql2 3.5.0 and broke in 3.5.1 with #2119 and the switch to `Tls.connect`.

sidorares:
Are you able to connect if you set `rejectUnauthorized` to false?

Is the error coming from https://github.com/sidorares/node-mysql2/blob/48c2b8de9010ac18954c2c39ade410ba45dfa29d/lib/connection.js#L375-L377 or from https://github.com/sidorares/node-mysql2/blob/48c2b8de9010ac18954c2c39ade410ba45dfa29d/lib/connection.js#L385 ?

sidorares:
Maybe we need to add [SNICallback](https://nodejs.org/dist/latest-v18.x/docs/api/tls.html#tlscreateserveroptions-secureconnectionlistener) to handle your scenario

pawelusfr:
Yes I can connect with `rejectUnauthorized: false`.

The error is coming from https://github.com/sidorares/node-mysql2/blob/48c2b8de9010ac18954c2c39ade410ba45dfa29d/lib/connection.js#L385

One possible workaround would be allowing to pass custom `checkServerIdentity` function to `Tls.connect()` (I did a quick check and it will connect if that custom function allows it). This way clients could write their own identity handling for such cases (I believe that's what e.g. `pg` allows for Postgres).
Though it couldn't be used in conjunction with `ssl: 'Amazon RDS'` as one would have to pass an object to `ssl`. That means the client would need a string version of the Amazon certs. Maybe you could add an export of it from `mysql2` for such cases?

sidorares:
I'm actually thinking to deprecate `"Amazon RDS"` and move it to a separate package, something along the line

```js
import amazonRdsSslProfile from "amozon-rds-certs";  // package name TBD


const conn = mysql.createConnection({
   ssl: {
      ...amazonRdsSslProfile,
      checkServerIdentity: () => null,
   }
[...]

PR #2253, opened 2023-10-19
Fix server identity check

KunZhou-at:
I don't think we are respecting the documented behavior for verifyIdentity. According to the [documentation](https://nodejs.org/api/tls.html#tlsconnectoptions-callback), in order to skip server identity check, we need to override checkServerIdentity by passing in a function as an argument to `tls.connect`. Currently we do that in the callback by which time the server identity check has already happened and error bubbled up.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -365,19 +365,14 @@ class Connection extends EventEmitter {
       secureContext,
       isServer: false,
       socket: this.stream,
-      servername
-    }, () => {
-      secureEstablished = true;
-      if (rejectUnauthorized) {
-        if (typeof servername === 'string' && verifyIdentity) {
-          const cert = secureSocket.getPeerCertificate(true);
-          const serverIdentityCheckError = Tls.checkServerIdentity(servername, cert);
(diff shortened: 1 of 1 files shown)


mikewang333:
Looking for this behavior as well! I'm trying to set my own custom checkServerIdentity--any idea when this might get reviewed/merged? @wellwelwel 

Thanks!

mikewang333:
Also looking for this behavior--is there any timeline for when this might get reviewed/merged? @wellwelwel 

Thanks!

wellwelwel:
> when this might get reviewed/merged?

Hi @mikewang333, I think there's an ongoing discussion about it in #2172.
But in any case, these changes look good to me.

In addition, #2295 seems to do the same in another way when reviewing the modifications.

sidorares:
closing this as a duplicate of #2376 Sorry @KunZhou-at your PR didn't get enough attention

wellwelwel:
@KunZhou-at, I noticed you brought it back to #2594.

Could you explain the difference to PR #2376?

PR #2259, opened 2023-10-25
Fix error code for protocol error

KunZhou-at:
The error code emitted for `ER_CLIENT_INTERACTION_TIMEOUT` is "4031" instead of the expected "ER_CLIENT_INTERACTION_TIMEOUT", this fixes that.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -29,6 +29,7 @@ const Packets = require('./packets/index.js');
 const Commands = require('./commands/index.js');
 const ConnectionConfig = require('./connection_config.js');
 const CharsetToEncoding = require('./constants/charset_encodings.js');
+const ErrorCodeToName = require('./constants/errors.js');
 
 let _connectionId = 0;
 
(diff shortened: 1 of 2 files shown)


KunZhou-at:
cc @wellwelwel  bump on this

wellwelwel:
> cc @wellwelwel bump on this

@KunZhou-at, thank you for your many contributions 💙

This one in particular I couldn't figure out what the issue was and what the solution does. Could you please show a failure that occurs without your fix?

> _More personally_, when I see that no tests have been changed and a new test covers a failure, I usually merge the _PRs_. When I don't understand or I realize a decision would need to be taken, I usually leave it to @sidorares and also to the community 🤝

KunZhou-at:
I believe `error.code` is a number in the current code for `ER_CLIENT_INTERACTION_TIMEOUT` but we have always propagated the `error.code` to the client as a string elsewhere. So this seems to be inconsistent behavior.

KunZhou-at:
@wellwelwel Closing since it's not getting attention. Feel free to open and I can resolve the conflicts if that changes.

PR #2262, opened 2023-10-27
Limit client flags to server capabilities

### LINKED ISSUE #2261: Client flags not limited to server capabilities
For some unknown reason, Metabase fails to return a handshake when the server advertises CLIENT_CONNECT_ATTRS. Attempting to remove CLIENT_CONNECT_ATTRS from a `mysql2` server led me to confirm the following:

1. When a `mysql2` server doesn't set CLIENT_CONNECT_ATTRS, clients ( the `mysql` CLI and Metabase) still respond with CLIENT_CONNECT_ATTRS set.
2. Neither apparently include the client attributes section of the handshake packet, despite setting the flag.
3. `mysql2` accepts whatever flags the client sends in the handshake response.
4. `mysql2` reads off the end of the handshake buffer (`Should not reach here: undefined`) trying to read the client attributes. 

The TL;DR is that if I patch `mysql2` to OR the server capabilities and client flags (which I believe is what the protocol specifies), handshakes succeed in all cases. 

For reference, here are the handshake packets returned:
```
# Metabase
8aa3be0000000040e00000000000000000000000000000000000000000000000616e79001435406cb703f65194d07ed3104a196e1a36a2396e72756e514c006d7973716c5f6e61746976655f70617373776f726400

# mysql CLI
8da6ff1900000001ff0000000000000000000000000000000000000000000000616e790014e37b647b58a150e2202cd63b3ce7050d2c68715372756e716c006d7973716c5f6e61746976655f70617373776f726400
```

KrisBraun:
Fixes #2261

### DIFF
--- a/lib/commands/server_handshake.js
+++ b/lib/commands/server_handshake.js
@@ -35,7 +35,7 @@ class ServerHandshake extends Command {
 
   readClientReply(packet, connection) {
     // check auth here
-    const clientHelloReply = Packets.HandshakeResponse.fromPacket(packet);
+    const clientHelloReply = Packets.HandshakeResponse.fromPacket(packet, this.args.capabilityFlags);
     // TODO check we don't have something similar already
     connection.clientHelloReply = clientHelloReply;
     if (this.args.authCallback) {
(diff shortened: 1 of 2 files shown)


sidorares:
continuing as #4227

PR #2276, opened 2023-11-14
Fix big number accuracy

KunZhou-at:
Previously, big numbers were sometimes returned as number instead of string even if the number is larger than `Number.MAX_SAFE_INTEGER` or smaller than `Number.MIN_SAFE_INTEGER`. The behavior was conditional on the equality of `bigNumber.toString() === fullAccuracyNumberInString`. However, the number may not be at full accuracy: for many big numbers `BigInt(bigNumber).toString() !== bigNumber.toString()`, e.g., `99000001013466380`

Screenshot 2023-11-13 at 7 43 29 PM

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -222,10 +222,9 @@ class Packet {
         return word1 * 0x100000000 + word0;
       }
       res = new Long(word0, word1, !signed); // Long need unsigned
-      const resNumber = res.toNumber();
       const resString = res.toString();
-      res = resNumber.toString() === resString ? resNumber : resString;
-      return bigNumberStrings ? resString : res;
+      const safeResNumberPreferred = res.greaterThan(Number.MAX_SAFE_INTEGER) || res.lessThan(Number.MIN_SAFE_INTEGER) ? resString : res.toNumber();
+      return bigNumberStrings ? resString : safeResNumberPreferred;
     }
     // eslint-disable-next-line no-console
     console.trace();
@@ -425,7 +424,7 @@ class Packet {
     return StringParser.decode(
       this.buffer,
       encoding,
-      this.offset - len, 
+      this.offset - len,
       this.offset
     );
   }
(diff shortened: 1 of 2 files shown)


lib/packets/packet.js
```diff
@@ -450,21 +449,18 @@ class Packet {
...
     if (supportBigNumbers) {
       if (numDigits >= 15) {
         str = this.readString(end - this.offset, 'binary');
-        result = parseInt(str, 10);
-        if (result.toString() === str) {
-          return sign * result;
-        }
-        return sign === -1 ? `-${str}` : str;
-      }
-      if (numDigits > 16) {
```
KunZhou-at, line 465: dead code?

KunZhou-at:
cc @wellwelwel @sidorares this was a pretty critical bug that caused data corruption for us, would love to contribute the fix to upstream

wellwelwel:
@KunZhou-at, in fact the `Number.MAX_SAFE_INTEGER` is `9007199254740991`.

> this was a pretty critical bug that caused data corruption for us

I couldn't understand the issue, could you show an example where this would fail? It would be great if we could add a test to cover this possible failure as well 🙋🏻‍♂️

KunZhou-at:
@wellwelwel thanks for taking a look! Basically my change is from
 ```
res = resNumber.toString() === resString ? resNumber : resString;
```
to
```
res = res.greaterThan(Number.MAX_SAFE_INTEGER) || res.lessThan(Number.MIN_SAFE_INTEGER) ? resString : res.toNumber();
```
There are numbers larger than `Number.MAX_SAFE_INTEGER` such that `resNumber.toString() === resString`, and the example I have provided is `99000001013466380`. While the number's toString does print out the same number, it does not behave nicely with other native node facilities such as bigint. I think it's expected that the behavior of numbers outside of the safe range is undefined.  I don't think toString is the correct way to determine if a number is at full accuracy.

KunZhou-at:
@wellwelwel I still think this is a critical bug that can cause database corruption, but closing since it's not getting attention. Feel free to open and I can resolve the conflicts if that changes.

PR #2277, opened 2023-11-14
fix: add decodeuricomponent to parse uri encoded special characters in host, username, password and datbase keys

samarpanB:
This is PR to make changes as per last discussion here - https://github.com/sidorares/node-mysql2/pull/1621#issuecomment-1236397707

Also resolves https://github.com/sidorares/node-mysql2/issues/1384

Added decodeURIComponent for username, password, host and database. Currently, if mysql DB has password with special characters like '@', it isnt working at all it seems. After this change, this seems to be working. Also the discussion also clarifies that we should go ahead with this minimal change at least to keep it consistent with MySQL docs.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -260,11 +260,11 @@ class ConnectionConfig {
   static parseUrl(url) {
     const parsedUrl = new URL(url);
     const options = {
-      host: parsedUrl.hostname,
+      host: decodeURIComponent(parsedUrl.hostname),
       port: parseInt(parsedUrl.port, 10),
-      database: parsedUrl.pathname.slice(1),
-      user: parsedUrl.username,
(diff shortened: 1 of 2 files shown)


sidorares:
thanks a lot @samarpanB . Would you be able to add a unit test covering your change?

samarpanB:
> thanks a lot @samarpanB . Would you be able to add a unit test covering your change?

I actually updated the existing test case for url parsing test, which was failing anyways. Would you like me to add more test cases for this ?

samarpanB:
Added a few more test cases

samarpanB:
@sidorares Can you please review this now ?

samarpanB:
@sidorares Is it good to merge ?

Issue #1276, opened 2021-01-14
Malformed FieldPacket


mgummelt:
When fetching `FieldPacket`s from a query, I'm observing that `packet.type` is undefined, in violation of its type: https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/protocol/packets/FieldPacket.d.ts#L18

Example code:

```
$ const [res, fields] = await conn.execute<RowDataPacket[]>(sql);
$ fields[0].type
undefined
$ console.log(fields[0])
{ catalog: 'def',
    schema: 'companies',
    name: 'created_at',
    orgName: 'created_at',
    table: 'companies',
    orgTable: 'companies',
    characterSet: 63,
    columnLength: 19,
    columnType: 12,
    flags: 128,
    decimals: 0 }
```

In fact, `FieldPacket` seems to have a significantly different structure than its type indicates.

luckyyyyy:
same issue here

@sidorares

sidorares:
yeah, types are ported from mysqljs/mysql and looks like there are few differences to mysql2 `Field` shape. Probably the easiest way to make them somewhat compatible is to add aliases, for example type -> columnType

PR #2280, opened 2023-11-17
fix: Malformed FieldPacket

yutak23:
This ought to fix https://github.com/sidorares/node-mysql2/issues/1276.

I confirm object in v3.6.3.

```
{
  catalog: 'def',
  schema: 'for_test',
  name: 'id',
  orgName: 'id',
  table: 'hotels',
  orgTable: 'hotels',
  characterSet: 63,
  encoding: 'binary',
  columnLength: 10,
  type: 3,
  flags: [ 'NOT NULL', 'PRIMARY KEY', 'AUTO_INCREMENT' ],
  decimals: 0,
  typeName: 'LONG'
}
```

### DIFF
--- a/typings/mysql/lib/protocol/packets/FieldPacket.d.ts
+++ b/typings/mysql/lib/protocol/packets/FieldPacket.d.ts
@@ -16,7 +16,7 @@ declare interface FieldPacket {
   orgTable: string;
   protocol41?: boolean;
   table: string;
-  type: number;
+  type?: number;
   columnType?: number
   zerofill?: boolean;
   typeName?: string;
--- a/typings/mysql/lib/protocol/packets/FieldPacket.d.ts
+++ b/typings/mysql/lib/protocol/packets/FieldPacket.d.ts
@@ -3,19 +3,18 @@ declare interface FieldPacket {
     name: 'FieldPacket';
   };
   catalog: string;
-  charsetNr: number;
-  db: string;
+  schema: string;
+  characterSet: number;
   decimals: number;
-  default: any;
-  flags: number;
-  length: number;
+  flags: string[];
   name: string;
   orgName: string;
   orgTable: string;
-  protocol41: boolean;
   table: string;
   type: number;
-  zerofill: boolean;
+  typeName: string;
+  encoding: string;
+  columnLength: number;
 }
 
 export { FieldPacket };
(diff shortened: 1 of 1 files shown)


wellwelwel:
Thanks, @yutak23 🙋🏻‍♂️

---

Comparing the console output from #1276 with your outputs, there are some differences:
- `columnType`:
  - Only in **1276** output
- `flags`:
  - It's a number in **1276** (and in current types too)
  - It's a string array in your output
- `encoding`:
  - Only in your output
- `type`:
  - Only in your output
- `typeName`
  - Only in your output

---

Now, compared to the current types, you've removed:
```diff
- protocol41
- charsetNr
- db
- zerofill
```

---

I haven't looked into it in depth, but at first, I think the differences may be optional fields (perhaps it depends on the **MySQL** version or the type of query - `SELECT`, `INSERT`, etc.).

> Just thoughts 💭

wellwelwel:
What do you think about it:
> Turning the differences between outputs into optional fields and multiple types for `flags`.

```ts
declare interface FieldPacket {
  constructor: {
    name: 'FieldPacket';
  };
  catalog: string;
  charsetNr?: number;
  db?: string;
  decimals: number;
  default: any;
  flags: number | string[];
  length: number;
  name: string;
  orgName: string;
  orgTable: string;
  protocol41?: boolean;
  table: string;
  type?: number;
  zerofill?: boolean;
  columnType?: number;
  schema: string;
  characterSet: number;
  typeName?: string;
  encoding?: string;
  columnLength: number;
}

export { FieldPacket };
```

yutak23:
@wellwelwel , I too thought it was safer to make it optional for compatibility, as it certainly could be different for different versions of MySQL.

wellwelwel:
@yutak23, to close #1276, we can turn `type` into an optional field too and include: `columnType?: number` 🧑🏻‍🔧

yutak23:
@wellwelwel , oops sorry, fixed it.

wellwelwel:
@yutak23, it's okay 🙋🏻‍♂️

So, since the #1276 was opened justly because `type` property is missing (undefined `fields[0].type`), after turn it into an optional field we can merge this *PR* and close the issue #1276 🌞

yutak23:
@wellwelwel ,thank you for pointing this out again and again. Understood, I push the correction. I would appreciate it if you could check it.

wellwelwel:
Thanks again, @yutak23 😎✨

PR #2295, opened 2023-11-22
feat: add checkServerIdentity override

brandonjfeldkamp:
This fixes the issue described by https://github.com/sidorares/node-mysql2/pull/2119#issuecomment-1655735558 and partially resolves https://github.com/sidorares/node-mysql2/issues/2172. 

This doesn't implement the changes described in the issue regarding deprecating the Amazon RDS profile in favor of moving that out to a new package but it does allow for a path forward to re-enabling `rejectUnauthorized`.

The follow scenarios are now true:
- Set `rejectUnauthorized: false` - never any server identity verification
- Set `rejectUnauthorized: true` or not provided:
  
  - set `verifyIdentity: true` or not provided - run provided `checkServerIdentity()` or use default `tls.checkServerIdentity()` function

  - set `verifyIdentity: false` - `checkServerIdentity` set to no-op function

How was this tested? 

I tested changes locally against a tls-enabled postgresql instance. I'm not sure of how to incorporate any unit or integration tests for this as there does not currently appear to be any. Any support here would be lovely.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -353,8 +353,9 @@ class Connection extends EventEmitter {
       minVersion: this.config.ssl.minVersion,
       maxVersion: this.config.ssl.maxVersion
     });
-    const rejectUnauthorized = this.config.ssl.rejectUnauthorized;
-    const verifyIdentity = this.config.ssl.verifyIdentity;
+    const rejectUnauthorized = this.config.ssl.rejectUnauthorized === undefined ? true : this.config.ssl.rejectUnauthorized;
+    const verifyIdentity = this.config.ssl.verifyIdentity === undefined ? true : this.config.ssl.verifyIdentity;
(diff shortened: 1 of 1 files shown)


wellwelwel:
> Just linking
- #2253

brandonjfeldkamp:
> while the code around this change in your active memory - could you also please help reviewing #2295 ? That would be a minor version bump as it adds `checkServerIdentity` option. I don't like the fact that we would have 2 similarly named flags ( one is flag, one is function with custom check logic ), but might still worth adding

@juanheyns @sidorares regarding this concern of similarly named "things" from #2376, would you like me to rename it? I think this is still important as without it by setting the `verifyIdentity` flag to false you become more vulnerable to MITM attacks. By overriding the `checkServerIdentity` function we could enforce server identity in a way that makes sense for us. This is my line of thinking at least, does it make sense?

PR #2308, opened 2023-12-05
ci: improve OSX workflows

wellwelwel:
👇🏻

PR #2349, opened 2024-01-09
ci: refactor CI - OSX

wellwelwel:
> See this _PR_ also as a discussion based on ["osx is a low priority and intended to be a smoke test"](https://github.com/sidorares/node-mysql2/pull/2308#issuecomment-1842002931) comment.

It's been some time that **GitHub** has been failing `macOS-based` tests.

These false positives motivated two other _PRs_ (#2308 and #2348), but the problem isn't really with the **CI**.

---

Since the tests "against" the **MySQL Server** already take place in `CI - Linux`, I would like to propose that the test focus on **Node.js** versions for `macOS-based` tests (how it's already done in `CI - Windows`), reducing the chance of false positives.

---

Currently, I frequently try and retry the `CI - OSX` until it passes without false positives.

### DIFF
--- a/.github/workflows/ci-osx.yml
+++ b/.github/workflows/ci-osx.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [18.x, 20.x]
-        mysql-version: ["mysql:8.0.22", "mysql:8.0.33"]
+        mysql-version: ["mysql:8.0.33"]
         use-compression: [0, 1]
         use-tls: [0]
         mysql_connection_url_key: [""]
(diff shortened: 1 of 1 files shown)


wellwelwel:
> In this _PR_ with less jobs, it reported 2 errors until I tried again 🤔

Another suggestion could be use only the **Node.js LTS**:

```yml
node-version: [20.x]
mysql-version: ["mysql:8.0.33"]
use-compression: [0, 1]
use-tls: [0, 1]
```

sidorares:
👍

sidorares (approved):

Issue #2177, opened 2023-08-16
 Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames


Jooeee:
Im trying to connect to a Google Cloud Platform MySQL instance with SSL enabled using a public IP. However I get the error: 

> Error: Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames: IP: xx.xx.xxx.xx is not in the cert's list 

If I set reject unauthorised to false it works correctly.

There is no way to add the IP to the altnames because Google generates the certificates automatically. I also can't connect to the instance using a hostname because that isn't supported by Google. I've tried adding `checkServerIdentity: () => null` to the ssl options but get the same results. Is there any secure work around to this?

sidorares:
unfortunately `checkServerIdentity` can't be passed from the config yet, but I plan to add that

linking related issue: https://github.com/sidorares/node-mysql2/issues/2172

PR #2376, opened 2024-01-17
fix: add condition which allows code in callback to be reachable

### LINKED ISSUE #2375: Specifying rejectUnauthorized=true and verifyIdentity=false, unreachable code
When I specify rejectUnauthorized=true and verifyIdentity=false, I still get an error: 

`Hostname/IP does not match certificate's altnames: ...`

So it seems like `verifyIdentity` does not have any effect. Looking at the code it seems like lib/connection.js line 372 is unreachable:

https://github.com/sidorares/node-mysql2/blob/1d983fa46031a77c689faea5f69e6e0baa1b3de7/lib/connection.js#L372C19-L372C19

```
      const ca = readFileSync(path.join(__dirname, `/ssl-ca/amazon-root-ca-1.pem`)).toString() // RDS Proxy
      
      pool = await createConnection({
        host,
        user,
        password,
        database,
        port,
        ssl: {
          ca,
          rejectUnauthorized: true,
          verifyIdentity: false
        }
      })
```

juanheyns:
This PR addresses the problem reported in #2375 

Fixes #2375 

Relates to breaking changes introduced with #2119

Possible workaround for #2177
Relates to #2172

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -362,6 +362,9 @@ class Connection extends EventEmitter {
     const secureSocket = Tls.connect({
       rejectUnauthorized,
       requestCert: rejectUnauthorized,
+      checkServerIdentity: verifyIdentity
+        ? Tls.checkServerIdentity
+        : function() { return undefined; },
       secureContext,
       isServer: false,
       socket: this.stream,


juanheyns:
Manually tested against AWS RDS Proxy:

Screenshot 2024-01-16 at 8 56 53 PM
Screenshot 2024-01-16 at 8 56 21 PM

sidorares:
LGTM, thanks @juanheyns

sidorares:
@juanheyns would this change make #2253 unnecessary?

juanheyns:
> @juanheyns would this change make #2253 unnecessary?

Yes it would seem so.

sidorares:
@juanheyns published as v3.7.1, thanks!

while the code around this change in your active memory - could you also please help reviewing #2295 ? That would be a minor version bump as it adds `checkServerIdentity` option. I don't like the fact that we would have 2 similarly named flags ( one is flag, one is function with custom check logic ), but might still worth adding

Issue #1917, opened 2023-03-27
Node process can't finish correctly after creation of Connection Pool


komanton:
I noticed that with some configuration of Connection Pool the Node process can't stop working and just hang.
In my case, it appears in mocha tests when mocha process can't finish after all tests.
This is simple project to reproduce the situation:

package.json

```
{
  "name": "mysql2-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "mocha"
  },
  "dependencies": {
    "mocha": "^10.2.0",
    "mysql2": "^3.2.0"
  },
  "author": "",
  "license": "ISC"
}

```

/test/test.js

```
const mysql = require('mysql2');

describe('test connection pool', function () {

  const pool = mysql.createPool({
    host: 'localhost',
    port: 3007,
    user: 'dev',
    password: 'dev',
    database: 'dev',
    multipleStatements: true,
    connectionLimit: 16,
    maxIdle: 0,
    idleTimeout: 60000,
    enableKeepAlive: true,
    keepAliveInitialDelay: 30 * 60 * 1000, // CloudRun idle * 2
  });

  it('test stub', function () {
    // just place holder
  });

  this.afterAll(function () {
    console.log('Clean resources after all tests :');
  
    pool?.end();
  });  
});
```

Steps to reproduce:
1. Run command `npm test`.
2. All tests should be passed.

Actual behaviour: node process (i.e. mocha) is hang and can't finish.
Expected behaviour: **node process (i.e. mocha) should finish correctly**.

Note 1: Actually, I think this is very important issue because it may affect not only mocha process but also some processes in Docker environment and Lambda where processes should quickly start and finish to satisfy auto-scalabity algorithms.
Note 2: According to source code of mysql2, I noticed that in case of maxIdle < connectionLimit, mysql2 call `clearTimeout`. But looks like, it was forgotten to call `clearTimeout` on pool.end(). That's why 'event loop' is not empty, and it prevents a process from finishing.

sidorares:
Thanks for your research. Can confirm this issue, clearing idle timer on pool end seems to solve it. I'll work on a fix & unit test

jeremysf:
I'm running into this issue also when running tests with `maxIdle < connectionLimit`.

Any suggested workarounds to more cleanly shutdown?

anton-fjg:
As a work around, you can use mocha's `--exit` flag.

songpr:
[...]

PR #2384, opened 2024-01-18
fix: The removeIdleTimeoutConnectionsTimer did not clean up when the …

foxling:
…pool was closed.

 #1917

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -100,6 +100,7 @@ class Pool extends EventEmitter {
 
   end(cb) {
     this._closed = true;
+    clearTimeout(this._removeIdleTimeoutConnectionsTimer);
     if (typeof cb !== 'function') {
       cb = function(err) {
         if (err) {
--- a/test/integration/test-pool-release-idle-connection.js
+++ b/test/integration/test-pool-release-idle-connection.js
@@ -38,7 +38,6 @@ pool.getConnection((err1, connection1) => {
           connection4.release();
           connection4.destroy();
           pool.end();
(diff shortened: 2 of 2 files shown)


wellwelwel:
Thanks, @foxling 🙋🏻‍♂️

Do you think there is a way to test it in an unit test? Maybe we can use the steps from https://github.com/sidorares/node-mysql2/issues/1917#issue-1641776181.

sidorares:
happy to merge without a unit test ( but if we can add - great ). We need to detect leaking timer handles though, the test https://github.com/sidorares/node-mysql2/issues/1917#issue-1641776181 would pass OK but the process then won't exit for some time. As an option - spawn test in its own child process and expect it to finish within reasonable time

foxling:
[setTimeout(() => process.exit(0), 1000)](https://github.com/sidorares/node-mysql2/pull/2384/files#diff-3ee61ffd0d1ccbda2b1eaa017a4e2cd3e60b34141199a2ec75655e90315fd8f7L41)
In this unit test, which probably encountered this problem at that time, removing this code would fit the test scenario, so is that it?

wellwelwel:
> [setTimeout(() => process.exit(0), 1000)](https://github.com/sidorares/node-mysql2/pull/2384/files#diff-3ee61ffd0d1ccbda2b1eaa017a4e2cd3e60b34141199a2ec75655e90315fd8f7L41) In this unit test, which probably encountered this problem at that time, removing this code would fit the test scenario, so is that it?

@foxling, that makes sense to me.

@sidorares, can we merge it?

PR #2389, opened 2024-01-22
Fix node 18 (based on #711)

nofarham:


wellwelwel:
Perhaps we could include this test case here:

https://github.com/sidorares/node-mysql2/blob/e6a9306ed118501697936b8c80419d6502fa54a3/test/integration/connection/test-stream.js#L48-L63

nofarham:
@wellwelwel  can you add this code after the "connection.execute"?
```
    const s3 = connection.query('SELECT * FROM announcements').stream();
    for await (const row of s3) {
      rows3.push(row);
    }

process.on('exit', () => {
  assert.deepEqual(rows.length, 2);
...
  assert.deepEqual(rows, rows3);
});
```

wellwelwel:
@nofarham, testing it locally, everything works great 🚀

- **with your fix:**
  - `exited with code 0`
- **without your fix:**
  - `Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close`

---

**Just an observation:** this error occurs from **Node 18**, so it also occurs in **Node 19**, **20** and **21**.

nofarham:
@wellwelwel Thank you I have added the tests

wellwelwel:
> Thank you I have added the tests

@nofarham, you're welcome 🙋🏻‍♂️

All the main tests were passed, just fix an indentation conflict with the lint and we can merge it 🚀

wellwelwel:
Thanks again, @nofarham 🚀

nofarham:
@wellwelwel, when is the new version will be released?

wellwelwel:
> when is the new version will be released?

@nofarham, it's here ([3.8.0](https://github.com/sidorares/node-mysql2/releases/tag/v3.8.0)) 🙋🏻‍♂️

PR #2391, opened 2024-01-22
Perf: cache iconv decoder

sandinmyjoints:
While getting an iconv decoder isn't particularly slow, I noticed it happens a lot when decoding lots of varchar rows. From internal testing, with 5000 rows of a table containing two varchar utf8mb3 columns (length 100 and 500), I'm seeing about 12% speedup with this change.

### DIFF
--- a/lib/parsers/string.js
+++ b/lib/parsers/string.js
@@ -1,22 +1,34 @@
 'use strict';
 
 const Iconv = require('iconv-lite');
-const decoderCache = new Map();
+const LRU = require('lru-cache');
+
+const decoderCache = new LRU({
+  max: 500,
+});
 
 exports.decode = function (buffer, encoding, start, end, options) {
   if (Buffer.isEncoding(encoding)) {
     return buffer.toString(encoding, start, end);
   }
 
-  const decoderArgs = { encoding, options: options || {} };
-  const decoder =
-    decoderCache.get(decoderArgs) ||
-    decoderCache
-      .set(
-        decoderArgs,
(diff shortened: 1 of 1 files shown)


lib/parsers/string.js
```diff
@@ -1,21 +1,42 @@
...
+  max: 500,
+});
+
+exports.decode = function (buffer, encoding, start, end, options) {
   if (Buffer.isEncoding(encoding)) {
     return buffer.toString(encoding, start, end);
   }
 
-  const decoder = Iconv.getDecoder(encoding, options || {});
+  // Optimize for common case: encoding="short_string", options=undefined.
```
sandinmyjoints, line 15: I figure encoding is probably constant for most users. For my use case, it's always cesu8 (for utf8mb3).

sandinmyjoints:
It looks like the CI failures are a few timeouts waiting for MySQL to come up. Unrelated to changes.

wellwelwel:
> It looks like the CI failures are a few timeouts waiting for MySQL to come up. Unrelated to changes. 

Hi @sandinmyjoints, about the OSX tests, please ignore them for now 🙋🏻‍♂️

I run the "re-run" from time to time until the tests runs to its end.

sidorares:
lgtm, thanks @sandinmyjoints !

wellwelwel:
A suggestion would be to require LRU as `const LRU = require('lru-cache').default` to improve the compatibility with Next/Nuxt, etc.

What do you think, @sandinmyjoints? 🙋🏻‍♂️

---

For context:
- #2004

sandinmyjoints:
@wellwelwel: updated!

PR #2416, opened 2024-01-30
Fix premature close of the stream when it is paused

jansivans:
This pull request addresses a problem I have found with this pull request - [#2389](https://github.com/sidorares/node-mysql2/pull/2389).
When using streams with a pause (introduced by a timeout or similar asynchronous operation), the stream prematurely closes, leading to an `Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close`.
Here's a code snippet to reproduce the issue:

```javascript
const mysql = require('mysql2/promise');

(async () => {
    const connection = {
        user: '***',
        host: '***',
        database: '***',
        password: '***',
    };
    const dbConnection = mysql.createPool(connection);
    const query = dbConnection.pool.query(`select n from sequential_data limit 5`).stream();
    for await (const row of query) {
        await new Promise(r => setTimeout(r, 1000));
        console.log(row)
    }
    await dbConnection.end();
})();
```
I have implemented a fix that handles the stream's pause state more gracefully, preventing the premature close error. 
With my approach we no longer need to use `setImmediate` which creates a risk of emitting `close` before the stream had fully processed the `null` signal.

Here is output without timeout and with timeout:
CleanShot 2024-01-30 at 14 51 59@2x

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -30,7 +30,7 @@ class Query extends Command {
     this._receivedFieldsCount = 0;
     this._resultIndex = 0;
     this._localStream = null;
-    this._unpipeStream = function() {};
+    this._unpipeStream = function () { };
     this._streamFactory = options.infileStreamFactory;
     this._connection = null;
   }
@@ -155,7 +155,7 @@ class Query extends Command {
     const onPause = () => {
       this._localStream.pause();
     };
-    const onData = function(data) {
+    const onData = function (data) {
       const dataWithHeader = Buffer.allocUnsafe(data.length + 4);
       data.copy(dataWithHeader, 4);
       connection.writePacket(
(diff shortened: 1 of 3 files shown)


wellwelwel:
Hi @jansivans, thanks! I was able to reproduce your error using your example 🙋🏻‍♂️

What do you you think about add a test case for it?

I tested it locally and it worked using your fix:

`test/integration/connection/test-stream.js`

```diff
+ const rows4 = [];

// ...

+ const s4 = connection.query('SELECT * FROM announcements').stream();
+ for await (const row of s4) {
+   await new Promise(resolve => setTimeout(resolve, 1000));
+   rows4.push(row);
+ }

// ...

+ assert.deepEqual(rows, rows4);

```

jansivans:
@wellwelwel I have added your test to this PR.

wellwelwel:
> I have added your test to this PR.

Thanks again 🙋🏻‍♂️

@jansivans, before merging, could you "restore" the `package-lock.json`? 🧑🏻‍🔧

jansivans:
@wellwelwel done!

PR #2462, opened 2024-03-01
Refactor: Squashed into one place some duplicated typescript declaration + typo fix

nikelborm:


wellwelwel:
Thanks @nikelborm, I love it!

Just _one_ thing, instead of `ExecuteOrQueryResultPackets`, we can use just `QueryResult`.

nikelborm:
> Thanks @nikelborm, I love it!
> 
> Just _one_ thing, instead of `ExecuteOrQueryResultPackets`, can we just use `QueryResult`?

Ok, one sec

PR #2472, opened 2024-03-06
chore: improve coverage and good practices for tests

wellwelwel:
### CI

- Update or upgrade all CI Action versions, one by one
- Cache from **Linux**, **OSX** and **Website** are now individuals
- **Windows** cache now works
- Include **Bun** `latest` and migrate `ci-bun.yml` to `ci-linux.yml`
- The tests are now more conservative:
  - It test all version from **MySQL** in a specific CI (`ci-mysql.yml`)
  - It test all versions from Node from **14** (he's back) to **21**
  - Now, every test in all CI is tested in both **SSL** `0|1` and **Compression** `0|1`
- The coverage now uses only **c8**: `c8 ... npm run test` to generate the coverage data
  - There's no more _"if builtin-runner"_, all tests now work together 🐷
  - The coverage test now has its own CI
- Now both _tests_ and _typings_ are checked for **Lint** rules (both **Prettier** and **ESLint**)

---

### Tests

- Since we now use **Poku**, it doesn't matter if the tests are **ESM** or **CJS**, so we can run all the tests at once (and yes, it will work for all **Node.js** versions - and **Bun**) 🐷
- Now all tests are reorganized as `.test.`
- The `builtin-runner` has been moved to `esm`
- Now **Bun** also uses the same paths as **Node.js**, but filtering (_for now_) the basic `select` tests
- To maintain good practices, as we use `.esm` in **test/esm**, all tests in **test/unit** and **test/integration** are now `.cjs`
- Indirectly, using `.test.` should to remove the false positives warnings from **CodeQL** in tests 🧑🏻‍🔧
- All tests that were using **utest** have been migrated to ***ESM***
- Now there's a test to ensure the extensions in test paths

---

### Dependencies

- **Poku** has been added
- **utest** was removed
- **urun** was removed in #2423

---

### Some problems were found 🐞

- All the tests that were using **utest** weren't being processed:
  - While the test was waiting for a callback, the process was completed as "successfully" before
  - By turn these tests back, an Issue was found for `namedPlaceholders`
    - _I'll open a problem for that_


wellwelwel:
@sidorares, if do you like, we can upload the coverage results to [Codecov](https://about.codecov.io/).

- It will cover all _PRs_ and compare the actual coverage from `master` branch (for all contributors)
- We can use the badge in _README_

To do this, you'll need to create a secret called "CODECOV_TOKEN" and allows the Codecov App to this repository.

If you do, send me a ping and I'll adapt the coverage workflow 🙋🏻‍♂️

sidorares:
> To do this, you'll need to create a secret called CODECOV_TOKEN and allow the Codecov App to this repository.

sure, I'm ok to use codecov, I'll add the secret

do you know if codecov can merge multiple coverage results or do we need to do that ourselves? Different matrix runs skip some parts ( platform of feature flags specific ) so Ideally we want to consider the "union" of all matrix run coverages

wellwelwel:
> so Ideally we want to consider the "union" of all matrix run coverages

That's why I migrated the coverage CI to its own workflow, which is happening now 🙋🏻‍♂️

I don't know if it's clear, but using Poku, it runs all the tests (CJS and ESM) in a unique flow, so there's no need to merge multiple coverage results.

Also, there is no longer distinction between the previous `built-in`.
All tests run on all CI in all versions or platforms, without exception.

wellwelwel:
> do you know if codecov can merge multiple coverage results or do we need to do that ourselves?

Looking at the **Codecov** documentation, it says that **Codecov** will search for all "coverage" files available.

Alternatively, we can specify the files manually.

> In my previous comment, I forgot to consider the specified flags.

sidorares:
@wellwelwel `CODECOV_TOKEN` secret added

wellwelwel:
> `CODECOV_TOKEN` secret added

@sidorares, all done ☔️

---

> _And_...

By using **Poku** to fully test **MySQL2**, in addition to the prestige of it, it's an even greater incentive to continue investing in it. So, thank you again 🙋🏻‍♂️

PR #2486, opened 2024-03-12
ci: report coverage data to Codecov

wellwelwel:
There are some different behaviors as mentioned by @sidorares in https://github.com/sidorares/node-mysql2/pull/2472#issuecomment-1980620283:

- New and old MySQL Server versions
- SSL as `0` and `1`
- Compression as `0` and `1`

As the tests now run equally for all versions of **Node.js** and both platforms (**OS**), there's no need to send a report for each version of them.

That's why I opted to create an individual CI for the coverage report.

Naturally, to compare the coverages with the `master` branch, `CI - Coverage` will now be run for each push on the `master` too.

### DIFF
--- a/.github/workflows/ci-coverage.yml
+++ b/.github/workflows/ci-coverage.yml
@@ -3,7 +3,7 @@ name: CI - Coverage
 on:
   pull_request:
   push:
-    branches: [ main ]
+    branches: [ master ]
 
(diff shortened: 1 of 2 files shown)


wellwelwel:
Screen Recording 2024-03-12 at 01 49 19

Issue #2493, opened 2024-03-12
Pool not releasing connections


danj1980:
All of my queries include dbConnection.release(), however the connections do not get released and once connectionLimit is reached, the execution pauses waiting for an available connection.

I've re-worked my code so that every pool.getConnection() has a dbConnection.release() right next to it, to make absolutely certain that I'm not accidentally leaving any connections unreleased.

However, this did not help, and it seems that there is a bug in the releasing of pool connections.

I store my pool in a static variable with a class called Global.

```
class globals
{
	static dbPool = null;
}
```

Runs once on startup
```
globals.dbPool = await mysql.createPool({
				host: databaseServer,
				port: databasePort,
				user: databaseUsername,
				password: databasePassword,
				database: databaseName,
				connectionLimit: 10
			});

```
Example usage
            ```
            const dbConnection = await globals.dbPool.getConnection();

            let mysqlResponse = null;

            await dbConnection.query(`SELECT * FROM guild;`).then(result => {
                mysqlResponse = result[0];

            }).catch(err => console.log(err));

            dbConnection.release();
```
Result after some queries have been executed 
![image](https://github.com/sidorares/node-mysql2/assets/12071228/17c7aca3-567e-4202-8ebe-b202fcba6c5d)

danj1980:
![image](https://github.com/sidorares/node-mysql2/assets/12071228/aa83ce7b-a10a-42b8-a405-3433d43bb2b3)

danj1980:
It seems I've found the issue.

If maxIdle is not set on the pool, _removeIdleTimeoutConnections is never executed, and therefore there is never any cleanup of released connections.
I think it's this line that's the problem.
https://github.com/sidorares/node-mysql2/blob/b87ee4456556712cb8b0538f2f264d9a9fe5764a/lib/pool.js#L30

danj1980:
Side note, if I manually kill a connection on MySQL server, execution continues in the pool until the connections fill back up again.

danj1980:
> Hi @danj1980, 🙋🏻‍♂️
> 
> It's not directly related, but can you check if this _PR_ affects this behavior in some way?
> 
> * [fix: removal failure of freeConnection causing infinite loop #2171](https://github.com/sidorares/node-mysql2/pull/2171)
> 
I dont think that's the same issue, as the CPU usage is 0% when the connections run out.

> Also, this workaround:
> 
[...]

PR #2496, opened 2024-03-13
:bug: streaming reads should emit the dataset number for each dataset

fluffynuts:
When upgrading from `mysql` to `mysql2`, to support authentication updates on our mysql server(s), I found that my code around streaming, particularly when streams contain multiple resultsets, wasn't working properly because the resultset index wasn't being passed through as it is in `mysql`. This PR aims to resolve that.


wellwelwel:
Thanks @fluffynuts, 🙋🏻‍♂️

By including your test into  `test/integration` directory, it will be performed when performing the `npm run test`.

Then, it's not necessary to add the `test-me` in the **package.json** 🙋🏻‍♂️

---

To test a specific file, you can use the `FILTER` environment variable:

```bash
FILTER=test-multi-result-streaming npm run test
```

---

Also, to keep good practices, there's a small suggestion:

> Change from

```js
const { createConnection } = require('../common.test.cjs');
(async function() {
  'use strict';

  const { assert } = require('poku');

  // ...
```

> To 

```js
'use strict';

const { assert } = require('poku');
const { createConnection } = require('../common.test.cjs');

(async () => {
  // ...
```

fluffynuts:
thanks @wellwelwel 

I didn't know how to filter poku tests (this is the first time I've seen poku, tbh), so that was supposed to be a temporary npm script until I had a passing test (but it turned out I could run through webstorm anyway, so...) - anyways, it's removed and the requested changes have been made

wellwelwel:
The failed test doesn't seem to have relation to your PR 🤔

wellwelwel:
> @fluffynuts, sorry, I hadn't noticed something before:

You need to name your test file as `*.test.cjs` in the `test/integration/**` path 🙋🏻‍♂️

```diff
- test/integration/test-multi-result-streaming.cjs
+ test/integration/test-multi-result-streaming.test.cjs
```

wellwelwel:
@fluffynuts, for some reason the `test/integration/test-multi-result-streaming.test.cjs:54:12` brings a random result:

- Sometimes it will just work.
- Then, sometimes it just won't.

Can you check it?

---

You can also test it by using the native Node.js command:

```bash
node test/integration/test-multi-result-streaming.test.cjs
```

---

I don't think that it's related to this issue, but you don't need to try-catch the assert:
```diff
-  try {
    assert.equal(captured1.length, 1);
    assert.equal(captured2.length, 1);
    assert.deepEqual(captured1[0], compare1[0][0]);
    assert.deepEqual(captured2[0], compare2[0][0]);
-    process.exit(0);
-  } catch (e) {
-    console.error(e);
-    process.exit(1);
-  }
```

🙋🏻‍♂️

wellwelwel:
About the lint error, you can just run `npx prettier --write ./test` 🧙🏻

fluffynuts:
@wellwelwel could you point me at an example failure please? The failures I see are all due to pool tests (node 21, or latest mysql).

fluffynuts:
I'm wondering if an update between 8.0.33 and 8.0.35 fixed something. There was an SSL-related update in 8.0.34 which affects systems shipping their own SSL libraries (eg good linux distros):
image

wellwelwel:
I don't believe that it's related to a specific Node and MySQL version.
For example, this time the previous issue happened for Node.js 14, 16, also for MySQL 8.0.33 and SSL `0` and `1` (same for compression).

This really does seem to happen at random.

Apart from debugging, the only similarity is that errors only occur on Linux, not on Windows and OSX.

About the "_Warning: got packets..._", it's causing the error 🙋🏻‍♂️

The issue happens here:
```
assert.deepEqual(captured2[0], compare2[0][0]);
```

---

Also, you don't need to force the process exit code 🙋🏻‍♂️

```diff
- process.exit(0);
```

---

> I'd appreciate any help on this

I can try to do some debugging at the weekend, but in any case I've added a `help wanted` label 🔬

wellwelwel:
Hey @fluffynuts, there's a CI exclusive for MySQL versions:

https://github.com/sidorares/node-mysql2/blob/fdbed9188628b455cd0fbc48d1486b1d1161c474/.github/workflows/ci-mysql.yml#L21-L28

In that case, we already use the `mysql:latest` 🙋🏻‍♂️

fluffynuts:
I've now run this test around 400x with and without TLS on an unbuntu 22.04.4 machine with mysql 8.0.36 (just installed) - no errors. I'm going to try bumping the mysql version in the github action file.

> Hey @fluffynuts, there's a CI exclusive for MySQL versions:
> 
> https://github.com/sidorares/node-mysql2/blob/fdbed9188628b455cd0fbc48d1486b1d1161c474/.github/workflows/ci-mysql.yml#L21-L28
> 
> In that case, we already use the `mysql:latest` 🙋🏻‍♂️

cool, but that's not the fixture that's failing - linux-ci is failing on version 8.0.33. I've tried upping that to 8.0.36, and I get _more_ failures (node 14 as well!), but all of them have this junk in their output:

image

and I haven't seen this in about 1000x runs on an ubuntu vm in virtualbox.

wellwelwel:
You can try using a Docker container?

Usually, I use the `node:${version}-alpine` images, but for this case, I think using `node:${version}` is better to test it and ensuring the Ubuntu environment.

> If you don't know how to reproduce these tests on Docker, check the https://github.com/sidorares/node-mysql2/issues/2493#issuecomment-1992619188 🙋🏻‍♂️

---

As before, the error occurs on assert (`assert.deepEqual(captured2[0], compare2[0][0])`), these warnings aren't causing the CI failure.

---

Also, force the process exit code to `0` can hinder the debug process:

```diff
- process.exit(0);
```

fluffynuts:
after running the full suite in the same ubuntu vm, I am seeing the same "Warning: got packets out of order", but those tests don't fail, so that's odd.

I'll try to get docker running

> As before, the error occurs on assert (assert.deepEqual(captured2[0], compare2[0][0])), these warnings aren't causing the CI failure.
I'd agree that warnings (in themselves) don't cause test failures. I'd also postulate that if network comms to the mysql server are not running correctly, writing code that assumes that it can query data from the database without error may well be in error. I only saw these warnings on the ubuntu GHA hosts, until right now when I ran the full set of tests (but I only see those warnings on other tests which aren't failing). What I've read about that warning suggests that data can be lost during transmission when that occurs - which would explain the failure.

wellwelwel:
Can you revert the changes in the `.github/workflows/ci-linux.yml`?

wellwelwel:
> Can you revert the changes in the `.github/workflows/ci-linux.yml`?

Sorry about that, but I don't see an advantage to change MySQL version on Node / Bun CI.

It should fit more in `CI - MySQL`, but since we've already gotten the error and it's happening in different cases, I don't see why we should include that version in the matrix at this moment.

fluffynuts:
> > Can you please revert the changes in the `.github/workflows/ci-linux.yml`?
> 
> Sorry about that, but I don't see an advantage to change MySQL version on Node / Bun CI.
> 
> It should fit more in `CI - MySQL`, but since we've already gotten the error and it's happening in different cases, I don't see why we should include that version in the matrix at this moment.

similarly, there's no disadvantage; however I've reverted.

wellwelwel:
> similarly, there's no disadvantage; however I've reverted.

It's not about that @fluffynuts. The **CI** for Node & Bun is focused in test the versions of Node and Bun using a stable MySQL Server version along the time.

Then, the `CI MySQL` is about test different versions of MySQL Server using a stable version of Node.

You can see this PR for more details:

- #2472

fluffynuts:
> If you don't know how to reproduce these tests on Docker, check the [#2493 (comment)](https://github.com/sidorares/node-mysql2/issues/2493#issuecomment-1992619188) 🙋🏻‍♂️

I've attempted to do what is laid out in that issue. As I understand it, the point here is to isolate to a specific version of mysql, on a linux host (because that's what docker on windows will provide)?

Unfortunately, this is what I get, multiple attempts. I'm no docker guru, so I have no idea how to continue from here.

image

wellwelwel:
Look that:

```
    uncaughtException [Error: B81A0000:error:0A00041A:SSL routines:ssl3_read_bytes:tlsv1 alert decode error:c:\ws\deps\openssl\openssl\ssl\record\rec_layer_s3.c:1590:SSL alert number 50    
    ] {    
      library: 'SSL routines',    
      reason: 'tlsv1 alert decode error',    
      code: 'ERR_SSL_TLSV1_ALERT_DECODE_ERROR',    
      fatal: true    
    }    

  ✘ 108/115 node test\integration\test-multi-result-streaming.test.cjs 
```

---

By removing `process.exit(0)`, we now have more debugging details, including the fact that this time the failure also occurred on Windows and OSX.

fluffynuts:
> > ok, well then I have no idea how to continue with this - very obviously, adding support for dataset index has nothing to do with TLS. That's an environmental error, surely?
> 
> This is a great advantage of automated testing, ensuring that a change has no side effects 🧙🏻‍♂️✨

agreed, _if_ the change and error can be linked somehow - I'd be very interested to know how changing these three lines can affect SSL

image

right now, it just looks like my patch is exposing some existing condition 🤷

wellwelwel:
I'm not using my devices these days, so I can do some more limited things.
But I intend to debug this carefully.

I'm going to try out a few things, if there are any positive results, I'll bring them back here 🙋🏻‍♂️

fluffynuts:
@wellwelwel sorry to bother, just an update that we're running our fork in production on a fairly busy system and all seems ok, so I have confidence in my patch, but would like to know what's breaking here ):

wellwelwel:
@fluffynuts, sorry to keeping your waiting.

I noticed a simple issue from results order, I've also changed from `.destroy` to `.end`.
Additionally, I just ensured some **ESLint** rules were not being checked by **CI**.

Now, all the tests have been passed. Could you review it (#2628)?

> All your commits were preserved 🤝

wellwelwel:
@fluffynuts, thanks again for your work! 

I preferred not to modify your branch directly, but since all your commits were preserved in PR #2628, welcome to **MySQL2** contributors 🚀

> And again, apologies for the delay.

---

References from [**mysqljs**](https://github.com/mysqljs/mysql):

- https://github.com/mysqljs/mysql/blob/dc9c152a87ec51a1f647447268917243d2eab1fd/lib/protocol/sequences/Query.js#L210-L213
- https://github.com/mysqljs/mysql/blob/dc9c152a87ec51a1f647447268917243d2eab1fd/lib/protocol/sequences/Query.js#L223-L225

PR #2506, opened 2024-03-16
ci: ensure coverage limits

wellwelwel:
@sidorares, if you _want_ to allow the **Codecov** app to this repository, these conditions (configurable) will now be considered **CI** failures:

- ❌ More than **2%** loss of coverage compared to the base
- ❌ Less than **90%** of total coverage

> Actual coverage is **90.25%**.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 90%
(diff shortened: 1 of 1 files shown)


sidorares:
👍 
lets see if its too strict, we can adjust later if we find current limits counter productive

PR #2515, opened 2024-03-20
Comment added.

Kriishnaa:


sidorares:
thanks @Kriishnaa but I don't think the comments add any valye as of now. If you want to contribute "getting staring development" documentation as a readme or wiki page that would be greatly appreciated but trivial comments like in you PR while seem useful initially often can quickly get out of sync with reality

PR #2518, opened 2024-03-21
Use the same regex as mysqljs for changed rows

KunZhou-at:
For `INSERT ... ON DUPLICATE KEY UPDATE` queries, the changed rows the EOF packet is named Duplicates instead of Changed. Here is the a sample message:
```
Records: 4  Duplicates: 0  Warnings: 0
```
I propose using the regex from the mysqljs project which also handles this situation.


wellwelwel:
Thanks, @KunZhou-at 🚀

Can you include a test for these variations?

Issue #2541, opened 2024-03-29
Updated Amazon RDS SSL CA cert chain is a single string


BenjaminBruenau:
The latest changes released in `3.9.3` seem to lead to a lot of issues e.g. SSL cert errors or even the problem I encountered in https://github.com/vercel/nft/issues/402

I guess this is due to https://github.com/sidorares/node-mysql2/pull/2131 which lead to the cert-chain being a single big string instead of separate array elements.

It is possible to work around this for now by downgrading back to `3.9.2`

PR #2542, opened 2024-03-29
fix(SSL): separate each certificate into an individual item

BenjaminBruenau:
This fixes the issue described in https://github.com/sidorares/node-mysql2/issues/2541.

There seems to have been a mistake while updating the cert-chain in https://github.com/sidorares/node-mysql2/pull/2131 , which was released in `3.9.3`.
The cert-chain was changed into a single string where before each of the certificates were separate array elements.

### DIFF
--- a/lib/constants/ssl_profiles.js
+++ b/lib/constants/ssl_profiles.js
@@ -29,7 +29,7 @@ exports['Amazon RDS'] = {
       'wQBBvkxQ71oOmAG0AwaGD0ORGUfbYry9Dz4a4IcUsZyRWRMADixgrFv6VuETp26s\n' +
       '/+z+iqNaGWlELBKh3iQCT6Y/1UnkPLO42bxrCSyOvshdkYN58Q2gMTE1SVTqyo8G\n' +
       'Lw8lLAz9bnvUSgHzB3jRrSx6ggF/WRMRYlR++y6LXP4SAsSAaC0=\n' +
-      '-----END CERTIFICATE-----\n' +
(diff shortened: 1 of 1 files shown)


wellwelwel:
Thanks, @BenjaminBruenau!

I can't review it at this time, but based on recent reports, it looks good to me.

BMO-tech:
Thanks for the update, I can test in on Monday when I'm back at work

sidorares:
If someone from AWS reads this - would be great to have small RDS instance so that we can have SSL smoke test in the CI ( if you not from AWS but know who to tag - please do so 😝 )

sidorares:
And thanks a lot for the fix @BenjaminBruenau . I think we can wait for extra confirmation from @BMO-tech before merging

ennioVisco:
> And thanks a lot for the fix @BenjaminBruenau . I think we can wait for extra confirmation from @BMO-tech before merging

As mentioned here, https://github.com/sidorares/node-mysql2/pull/2131#issuecomment-2028513908, in our case it seems to brake the node modules import for us, still investigating how it is related. If we have updates, we'll report here.

wellwelwel:
It's a really basic review: since there are 115 certs, this _PR_ ensures 115 array items.

I'll merge it, since it's currently a single string for 115 certificates (potentially causing tokenization limit issues - https://github.com/vercel/nft/issues/402), but we'd appreciate to have an extra check (in a real **RDS**).

Thanks again, @BenjaminBruenau 🙋🏻‍♂️

> cc: @BMO-tech, @pedrovanzella

PR #2572, opened 2024-04-09
fix(security): improve sanitization for big numbers support

wellwelwel:
This _PR_ continues the #2424 work, ensuring the type of parameters received in the connection and query options for both text and binary parsers.


wellwelwel:
@sidorares, as these changes are simple, I'll merge it to release in #2566 (`3.9.4`) and follow the same way from #2529 (manually edit the _Changelog.md_ to preserve the commit description).

PR #2574, opened 2024-04-09
fix(security): improve results object creation

wellwelwel:
Finishing the tasks started in #2424 (and #2572).

This _PR_ ensures a clean object evaluation for every item in results.

This is the final evaluation for both text and binary parsers:

```js
const result = Object.create(null);

Object.defineProperty(result, 'constructor', {
  value: Object.create(null),
  writable: false,
  configurable: false,
  enumerable: false
});

// ...

// nestTables
if (!result[${tableName}]) result[${tableName}] = Object.create(null);
```

> I also tried a simple `Object.freeze(result)`, but the unit tests (specific to this) didn't pass as expected 🔓

---

### Note

These changes preserve security without affecting the end user:

```js
const [results] = await connection.query('SELECT 1+1 AS `test`'); // or execute

results[0].test.toFixed(2); // '2.00' ✅
results[0].addCustomProp = true; // true ✅
```

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -5,7 +5,7 @@
   "reporter": ["text", "lcov", "cobertura"],
   "statements": 88,
   "branches": 84,
-  "functions": 78,
+  "functions": 77,
   "lines": 88,
   "checkCoverage": true,
   "clean": true
--- a/lib/parsers/binary_parser.js
+++ b/lib/parsers/binary_parser.js
@@ -122,7 +122,13 @@ function compile(fields, options, config) {
   if (options.rowsAsArray) {
     parserFn(`const result = new Array(${fields.length});`);
   } else {
-    parserFn('const result = {};');
+    parserFn('const result = Object.create(null);');
+    parserFn(`Object.defineProperty(result, "constructor", {
+      value: Object.create(null),
+      writable: false,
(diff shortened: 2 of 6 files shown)


wellwelwel:
@sidorares, as done in #2572, I'll merge and attribute the credits manually.

Also, I've sent you some more detailed information by e-mail 🙋🏻‍♂️

ert78gb:
Hi,

Sorry for commenting on a closed PR, but it introduced a breaking changes for me.
I use node `assert/strict` library for assertion and it recognise the `constructor` own property of the "mysql object" as a new property. I construct the expected object as object literals, that does not contains the `null prototype constructor`

I created a reproduction repo https://github.com/ert78gb/mysql2-object-creation

I would like to know what is your opinion about this "breaking changes". Do you think `node:assert` library should change the comparison algorithm or you would like to modify the `mysql2` lib.

If you would like I could open a new issue.

Thx,
Robi

wellwelwel:
Hm... 🤔

@ert78gb, I can't answer yet whether it's an unintentional _breaking change_ for deep tests comparisons or if it can be fixed. I need to elaborate some tests.

Thanks for the basic repro, I'll start with that.

Issue #2585, opened 2024-04-10
Release 3.9.4 breaks code that depended on .hasOwnProperty()


isdampe:
Pull #2574 looks like it introduced breaking changes that set the `Object` prototype to null (and not `Object.prototype`). A consequence of this is that returned records can no longer call methods of `Object.prototype` such as `hasOwnProperty`.

sidorares:
@wellwelwel maybe freeze Object.prototype instead of inheriting from null?

https://cheatsheetseries.owasp.org/cheatsheets/Prototype_Pollution_Prevention_Cheat_Sheet.html
https://portswigger.net/web-security/prototype-pollution/preventing

@isdampe the "correct" way of calling `hasOwnProperty` should account for a possible null-prototype https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#objects_created_with_object.createnull but you are right, this should not be introduced in a minor version change

We either need to fix this error somehow in 3.x or bump major version ( and fix in 3.x )

wellwelwel:
@sidorares, I'm reviewing the fundamentals. The security tests didn't pass locally because I tried to do it based on the `prototype` property of an object. For example:

Object:

```js
const test = {};
// test.prototype is undefined;

Object.freeze(test.__proto__); // ✅
```

Class:

```js
const test = class {};
// test.__proto__ is undefined;

Object.freeze(test.prototype); // ✅
```

wellwelwel:
I don't think we'll need a major version bump for this, but I'll perform some tests and check again to be sure.

Related: https://github.com/sidorares/node-mysql2/pull/2574#issuecomment-2047118083

wellwelwel:
@sidorares, we have some problems about [**generate-function**](https://github.com/mafintosh/generate-function) 🔓

---

While trying to solve this issue, I found a vulnerability on [**generate-function**](https://github.com/mafintosh/generate-function) that blocks me to fix it.

Normally, this should work, but it's not possible to do this using `parserFn` due to that vulnerability:

```js
const obj = {};
Object.freeze(obj.__proto__);
```

---

[**generate-function**](https://github.com/mafintosh/generate-function) has ~**1.7 million** dependent public repositories, ~**96** direct public dependencies (npmjs) and over **3 million** downloads per week, so I intend to report it to the creator.

[...]

PR #2591, opened 2024-04-12
fix: revert breaking change in results creation

wellwelwel:
The previous solution (#2574) aimed to create a clean object, but it caused a major change.

Instead, I just grouped the object's native properties:

```js
const privateObjectProps = new Set([
  '__defineGetter__',
  '__defineSetter__',
  '__lookupGetter__',
  '__lookupSetter__',
  '__proto__',
]);
```

Then, it will perform a simple validation to ensure that the `fields[i].name` is safe:

```js
if (helpers.privateObjectProps.has(fields[i].name)) {
  throw new Error(
    `The field name (${fieldName}) can't be the same as an object's private property.`,
  );
}
```

By this way, it's possible to:

- use the native methods available from an object
- deep compare results with an expected variable
- test the vulnerability without use a query for that
- If an user, for some reason, wants to create or customize a `prototype` after the query has been finished, they can

> Basically, as it should be 🙋🏻‍♂️


wellwelwel:
@ert78gb, I included a simple test to ensure your repro:

```js
const expected = [
      {
        test: 2,
      },
    ];

const [results] = await connection.query('SELECT 1+1 AS `test`');

assert.deepStrictEqual(results, expected, 'Ensure clean object "results"');
```

sidorares:
Let's run some benchmarks before merging

wellwelwel:
Particularly, I don't like this solution, even if the benchmark is good (I think it's too much).

Even using freeze, this would still be a major change, as we would be blocking the user if they later wanted to use a prototype of their own.

---

> I believe this could be much simpler 💡

Instead of creating a complex object, we can just check if the `field` is not on a blacklist, without even needing to `freeze`. For example:

```js
const nativeProps = ['__proto__', 'hasOwnProperty', 'toString', 'valueOf'];

if (nativeProps.includes(field)) {
  // error
}
```

Now, we keep it safe, we don't cause any breaks, and we keep the solution simple and the user can use, change and create any prototype they need.

> For now, I'm just thinking about it, but I believe it could be a definitive solution.

wellwelwel:
@sidorares, I rewrote the logic and description of the _PR_.

Could you check again? 🙋🏻‍♂️

wellwelwel:
Testing all the standard object properties, considering **Node.js** long term versions:

```js
const emptyObject = {};
const proto = Object.getPrototypeOf(emptyObject);
const nativeObjectProps = Object.getOwnPropertyNames(proto);

const blockedFields = nativeObjectProps.map((prop) => [{ name: prop }]);

blockedFields.forEach((fields) => {
  try {
    getBinaryParser(fields, {}, {});
    assert.fail('An error were expected');
  } catch (error) {
    assert.strictEqual(
      error.message,
      `The field name (${srcEscape(fields[0].name)}) cannot be the same as the property of a native object.`,
      `Ensure safe ${fields[0].name}`,
    );
  }
});
```

- Supposing that someday **Node.js** includes some new method/property, this test will fail 🔐
- Same idea to `TextRowParser`.

wellwelwel:
> Using `Set` and `has`, instead of `Array` and `includes` due to:

Basic benchmark:

```js
const benchmarkArrayIncludes = (field) => {
  const nativeObjectProps = [
    '__defineGetter__',
    '__defineSetter__',
    '__lookupGetter__',
    '__lookupSetter__',
    '__proto__',
    'constructor',
    'hasOwnProperty',
    'isPrototypeOf',
    'propertyIsEnumerable',
    'toLocaleString',
    'toString',
    'valueOf',
  ];

  console.time('Array.includes');
  for (let i = 0; i < 1000000; i++) nativeObjectProps.includes(field);
  console.timeEnd('Array.includes');
};

const benchmarkSetHas = (field) => {
  const nativeObjectProps = new Set([
    '__defineGetter__',
    '__defineSetter__',
    '__lookupGetter__',
    '__lookupSetter__',
    '__proto__',
    'constructor',
    'hasOwnProperty',
    'isPrototypeOf',
    'propertyIsEnumerable',
    'toLocaleString',
    'toString',
    'valueOf',
  ]);

  console.time('Set.has');
  for (let i = 0; i < 1000000; i++) nativeObjectProps.has(field);
  console.timeEnd('Set.has');
};

for (let i = 0; i < 5; i++) benchmarkArrayIncludes('field');
for (let i = 0; i < 5; i++) benchmarkSetHas('field');
```

---

Results:

```
Array.includes: 1.891ms
Array.includes: 0.586ms
Array.includes: 0.593ms
Array.includes: 0.59ms
Array.includes: 0.646ms
Set.has: 1.228ms
Set.has: 0.563ms
Set.has: 0.563ms
Set.has: 0.561ms
Set.has: 0.559ms
```

---

Compatibility:

`Set`: [Node.js 0.12.18](https://node.green/#ES2015-built-ins-Set).

wellwelwel:
Bringing a comment here: https://github.com/sidorares/node-mysql2/commit/74abf9ef94d76114d9a09415e28b496522a94805#r140992502 🤝

wellwelwel:
> As suggested by @sidorares:

Instead of checking the field name manually, we have some alternatives:

- Refactor the `results` object to a mapped object
- Keep it as `Object.create(null)` by adding a config option to use a standard object for results (not sure about bump a major version for this)

> Although the issue #2585 doesn't have that many interactions, I think it's important to consider the users who don't participate directly 🙋🏻‍♂️

sidorares (approved):

wellwelwel:
@sidorares, can we merge this? 🙋🏻‍♂️

wellwelwel:
Merging this.

For a major bump version, I recommend reverting this PR and include a safer approach in **docs** such as:

```js
const obj = Object.create(null);

obj.test = true;

Object.hasOwn(obj, 'test'); // ✅
```

PR #2601, opened 2024-04-18
fix: binary parser sometimes reads out of packet bounds when results …

### LINKED ISSUE #2602: typeCast: false causes out of bounds reads to packet when result set contains NULL values and binary protocol is used.
I noticed this when debugging a separate issue relating to out of bounds reads in the binary parser.

Presently the binary parser does not check the NULL mask when the MySQL connection is configured with `typeCast: false` and will always attempt to `readLengthCodedBuffer` which causes the results to be nonsensical or the whole connection to panic. In my testing if there were less than three NULL values in the result row it just caused mangling, but after the fourth the connection state was corrupted.

[I've created a PR](https://github.com/sidorares/node-mysql2/pull/2601) that checks `nullBitmaskByte` before attempting to `readLengthCodedBuffer()` to go along with this issue. When the cell is null the parser will push `NULL` to the results. I considered doing an empty buffer, but thought that might be ambiguous since length 0 strings and empty blob fields would also be empty buffers, I think. I don't use `typeCast: false` in a meaningful capacity though so I am not attached to either approach.

### Test Table
```mysql
Create Table: CREATE TABLE `binparse` (
  `id` int NOT NULL AUTO_INCREMENT,
  `n1` varchar(32) DEFAULT NULL,
  `n2` varchar(32) DEFAULT NULL,
  `n3` varchar(32) DEFAULT NULL,
  `n4` varchar(32) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
```
```
+----+------+------+------+------+
| id | n1   | n2   | n3   | n4   |
+----+------+------+------+------+
|  1 | NULL | NULL | NULL | NULL |
|  2 | NULL | NULL | NULL | NULL |
|  3 | NULL | NULL | NULL | NULL |
+----+------+------+------+------+
```

### Three Nulls
```javascript
const connection = await mysql.createConnection({
	debug: true,
	typeCast: false
        ...
});
connection.execute(
	'SELECT n1,n2,n3 FROM binparse',
	[],
	(error, result, fields) => {
		console.log(result);
	}
);
```
```
0 20154 ==> Execute#row(9,EOF,9)
[
  [Object: null prototype] {
    n1: <Buffer 00 00>,
    n2: <Buffer 00 1c 02 00 00 08 00>,
    n3: <Buffer 05 00 00 09 fe 00 00 22 00>                                                                                                                                                                                                                                    },
  [Object: null prototype] {                                                                                                                                                                                                                                                     n1: <Buffer 00 00>,
    n2: <Buffer 00 1c 05 00 00 09 fe 00>,
    n3: <Buffer >
  },
  [Object: null prototype] {
    n1: <Buffer 00 00 09 fe 00>,
    n2: <Buffer >,
    n3: <Buffer 00>
  }
]
```


### Four Nulls
```javascript
const connection = await mysql.createConnection({
	debug: true,
	typeCast: false
        ...
});
connection.execute(
	'SELECT n1,n2,n3,n4 FROM binparse',
	[],
	(error, result, fields) => {
		console.log(result);
	}
);
```
```
Error: Should not reach here: undefined
    at Packet.readLengthCodedNumberExt (/home/wyatt/Repos/2873-13-vw-cup-lxp-api/node_modules/mysql2/lib/packets/packet.js:232:11)
    at Packet.readLengthCodedNumber (/home/wyatt/Repos/2873-13-vw-cup-lxp-api/node_modules/mysql2/lib/packets/packet.js:193:17)                                                                                                                                                  at Packet.readLengthCodedBuffer (/home/wyatt/Repos/2873-13-vw-cup-lxp-api/node_modules/mysql2/lib/packets/packet.js:398:22)
    ...
0 20130 ==> (no command)#(no command)(9,maybeOK,6)
node:events:491                                                                                                                                                                                                                                                                    throw er; // Unhandled 'error' event
      ^
...
Error: Unexpected packet while no commands in the queue
    at Connection.protocolError (/home/wyatt/Repos/2873-13-vw-cup-lxp-api/node_modules/mysql2/lib/connection.js:409:17)
    at Connection.handlePacket (/home/wyatt/Repos/2873-13-vw-cup-lxp-api/node_modules/mysql2/lib/connection.js:454:14)
    ...
  fatal: true,
  code: 'PROTOCOL_UNEXPECTED_PACKET'
```

Parsonswy:
Fixes #2602.

Check the `nullBitmaskByte` before trying to `readLengthCodedBuffer()` when `typeCast: false` on binary results.

### DIFF
--- a/lib/parsers/binary_parser.js
+++ b/lib/parsers/binary_parser.js
@@ -169,6 +169,10 @@ function compile(fields, options, config) {
       lvalue = `result[${fieldName}]`;
     }
 
+    parserFn(`if (nullBitmaskByte${nullByteIndex} & ${currentFieldNullBit}) `);
+    parserFn(`${lvalue} = null;`);
+    parserFn('else {');
+
     if (options.typeCast === false) {
       parserFn(`${lvalue} = packet.readLengthCodedBuffer();`);
     } else {
@@ -176,9 +180,6 @@ function compile(fields, options, config) {
       parserFn(`const ${fieldWrapperVar} = wrap(fields[${i}], packet);`);
       const readCode = readCodeFor(fields[i], config, options, i);
 
-      parserFn(`if (nullBitmaskByte${nullByteIndex} & ${currentFieldNullBit})`);
-      parserFn(`${lvalue} = null;`);
-      parserFn('else {');
       if (typeof options.typeCast === 'function') {
         parserFn(
           `${lvalue} = options.typeCast(${fieldWrapperVar}, function() { return ${readCode} });`,
@@ -186,8 +187,9 @@ function compile(fields, options, config) {
       } else {
         parserFn(`${lvalue} = ${readCode};`);
       }
-      parserFn('}');
     }
+    parserFn('}');
+
 
     currentFieldNullBit *= 2;
     if (currentFieldNullBit === 0x100) {
--- a/test/integration/config/test-typecast-global-false.test.cjs
+++ b/test/integration/config/test-typecast-global-false.test.cjs
@@ -1,0 +1,33 @@
+'use strict';
+
+const common = require('../../common.test.cjs');
+const connection = common.createConnection({
+  typeCast: false,
+});
+
+const { assert } = require('poku');
+
+const COL_1_VALUE = 'col v1';
+const COL_2_VALUE = 'col v2';
+
+function executeTests(res) {
+  assert.equal(res[0].v1.toString('ascii'), COL_1_VALUE);
+  assert.equal(res[0].n1, null);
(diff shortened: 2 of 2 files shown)


sidorares:
thanks for fix PR and detailed report @Parsonswy ! Do you think it would be possible to include a unit test ( failing without `lib/parsers/binary_parser.js` changes and passing with )?

Parsonswy:
@sidorares Certainly. Test added.

Issue #1446, opened 2021-11-07
RFC: deprecate typeCast and introduce resultsMapper ( exact name TBD )


sidorares:
There are few issues with `typeCast` currently, and most of them better solved by moving it to a different level where it sits in the driver, I think it applies both to `mysql2` / `mysqljs/mysql`

Problems I'd like to solve with the change:

1) most important - typeCast sits too low and is too powerful. It's super easy to make connection go into completely undefined / unpredictable state by reading incorrect number of bytes
2) performance: mysql2 generates row parser JS code in the run time and just the fact that you have custom typeCast function makes generated parser slower and more likely to bail out from optimisations
3) difficult to implement for `execute()`. `typeCast` in its current form expects user to know binary serialization format of results and data for `execute()` if often encoded differently from data for `query()`
3) one can't join 2 rows into one etc as `typeCast` is called before every column of a row is read

Most importantly: almost every use case I see is "map rows from default types into something that works best for our use case" and not "I wan't to change / extend parsing logic".

- #649
- #948
- #889
- https://github.com/sidorares/node-mysql2/issues/1426#issuecomment-945101066
- #709

## Proposed solution

Introduce `resultsMapper` config function (happy to hear better name suggestions). If set, it's applied to every row and result of the function is then used as returned row.

signature:

```js
function resultsMapper(row: Row, fields: Array<Column>) => CustomRow
```

## Deprecation strategy

- soft deprecate `typeCast` - note in the documentation
- next step - decide if we want to remove typeCast completely - any use cases that can't be efficiently solved without having typeCast. If nothing, add deprecation message in the code
- finally, leave a stub throwing an error and pointing to documentation/migration guide

CC ( not sure if I'm tagging correct people, if you know who is better suited to comment please let me know )
seqelize - @sushantdhiman , @papb 
knex - @tgriesser , @kibertoad
mysqljs/mysql @dougwilson

sushantdhiman:
Although I am not currently maintaining Sequelize, I think apart from some refactor work in Sequelize this new implementation can be adopted without much effort, given that `Column` exposes underlying data type (string or symbol).

[...]

PR #2607, opened 2024-04-20
Add .value() method to the typeCast field wrapper

Parsonswy:
I was a bit crude here and most or less copy pasted the `readCodeFor` methods that compile the parser and am open to other approaches.

Another option I considered was modifying the compiled parser so that it could be invoked recursively, but I felt that was going to be a bit hacky since `wrap()` would need to be able to invoke the parser, but the parser is defined using `wrap`. It would avoid duplicating the switch statement though.

I also included a simple benchmark that compares querying with default the default cast behavior, custom type cast with the current wrapper methods, and type cast using the added `.value()` method.

These are the results on my machine against a local docker container with mysql 8.3.0.
```
query raw: AVG 0.0762ms
query typecast.string(): AVG 0.0705ms
query typecast.value(): AVG 0.0677ms
```

### DIFF
--- a/benchmarks/benchmark-typecast.js
+++ b/benchmarks/benchmark-typecast.js
@@ -1,0 +1,79 @@
+'use strict';
+
+const createConnection = require('../test/common.test.cjs').createConnection;
+const connection = createConnection();
+const NUM_SAMPLES = 10000;
+
+function typeCastRaw(field, next) {
+  if (field.type === 'VARCHAR') {
+    return field.string();
+  }
+
+  if (field.type === 'BINARY') {
+    return field.buffer().toString('ascii');
+  }
+
+  return next();
+}
+
+function typeCastValue(field, next) {
+  if (field.type === 'VARCHAR') {
+    return field.value();
+  }
+
+  if (field.type === 'BINARY') {
+    return field.value().toString('ascii');
+  }
+
+  return next();
+}
+
+async function benchmark(iterations, executor, typeCast) {
+  await new Promise((resolve, reject) => {
+    connection.query(
+      'TRUNCATE benchmark_typecast',
+      (err) => {
+        if (err) reject(err);
+        resolve();
+      },
+    );
+  });
+
+  await new Promise((resolve, reject) => {
+    connection.query(
+      'INSERT INTO benchmark_typecast VALUES ("hello", 0x1234)',
+      (err) => {
+        if (err) reject(err);
+        resolve();
+      },
+    );
+  });
+
+  const samples = [];
+  for (let i = 0; i < iterations; i++) {
+    const start = Date.now();
+    await new Promise((resolve, reject) => {
+      connection[executor]({ sql: `SELECT * FROM benchmark_typecast`, typeCast }, (err) => {
+        if (err) reject(err);
+        resolve();
+      });
+    });
+    samples.push(Date.now() - start);
+  }
+
+  console.log(
+    `${executor} ${typeCast ? typeCast : 'raw'}: AVG ${samples.reduce((a, b) => a + b, 0) / samples.length}ms`,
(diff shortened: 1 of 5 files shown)


wellwelwel:
Thanks, @Parsonswy 🚀

There are some **Sequelize** traces on `typeCast` + `execute` that I believe the issues come from **MySQL2**:

- https://github.com/wellwelwel/sequelize/issues/1

> I think these changes can clarify this issue. As soon as possible I intend to check it out in detail 🙋🏻‍♂️

wellwelwel:
A question: this _PR_ would work as a `feat`, right?

Regardless of benchmark, do you think it's possible to make these modifications without "keeping" the breaking changes you mentioned in https://github.com/sidorares/node-mysql2/issues/1446#issuecomment-2067516735?

sidorares:
I'm a bit confused. What's the difference between added `readCodeFor` and existing `readValueFor` ?
Also looks like what we currently pass as a second parameter ( "next()" function ) to a typeCast call is what we actually need from `value()`? Which means that we only need to update documentation

instead of 

```js
  typeCast: function (field, next) {
    if (field.type === 'LONGLONG') {
      return field.string() === '1';
    }
    return next();
  },
```

we can have an example

```js
  typeCast: function (field, getValue) {
    const value = getValue();
    if (field.type === 'LONGLONG') {
      return value === '1';
    }
    return value;
  },
```

And later we can deprecate `typeCast` and introduce `mapValue` which would get value as a parameter:

```js
  mapValue: function (field, value) {
    if (field.type === 'LONGLONG') {
      return value === '1';
    }
    return value;
  },
```

Parsonswy:
>Regardless of benchmark, do you think it's possible to make these modifications without "keeping" the breaking changes you mentioned in https://github.com/sidorares/node-mysql2/issues/1446#issuecomment-2067516735

I don't think I understand the question. By "breaking change" I meant invoking `typeCast` on binary results at all was a break as there are results which when handled by text protocol with a given implementation of `typeCast`will work, but the same implementation would then break if the results are binary parsed. I don't think this PR would have introduced a revert of a breaking change.

Going forward, I don't think it is safe to use the `read*` methods on the field wrapper though because of that text/bin discrepancy. I think the safe/protocol agnostic way to cast is what @sidorares pointed out.

> Also looks like what we currently pass as a second parameter ( "next()" function ) to a typeCast call is what we actually need from value()? Which means that we only need to update documentation.

Hmm. Yes. I think this is totally possible with the current API and I just zeroed in on `.value()`. Updating the docs to to codify that it is ok to invoke `next();` as a way to get a "default" deserialization of the MySQL value and then perform any casting off of that makes sense to me.

I think I can close this PR, but will wait for further comment. Maybe the benchmarks are useful?

sidorares:
let's close `.value()` for now and potentially focus on re-packaging it via documentation changes

`typeCast` benchmarks would be definitely valuable

max-zu:
Any progress on this ? We need to update mysql2 package due to [this](https://github.com/sidorares/node-mysql2/pull/2608) security update. But we can't due to this bug https://github.com/sequelize/sequelize/issues/17141

wellwelwel:
@max-zu

> Any progress on this ? We need to update mysql2 package due to [this](https://github.com/sidorares/node-mysql2/pull/2608) security update. But we can't due to this bug [sequelize/sequelize#17141](https://github.com/sequelize/sequelize/issues/17141)

Although the main issue was opened in **Sequelize**, the issue is seems closer to this https://github.com/sidorares/node-mysql2/issues/1446#issuecomment-2067516735 than to **Sequelize**.

Unfortunately I haven't been able to reproduce this error _(from **Sequelize**)_ using **MySQL2** only. Any help on this would be appreciated.

PR #2608, opened 2024-04-21
fix(security): sanitize timezone parameter value to prevent code injection

sidorares:


wellwelwel:
All the examples I had stopped working 🔐

wellwelwel (approved):

PR #2628, opened 2024-04-26
fix(stream): reads should emit the dataset number for each dataset

wellwelwel:
> This PR is a debug/fix of #2496, preserving all its commits.

Testing locally:

```bash
seq 1000 | xargs -I {} sh -c 'FILTER=test-multi-result-streaming npm run test || echo "Failure at iteration {}" >> failures.log'
```

- I ran all the **CI** 3 times, just to make sure.


wellwelwel:
Just confirming from [**mysqljs**](https://github.com/mysqljs/mysql) side before merging:

- https://github.com/mysqljs/mysql/blob/dc9c152a87ec51a1f647447268917243d2eab1fd/lib/protocol/sequences/Query.js#L210-L213
- https://github.com/mysqljs/mysql/blob/dc9c152a87ec51a1f647447268917243d2eab1fd/lib/protocol/sequences/Query.js#L223-L225

> I also tested on **Deno**'s latest tests and all worked properly.

PR #2699, opened 2024-05-24
Fixed typing for caching_sha2_password where misspelt jonServerPublicKey is replaced by onServerPublicKey

johanwestling:
Just a minor character missing as I assume it was supposed to be `joinServerPublicKey`  instead of `jonServerPublicKey`.

### DIFF
--- a/typings/mysql/lib/Auth.d.ts
+++ b/typings/mysql/lib/Auth.d.ts
@@ -14,7 +14,7 @@ export const authPlugins: {
   caching_sha2_password: AuthPluginDefinition<{
     overrideIsSecure?: boolean;
     serverPublicKey?: RsaPublicKey | RsaPrivateKey | KeyLike;
-    joinServerPublicKey?: (data: Buffer) => void;
+    onServerPublicKey?: (data: Buffer) => void;
   }>;
   mysql_clear_password: AuthPluginDefinition<{
     password?: string;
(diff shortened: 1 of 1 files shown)


sidorares:
hi @johanwestling , thanks for noticing the typo. In should be `onServerPublicKey`, see how its used here: https://github.com/sidorares/node-mysql2/blob/5c75802493cb37997d0eddf4fd9d1e9320377007/lib/auth_plugins/caching_sha2_password.js#L90

this is the callback which users can pass to be able to cache server public key and use it with initial connection to speed it up ( one less roundtrip ) or make more secure if tls is not used ( password is sent RSA encrypted using the key )

johanwestling:
> hi @johanwestling , thanks for noticing the typo. In should be `onServerPublicKey`, see how its used here:
> 
> https://github.com/sidorares/node-mysql2/blob/5c75802493cb37997d0eddf4fd9d1e9320377007/lib/auth_plugins/caching_sha2_password.js#L90
> 
> this is the callback which users can pass to be able to cache server public key and use it with initial connection to speed it up ( one less roundtrip ) or make more secure if tls is not used ( password is sent RSA encrypted using the key )

Sorry for late response, was in transit. Now the `joinServerPublicKey` is replaced with the correct `onServerPublicKey` instead.

PR #2702, opened 2024-05-24
fix(security): sanitize fields and tables when using nestTables

wellwelwel:
Improves the changes from #2591 by validating `nestTables` final fields and tables.

---

I'm not sure about the need to validate when used in combination with the `rowAsArray` option:

> For now, I've applied validation to all the properties created in `result` object.

```js
} else if (options.rowsAsArray) {
  lvalue = `result[${helpers.fieldEscape(i.toString(10))}]`;
} else {
```

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -73,3 +73,15 @@ const privateObjectProps = new Set([
 ]);
 
 exports.privateObjectProps = privateObjectProps;
+
+const fieldEscape = (field) => {
+  if (privateObjectProps.has(field)) {
+    throw new Error(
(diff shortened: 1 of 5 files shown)


sidorares:
thanks @wellwelwel !
LTM, but I can't see how can `i.toString(10)` ever possibly dangerous, probably no need for sanitization here?

wellwelwel:
> thanks @wellwelwel ! LTM, but I can't see how can `i.toString(10)` ever possibly dangerous, probably no need for sanitization here?

It's done 🙋🏻‍♂️

PR #2714, opened 2024-05-29
ci(deno): run the entire test suite with Deno

wellwelwel:
This _PR_ includes the entire test suite to be run with **Deno**.

### Some observations

- There is some incompatibility with **Deno**'s `tls`, which is why the matrix only executes `SSL=0`.
- Tests that deconstruct the `Error` class or wait for the `uncaughtException` event cause errors for both **Bun** and **Deno**.
- Tests that use `createServer` never end.
- In total, 19 tests were silenced for **Deno**:
  - _test-server-close.test.cjs_
  - _test-pool-connect-error.test.cjs_
  - _test-handshake-unknown-packet-error.test.cjs_
  - _test-auth-switch.test.cjs_
  - _test-auth-switch-plugin-error.test.cjs_
  - _test-auth-switch-plugin-async-error.test.cjs_
  - _test-auth-switch-multi-factor.test.cjs_
  - _test-async-stack.test.cjs_
  - _test-stream-errors.test.cjs_
  - _test-quit.test.cjs_
  - _test-query-timeout.test.cjs_
  - _test-protocol-errors.test.cjs_
  - _test-disconnects.test.cjs_
  - _test-connect-time-error.test.cjs_
  - _test-connect-sha1.test.cjs_
  - _test-connect-connection-closed-error.test.cjs_
  - _test-connect-after-connection-error.test.cjs_
  - _test-change-user-multi-factor.test.cjs_
  - _test-connect-timeout.test.cjs_

> Note that this _PR_ isn't intended to fix possible issues, but to allow tests to be run at runtime.


wellwelwel:
Most tests also work with **Bun**, but there is some incompatibility when using callbacks with events that don't terminate processes.

To do the same with **Bun**, all tests that use callbacks would need to include this in the last successful event:

```js
connection.end((err) => {
  assert.ifError(err);
  process.exit(0);
});
```

PR #2757, opened 2024-06-11
fix: setMaxParserCache throws TypeError

### LINKED ISSUE #2752: [bug][regression] mysql2.setMaxParserCache fails with TypeError
Hi!

If I call `setMaxParserCache` imported from mysql2, I get an error:

```
<snip>/node_modules/mysql2/lib/parsers/parser_cache.js:54
  parserCache.max = max;
                  ^
TypeError: Cannot set property max of #<LRUCache> which has only a getter
```

I *really need* to downsize this cache because it always ends up being full and consuming about 300Mb RAM. Probably something about my use case that makes it grow so big…

I believe somewhere along the way `lru-cache` changed its internals and cannot be resized anymore. I'd like if `setMaxParserCache` would throw away an old cache and create a new one with right size in its place…

constb:
replace parser cache with a new one on setMaxParserCache to avoid TypeError, because `max` is now read-only on LRU caches and cannot be altered.

### DIFF
--- a/lib/parsers/parser_cache.js
+++ b/lib/parsers/parser_cache.js
@@ -2,7 +2,7 @@
 
 const LRU = require('lru-cache').default;
 
-const parserCache = new LRU({
+let parserCache = new LRU({
   max: 15000,
 });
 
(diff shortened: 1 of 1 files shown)


wellwelwel:
Hi @constb, thanks 🚀

About the `codecov/patch`, as you changed a file that was no longer covered, it blocks the **CI**, even though it's not related to your change.

---

This _PR_ allows to continue showing the alerts, but without blocking the CI due to files individually.

- #2674

PR #2802, opened 2024-06-22
Modifying Chinese documentation in some places is still a matter for English

WenisOK:
Modifying Chinese documentation in some places is still a matter for English


wellwelwel:
Thanks, @WenisOK 🚀

PR #2866, opened 2024-07-18
Add basic vector type

dbussink:
This adds the new MySQL 9.0 vector type. It can still be handled like a binary blob for now I think. Maybe it's worth in the future to directly decode / parse it into an array of floats here?

### DIFF
--- a/lib/constants/types.js
+++ b/lib/constants/types.js
@@ -51,6 +51,7 @@ module.exports.YEAR = 0x0d; // aka YEAR, 1 byte (don't ask)
 module.exports.NEWDATE = 0x0e; // aka ?
 module.exports.VARCHAR = 0x0f; // aka VARCHAR (?)
 module.exports.BIT = 0x10; // aka BIT, 1-8 byte
+module.exports.VECTOR = 0xf2;
 module.exports.JSON = 0xf5;
 module.exports.NEWDECIMAL = 0xf6; // aka DECIMAL
 module.exports.ENUM = 0xf7; // aka ENUM
--- a/typings/mysql/lib/constants/Types.d.ts
+++ b/typings/mysql/lib/constants/Types.d.ts
@@ -16,6 +16,7 @@ interface Types {
   0x0e: string;
   0x0f: string;
   0x10: string;
+  0xf2: string;
   0xf5: string;
   0xf6: string;
   0xf7: string;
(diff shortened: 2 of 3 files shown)


dbussink:
See https://dev.mysql.com/doc/dev/mysql-server/latest/field__types_8h.html also for the constant value here.

sidorares:
hey @dbussink do you have any pointers to documentation on how the value is transmitted over the wire? Or maybe you can help with some examples I can run and try to reverse engineer serialisation format.

Happy to merge without detailed deserealizing, we can tackle that later

wellwelwel:
@sidorares, can we include a _symbolic_ commit to release it? Also, to include it in change logs.

Or if you prefer, we can wait until a next release to bring this update along, but it won't be included in the logs.

dbussink:
> hey @dbussink do you have any pointers to documentation on how the value is transmitted over the wire? Or maybe you can help with some examples I can run and try to reverse engineer serialisation format.

See https://dev.mysql.com/doc/refman/9.0/en/vector.html and if you look in the MySQL CLI, it's a binary blob of data. It's always a multiple of 4 bytes, since it's up to `N` 32 bit floating point values all concatenated together in that single blob.

sidorares:
@wellwelwel sorry I should've updated commit to have `feature:` prefix. I'll try to add deserealisation soon and reference this PR

sidorares:
> > hey @dbussink do you have any pointers to documentation on how the value is transmitted over the wire? Or maybe you can help with some examples I can run and try to reverse engineer serialisation format.
> 
> See https://dev.mysql.com/doc/refman/9.0/en/vector.html and if you look in the MySQL CLI, it's a binary blob of data. It's always a multiple of 4 bytes, since it's up to `N` 32 bit floating point values all concatenated together in that single blob.

So it might be as simple as doing `packet.readFloat()` until the end and pushing numbers to the array

sidorares:
yes, `readFloat` should be enough

```js
connection.execute(`SELECT TO_VECTOR("[1.05, -17.8, 32, 123.456]")`, (err, res) => {
  if (err) throw err;
  const buf = Object.values(res[0])[0];

  let offset = 0;
  while (offset < buf.length) {
    const v = buf.readFloatLE(offset);
    console.log(v);
    offset += 4;
  }

  connection.end();
});
```

prints

```
1.0499999523162842
-17.799999237060547
32
123.45600128173828
```

sidorares:
@dbussink this + #2894 published as v3.11.0

PR #2894, opened 2024-07-27
feat: return VECTOR values as array of numbers instead of raw Buffer

sidorares:
this is a continuation of work started in https://github.com/sidorares/node-mysql2/pull/2866

with this change [VECTOR](https://dev.mysql.com/doc/refman/9.0/en/vector.html) values are returned as array of JS numbers

### DIFF
--- a/.github/workflows/ci-mysql.yml
+++ b/.github/workflows/ci-mysql.yml
@@ -24,6 +24,7 @@ jobs:
             'mysql:8.0.18',
             'mysql:8.0.22',
             'mysql:8.0.33',
+            'mysql:9.0.1',
             'mysql:latest',
           ]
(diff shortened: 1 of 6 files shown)


wellwelwel:
> _Just a note_

@sidorares, for some reason the _(non-related)_ `test/esm/regressions/2052.test.mjs` is failing on **Windows**.

sidorares:
> > _Just a note_
> 
> @sidorares, for some reason the _(non-related)_ `test/esm/regressions/2052.test.mjs` is failing on **Windows**.

It was failing on a mac as well, I believe `hasIncorrectPrepareParameter` was calculated incorrectly for 8.0.39

PR #2983, opened 2024-08-26
ci: remove workaround in Bun tests

wellwelwel:
After https://github.com/oven-sh/bun/pull/13212 and https://github.com/oven-sh/bun/pull/13412, conflicts related to `net` that didn't allow processes to terminate have been fixed.

If these two tests pass, we can run the entire test suite for `Bun` too 🧑🏻‍🔬


wellwelwel:
Interesting, the problem still seems to occur when **SSL** is enabled.

PR #3017, opened 2024-09-05
Add pattern and selector to PromisePoolCluster#getConnection

### LINKED ISSUE #1381: PromisePoolCluster missing functionality
Tried to use the PromisePoolCluster implementation that was recently included.

It appears to be missing at least the `remove` inherited event. Not sure if there are others.
It also doesn't support passing a pattern or selector to `getConnection`.

Seems like you maybe be able to workaround the first by using `promisePoolCluster.poolCluster.on('remove'...`
And I suppose you can always workaround the second by refactoring your code to call `.of(pattern, selector)`, but it seems like these should be added to match the documentation for PoolCluster.

TuckerWhitehouse:
The implementation of `PromisePoolCluster#getConnection` does not match the capabilities of `PoolCluster#getConnection` nor the definition in `promise.d.ts`.

This change corrects this by adding the missing `pattern` and `selector` parameters which are passed through to the underlying method.

Fixes #1381 

Note: I may have missed them, but I didn't see any tests for PromisePoolCluster to update or add to with this change.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -451,10 +451,10 @@ class PromisePoolCluster extends EventEmitter {
     inheritEvents(poolCluster, this, ['warn', 'remove']);
   }
 
-  getConnection() {
+  getConnection(pattern, selector) {
     const corePoolCluster = this.poolCluster;
     return new this.Promise((resolve, reject) => {
-      corePoolCluster.getConnection((err, coreConnection) => {
+      corePoolCluster.getConnection(pattern, selector, (err, coreConnection) => {
         if (err) {
           reject(err);
         } else {


wellwelwel:
Thanks, @TuckerWhitehouse 🙋🏻‍♂️

Just confirming it:

https://github.com/sidorares/node-mysql2/blob/dc3a68081d7293628e1d44591b0bf8f5cd217885/lib/pool_cluster.js#L165-L178

I just need to check if the typings are correct.

wellwelwel:
LGTM

https://github.com/sidorares/node-mysql2/blob/dc3a68081d7293628e1d44591b0bf8f5cd217885/promise.d.ts#L112-L114

Issue #3030, opened 2024-09-10
Exposed .escape() functions exposes but don't cover the MySQL original typing


ScrapsPrograms:
During some experimentation, I found that the type definition of the `.escape(value: any)` function isn't correct. Currently, for all locations, it's described as `.escape(value: any): string`.

The function exposed by the package, however, has a different interface: `.escape(value: any, stringifyObjects: boolean, timeZone: string | "local")`. From what I've found so far, this applies only to the loose exposed function, not to any of the class defined methods, which seem to wrap this function by abstraction. 

For reference, the classes with a `.escape(value: any)` method are in Pool and Connection.

I'm willing to put a PR for this, if you agree the typing information should be updated, or if an abstraction should be applied to the loose exposed function similar to the class defined `.escape` methods.

wellwelwel:
Thanks for the report, @ScrapsPrograms 🙋🏻‍♂️

> I'm willing to put a PR for this, if you agree the typing information should be updated, or if an abstraction should be applied to the loose exposed function similar to the class defined `.escape` methods.

Please, feel free to contribute 🤝

wellwelwel:
> .escape(value: any, stringifyObjects: boolean, timeZone: string | "local")

Just a tip:

Instead of `string | 'local'`, you can use `'local' | string & {}`. It will allow using a dynamic string, keeping the auto suggestion support, e.g.:

<img width="641" alt="Screenshot 2024-09-10 at 05 43 04" src="https://github.com/user-attachments/assets/0f5fe9d4-66bd-4932-8274-144d1d0b4c6f">

PR #3047, opened 2024-09-11
fix: Export types directly from sqlstring dependency

ScrapsPrograms:
Instead of defining new types, this change exports the `sqlstring` dependency's typing information directly.

Originally the idea was to just update the type of `escape`, but looking a bit further, this change seems to make more sense.
For reference, see [this issue](https://github.com/sidorares/node-mysql2/issues/3030).

### DIFF
--- a/typings/mysql/index.d.ts
+++ b/typings/mysql/index.d.ts
@@ -1,3 +1,6 @@
+/**
+ * sqlstring types are based on https://www.npmjs.com/package/@types/sqlstring, version 2.3.2
+ */
 import { Pool as BasePool, PoolOptions } from './lib/Pool.js';
 import {
   Connection as BaseConnection,
@@ -53,35 +56,10 @@ export function createPool(config: PoolOptions): BasePool;
 
 export function createPoolCluster(config?: PoolClusterOptions): PoolCluster;
 
-/**
- * @file index.d.ts
- * 
- * @author marvinhagemeister
- * @date 11 July 2017
- * 
- * @dependency \@types/sqlstring
- * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
- */
 export function escape(value: any, stringifyObjects?: boolean, timeZone?: 'local' | NonNullable<unknown>): string;
-/**
- * @file index.d.ts
- * 
- * @author marvinhagemeister
- * @date 11 July 2017
- * 
- * @dependency \@types/sqlstring
- * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
- */
+
 export function escapeId(value: any, forbidQualified?: boolean): string;
-/**
- * @file index.d.ts
- * 
- * @author marvinhagemeister
- * @date 11 July 2017
- * 
- * @dependency \@types/sqlstring
- * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
- */
+
 export function format(sql: string): string;
 export function format(
   sql: string,
@@ -89,15 +67,7 @@ export function format(
   stringifyObjects?: boolean,
   timeZone?: 'local' | NonNullable<unknown>,
 ): string;
-/**
- * @file index.d.ts
- * 
- * @author David Gomes
- * @date 3 January 2019
- * 
- * @dependency \@types/sqlstring
- * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
- */
+
 export function raw(sql: string): {
   toSqlString: () => string;
 };
--- a/typings/mysql/index.d.ts
+++ b/typings/mysql/index.d.ts
@@ -53,10 +53,35 @@ export function createPool(config: PoolOptions): BasePool;
 
 export function createPoolCluster(config?: PoolClusterOptions): PoolCluster;
 
+/**
+ * @file index.d.ts
+ * 
+ * @author marvinhagemeister
+ * @date 11 July 2017
+ * 
+ * @dependency \@types/sqlstring
+ * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
+ */
 export function escape(value: any, stringifyObjects?: boolean, timeZone?: 'local' | NonNullable<unknown>): string;
-
+/**
+ * @file index.d.ts
+ * 
+ * @author marvinhagemeister
+ * @date 11 July 2017
+ * 
+ * @dependency \@types/sqlstring
+ * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
+ */
 export function escapeId(value: any, forbidQualified?: boolean): string;
-
+/**
+ * @file index.d.ts
+ * 
+ * @author marvinhagemeister
+ * @date 11 July 2017
+ * 
+ * @dependency \@types/sqlstring
+ * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
+ */
 export function format(sql: string): string;
 export function format(
   sql: string,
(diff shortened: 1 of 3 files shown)


wellwelwel:
Thanks, @ScrapsPrograms 🚀

I think it's better to include it in the documentation, similar to what is done with `@types/node`, instead of bringing `@types/sqlstring` to dependencies:

https://github.com/sidorares/node-mysql2/blob/3d2327d683254ad3642b9c3644ebc486dab7797e/website/docs/index.mdx#L55-L58

---

For example, you're re-exporting all types from `@types/sqlstring`, but as a development dependency, the final user will not have these types installed.

What do you think, @ScrapsPrograms?

wellwelwel:
> For example, you're re-exporting all types from `@types/sqlstring`, but as a development dependency, the final user will not have these types installed.

It's important to note that this approach can imply a breaking change for **TypeScript** users 🙋🏻‍♂️

ScrapsPrograms:
TLDR; The added `@types/sqlstring` may be moved to dependencies and the original overloads for the `format` function could be returned in a simplified form `format(sql: string, args?: any | any[], stringifyObjects?: boolean, timeZone?: string): string`, so the `args?: object | any[]` argument type does not cause conflicts.

> I think it's better to include it in the documentation, similar to what is done with @types/node, instead of bringing @types/sqlstring to dependencies:

If I understand this correctly, the documentation here asks users to install the type definitions themselves. I would argue that, for a function exported by the node-mysql2 package, it'd make more sense for the typing to be exported alongside it, rather than requiring a separate user installation. 

> as a development dependency, the final user will not have these types installed.

This is true, and I have to admit I overlooked that. 

> It's important to note that this approach can imply a breaking change for TypeScript users

This is understandable. Looking at the originally exported types and the types exported by `sqlstring`, they are mostly the same. Only the function `format`, which had a few overloads in the node-mysql2 package, provides a different type.
Most arguments in the `sqlstring` package `format` function are optional. Additionally, the type of the argument `value: any` in the node-mysql2 package's `escape` definition is is different in the `sqlstring` package's definition: `args: object | any[]`. It may be a breaking change for **TypeScript** users if the provided `value` is neither an object nor an array of any kind. The other exposed functions cannot break any existing usage of node-mysql2's exposed `sqlstring` functions, as the typing is either the same or more limited.

To ensure no breaking changes are applied, an overload for the `format` function may be implemented to replace `args: object | any[]` with the originally applied `args: any | any[]`. I would like to have your thoughts on this matter.

wellwelwel:
> The added `@types/sqlstring` may be moved to dependencies

I'm not sure about this idea. Type dependencies shouldn't affect JavaScript users, for example by requiring `@types/node`, it will increase more than 2MB to **MySQL2** install size.

- Just to clarify, I like the idea of bringing in the development dependency and exposing its types, documenting it as a dependency to be exported manually for TypeScript users. Although I think this is ideal, the problem would be to cause a breaking change because of this.

> To ensure no breaking changes are applied, an overload for the `format` function may be implemented to replace `args: object | any[]` with the originally applied `args: any | any[]`. I would like to have your thoughts on this matter.

For now, I think it's better to set the types directly in **MySQL2**. Let me know your thoughts 🙋🏻‍♂️

ScrapsPrograms:
This seems to push the `@types/sqlstring` dependency towards becoming a peerDependency. AFAIK, this means it won't be included in the JavaScript version, as npm doesn't install peerDependencies by itself. With this change, the documentation will indeed need to be updated, as per your previous suggestion.
Sidenote: A different package manager, such as `pnpm`, might install peerDependency packages, but often this isn't a problem.
Let's call this possible solution 1.

Another option is to keep going the way node-mysql2 originally handled this: copy the types and handle it in the types package. For this PR, that'd just be a matter of copying the types exposed in `@types/sqlstring` and replacing the current types with that. 
Personally, I'd prefer to find ways to have the dependencies handle their own typing, as changes can be updated through project dependencies rather than having to be updated manually when a change is requested, it seems a lot easier. 

It seems the `sqlstring` package itself hasn't been updated since March 2022, with no PRs having been opened since a year prior to its last release. I would assume the typing information won't change anytime soon.

> It may be a breaking change for TypeScript users if the provided value is neither an object nor an array of any kind.
This can be covered by replacing the type `value: object | any[]` with `value: any | any[]`. I've never used this `format` function myself, so I cannot say if anyone's passed a non-object, non-array type to it as the second argument or why they would be doing so (or if it works even).
Let's call this possible solution 2.

Lastly, and I'm not much a fan of this reasoning, but it must be mentioned to cover the last option I can come up with at this time, the typing information of `@types/sqlstring` exists of only one or two files, resulting in about 16kb added to the dependencies. It _could_ be considered small enough to be irrelevant, and thus acceptable to be added. But again, I'm personally not a fan of that.

Unless a means of properly distributing the `@types/sqlstring` dependency can be applied without affecting vanilla users, my vote goes to option 2, which is to copy the typing information.

wellwelwel:
> It seems the `sqlstring` package itself hasn't been updated since March 2022, with no PRs having been opened since a year prior to its last release. I would assume the typing information won't change anytime soon.

`sqlstring` repository is closed only form members (no possibility to open issues or PRs).

---

> Unless a means of properly distributing the `@types/sqlstring` dependency can be applied without affecting vanilla users, my vote goes to option 2, which is to copy the typing information.

For version `3.x.x`, I think that define the types directly in **MySQL2** is a better approach. So when there is a milestone for version `4`, include the dependency point as a goal.

Just a note:

- If a copy is indeed made, it is important to credit the source in a "header comment" 🙋🏻‍♂️

typings/mysql/index.d.ts
```diff
@@ -53,25 +53,51 @@ export function createPool(config: PoolOptions): BasePool;
...
-  sql: string,
-  values: any,
+  values: any | any[],
   stringifyObjects?: boolean,
-  timeZone?: string,
+  timeZone?: 'local' | NonNullable<unknown>,
 ): string;
-
+/**
+ * @file index.d.ts
```
wellwelwel, line 93: @ScrapsPrograms, I was thinking something simple like:

> - Top of the file ⬇️

```ts
/**
 * sqlstring typings were taken from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sqlstring/index.d.ts
 */

// ...
```

Or "based" instead of "taken", since the typings aren't the same 🙋🏻‍♂️

wellwelwel (changes requested):

typings/mysql/index.d.ts
```diff
@@ -53,25 +53,51 @@ export function createPool(config: PoolOptions): BasePool;
...
+ * 
+ * @author marvinhagemeister
+ * @date 11 July 2017
+ * 
+ * @dependency \@types/sqlstring
+ * @see {@link https://www.npmjs.com/package/@types/sqlstring} for more information
+ */
+export function escape(value: any, stringifyObjects?: boolean, timeZone?: 'local' | NonNullable<unknown>): string;
```
wellwelwel, line 65: I reviewed the `sqlstring` code and the timezone always waits for a string, with two as default: `"local"` and `"Z"`.

I suppose the reason for choosing `NonNullable<unknown>` was to allow auto-suggestions.

`'local' | 'Z' | string & {}` seems more appropriate.

If `string & {}` sounds unfamiliar, we can use it like:

```ts
type CustomString = string & {};

type TimeZone = 'local' | 'Z' | CustomString;
```

wellwelwel (changes requested):

typings/mysql/index.d.ts
ScrapsPrograms: The NonNullable<unknown> was suggested by my IDE to be the perfect replacement for `string & {}`. Personally I don't really care which one's used. Oddly enough, the type of NonNullable is defined as `T & {}` yet using NonNullable<string> does not yield the same results as `string & {}`. 

I'll apply your change requests momentarily. Thank you for your assistance here.

ScrapsPrograms: Is it alright with you if I'm keeping `string & NonNullable<string>` in, so the typescript-eslint rule "@typescript-eslint/ban-types" doesn't cause any errors?

wellwelwel (approved):

wellwelwel:
Thanks, @ScrapsPrograms 🚀

PR #3169, opened 2024-10-30
Optimize Error Object Creation in `promise.js` Functions to Improve Performance

Sunghak:
### Summary

This merge request optimizes several functions in `promise.js`, including `ping()`, `execute()`, `query()`, and other similar methods, by moving the creation of `Error` objects inside their respective error handling blocks. This change ensures that `Error` objects are instantiated only when actual errors occur, reducing unnecessary overhead during successful operations.

### Background and Motivation

- **Performance Issue:**
  - Previously, `Error` objects were being created every time these functions were called, regardless of whether an error occurred.
  - In high-throughput applications where these functions are invoked frequently (e.g., 280,000 times), this leads to significant CPU and memory overhead.

- **Performance Analysis:**
  - Used Clinic Flame Graph to analyze performance over 280,000 function calls.
  - Found that creating `Error` objects only when needed reduces CPU usage and execution time.
  - Observed improved performance and reduced CPU time when the `Error` object instantiation is deferred until an actual error occurs.

### Changes Made

- **Modified Functions:**
  - `ping()`
  - `execute()`
  - `query()`
  - Other functions in `promise.js` exhibiting the same pattern.

- **Optimization Details:**
  - Moved the instantiation of `Error` objects inside the `if (err)` blocks within each function.
  - Ensured that `Error` objects are only created when an error is present.
  - Example change:

    ```javascript
    // Before
    someFunction() {
      const c = this.connection;
      const localErr = new Error(); // Error created here regardless of error occurrence
      return new this.Promise((resolve, reject) => {
        c.someOperation((err, result) => {
          if (err) {
            localErr.message = err.message;
            // ... copy other properties
            reject(localErr);
          } else {
            resolve(result);
          }
        });
      });
    }

    // After
    someFunction() {
      const c = this.connection;
      return new this.Promise((resolve, reject) => {
        c.someOperation((err, result) => {
          if (err) {
            const localErr = new Error(); // Error created here only when an error occurs
            localErr.message = err.message;
            // ... copy other properties
            reject(localErr);
          } else {
            resolve(result);
          }
        });
      });
    }
    ```

### Benefits

- **Performance Improvement:**
  - Reduces CPU and memory usage by avoiding unnecessary creation of `Error` objects during successful operations.
  - Decreases pressure on the garbage collector, leading to smoother runtime performance.

- **Scalability:**
  - Enhances the efficiency and scalability of applications that frequently use these functions.
  - Particularly beneficial in high-load scenarios where small optimizations have a significant impact.

- **Best Practices Compliance:**
  - Aligns with modern JavaScript best practices by limiting variable scope and using lazy instantiation.
  - Improves code readability and maintainability by keeping variable declarations close to where they are used.

### Testing

- **Performance Testing:**
  - Conducted benchmarks to compare CPU usage before and after the changes.
  - Confirmed a measurable decrease in CPU time and improved performance in high-throughput tests.

### Compatibility

- **Backward Compatibility:**
  - Changes are fully backward-compatible with the existing API.
  - No alterations to function signatures or expected inputs/outputs.

### Additional Notes

- **User Impact:**
  - Users should experience improved performance without any changes required on their part.
  - The optimization is internal and does not affect how the library is used externally.

### Conclusion

By optimizing the creation of `Error` objects in `ping()`, `execute()`, `query()`, and other functions within `promise.js`, this change enhances the performance and scalability of the `mysql2` library. Avoiding unnecessary instantiation of `Error` objects reduces CPU and memory overhead, leading to more efficient resource utilization, especially in applications with high transaction volumes.

### DIFF
--- a/promise.js
+++ b/promise.js
@@ -4,9 +4,10 @@ const core = require('./index.js');
 const EventEmitter = require('events').EventEmitter;
 const parserCache = require('./lib/parsers/parser_cache.js');
 
-function makeDoneCb(resolve, reject, localErr) {
+function makeDoneCb(resolve, reject) {
   return function (err, rows, fields) {
     if (err) {
+      const localErr = new Error();
       localErr.message = err.message;
       localErr.code = err.code;
       localErr.errno = err.errno;
@@ -52,9 +53,8 @@ class PromisePreparedStatementInfo {
 
   execute(parameters) {
     const s = this.statement;
-    const localErr = new Error();
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       if (parameters) {
         s.execute(parameters, done);
       } else {
@@ -91,14 +91,13 @@ class PromiseConnection extends EventEmitter {
 
   query(query, params) {
     const c = this.connection;
-    const localErr = new Error();
     if (typeof params === 'function') {
       throw new Error(
         'Callback function is not available with promise clients.'
       );
     }
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       if (params !== undefined) {
         c.query(query, params, done);
       } else {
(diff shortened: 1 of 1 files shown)


sidorares:
This optimization has been rebased onto the current modular codebase and continued in #4257.

The original `promise.js` has since been refactored into separate modules (`lib/promise/connection.js`, `lib/promise/pool.js`, `lib/promise/make_done_cb.js`, etc.), so the same deferred-Error-creation pattern has been applied across all of them.

Thank you @gunmanplus for the original contribution and analysis!

PR #3177, opened 2024-11-01
fix: fix wrong length number write to packet

remones:


wellwelwel:
Hi, @remones 🙋🏻‍♂️

Could you provide some context?
As a fix, do you think a test can be added to cover the issue that is being fixed?

remones:
> Hi, @remones 🙋🏻‍♂️
> 
> Could you provide some context? As a fix, do you think a test can be added to cover the issue that is being fixed?

Sure, I will provide the test code later. 

Please let me explain the issue first: my node.js service uses mysql2 to access a MySQL proxy service. When attempting to write a 32MB TEXT field, it raises some errors. we found the reason is that the field length in the MySQL protocol packet read by the proxy far exceeds 32MB. I believe the issue is caused by JavaScript bitwise operations not supporting numbers larger than 32 bits.

wellwelwel:
Thanks, @remones! I fixed a `RangeError` issue and added tests to cover the fix.

PR #1147, opened 2020-04-19
fix issue #1040

sidorares:
Fixes #1040.

PR #3204, opened 2024-11-12
fix: datetime fields returned without time part when time is 00:00:00

### LINKED ISSUE #1040: datetime fields returned as date when time is 00:00:00
Setting a DATETIME field with the value 2019-10-01 00:00:00
Then retrieve the data using the execute command:
`const [result] = connection.execute('SELECT * FROM my_table', [])`

result will equal to '2019-10-01' instead of '2019-10-01 00:00:00' although it's a DATETIME field and not merely a DATE

note that when using query the behavior is as expected

w666:
Fixes #1040.

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -9,7 +9,7 @@ const ErrorCodeToName = require('../constants/errors.js');
 const NativeBuffer = require('buffer').Buffer;
 const Long = require('long');
 const StringParser = require('../parsers/string.js');
-
+const Types = require('../constants/types.js');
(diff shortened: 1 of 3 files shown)


w666:
@sidorares this is pretty much the same as https://github.com/sidorares/node-mysql2/pull/1147.
I am not familiar with poku, but it was very easy to update existing test, hopefully this is enough.

w666:
Sorry for the failed build, transferred that patch from work machine to submit from my personal one and it did not work as expected, fixing now.

Issue #2667, opened 2024-05-10
Incorrect typings lead to pool.destroy is not a function


dancrumb:
I'm using version 3.9.7 of `mysql2`, running TypeScript 5.4.3

I'm importing this module with `import mysql from "mysql2/promise";`

When I create a `Pool` with `mysql.createPool`, I get my pool and am able to use it without any problems.

However, when I come to wrap things up, I call `pool.destroy()`. No complaints from TS, but when I run the code, I get:

```
<REDACTED>/db/stores/SqlStore.js:19
        pool.destroy();
             ^

TypeError: pool.destroy is not a function
    at SqlStore.shutdown (<REDACTED>/db/stores/SqlStore.js:19:14)
    at <REDACTED>/tools/migrate-to-mongo.mts:85:54
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I took a look at `promises.d.ts` in the code base and I can see why there's no type error:

```
export interface Pool extends Connection {
```

Since `destroy` is a method in the `Connection` interface, it's also a method in the `Pool` interface.

I'm not sure if the bug is the assertion that `Pool` extends `Connection` or whether it's just that `Pool` only supports _some_ methods of `Connection`.

dancrumb:
I'm guessing that this is probably what's going on with https://github.com/sidorares/node-mysql2/issues/2465

wellwelwel:
Thanks, @dancrumb 🙋🏻‍♂️

---

I'll keep some debugging to check later:

```ts
import mysql from 'mysql2';
// import mysql from 'mysql2/promise';

const pool = mysql.createPool({});

typeof pool.addListener === 'undefined' && console.log('addListener');
typeof pool.beginTransaction === 'undefined' && console.log('beginTransaction');
typeof pool.changeUser === 'undefined' && console.log('changeUser');
typeof pool.commit === 'undefined' && console.log('commit');
typeof pool.config === 'undefined' && console.log('config');
typeof pool.connect === 'undefined' && console.log('connect');
typeof pool.destroy === 'undefined' && console.log('destroy');
typeof pool.emit === 'undefined' && console.log('emit');
typeof pool.end === 'undefined' && console.log('end');
typeof pool.escape === 'undefined' && console.log('escape');
typeof pool.escapeId === 'undefined' && console.log('escapeId');
typeof pool.eventNames === 'undefined' && console.log('eventNames');
typeof pool.execute === 'undefined' && console.log('execute');
[...]

Issue #3273, opened 2024-12-09
Breaking Change: PoolConnection is no longer derived from Connection since 3.11.5.


wakabayashik:
```js
import mysql from 'mysql2';
const Connection = mysql.Connection;
const PoolConnection = mysql.PoolConnection;
console.log(PoolConnection.prototype instanceof Connection);
```

mysql2@3.11.4 : true
mysql2@3.11.5 : false

wellwelwel:
Probably related:

- https://github.com/sidorares/node-mysql2/pull/3081

PR #3081, opened 2024-09-27
fix: circular dependencies

tpraxl:
This commit addresses circular dependencies between index.js and promise.js, as well as within files in the lib directory such as pool.js and pool_connection.js.

It introduces a new common.js module to centralize shared functionalities, thus breaking the direct dependencies among these files.

This approach resolves issues with the [@opentelemetry/instrumentation-mysql2](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-mysql2) package, which was losing several exports due to the circular dependencies.

You can reproduce the problem with this simple test repo:

https://github.com/tpraxl/test-esm-mysql-otel

Follow the instructions in the README.md to see the respective warnings.

To verify that this PR fixes the issue, check it out, run `npm i && npm pack` and install the fixed version into the test repo using `npm i $PATH_TO_MYSQL2/mysql2-3.11.3.tgz`. Run `npm start` again in the checked out test repo to see the warnings vanish.

This PR would fix https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1511

PR #3321, opened 2025-01-13
fix: connection types mismatched code

### LINKED ISSUE #3238: Promise Pool events inherit original pool connection class, not promise pool
When using a Promise Pool, and subscribing to events (e.g.`pool.on('acquire', (connection) => {})`) the connection object does not match the type.

In the types, the connection type is exposed as a Promise PoolConnection, where in reality there is no conversion from the original, and so it's actually a base PoolConnection type. Meaning if I want to use it, i have to call the `Promise()` helper, which isn't on the Promise PoolConnection type.

This is causing a bit of confusion on my end, and i'm not sure what the expected behaviour is? Should these be converted to PromisePoolConnections, or should the types be updated appropriately?

```typescript
import { PoolConnection } from 'mysql2'
import mysql from 'mysql2/promise'

// this is a PromisePool
const pool = mysql.createPool(options);

// PoolConnection here is NOT a PoolConnection from the Promise lib
// but typescript thinks it is, and so is giving an error.

pool.on('acquire', async (connection: PoolConnection) => {
    await connection.promise().beginTransaction();
  });
```

ashleybartlett:
This fixes a couple of type errors compared to the underlying code. 

This is likely a breaking change for some users as it would have required some working around to make it work before.

Should fix #3238 and #2667

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -7,9 +7,11 @@ import {
   FieldPacket,
   QueryOptions,
   ConnectionOptions,
+  Connection as CoreConnection,
   PoolOptions,
   PoolClusterOptions,
   Pool as CorePool,
+  PoolConnection as CorePoolConnection,
 } from './index.js';
 import { ExecutableBase as ExecutableBaseClass } from './typings/mysql/lib/protocol/sequences/promise/ExecutableBase.js';
 import { QueryableBase as QueryableBaseClass } from './typings/mysql/lib/protocol/sequences/promise/QueryableBase.js';
@@ -41,6 +43,8 @@ export interface PreparedStatementInfo {
 
 export interface Connection extends QueryableAndExecutableBase {
   config: ConnectionOptions;
+
+  connection: CoreConnection;
 
   threadId: number;
 
@@ -78,17 +82,16 @@ export interface Connection extends QueryableAndExecutableBase {
 
 export interface PoolConnection extends Connection {
   release(): void;
-  connection: Connection;
 }
 
-export interface Pool extends Connection {
+export interface Pool extends Pick<CorePool, 'execute' | 'query' | 'escape' | 'escapeId' | 'format'> {
   getConnection(): Promise<PoolConnection>;
 
   releaseConnection(connection: PoolConnection): void;
 
-  on(event: 'connection', listener: (connection: PoolConnection) => any): this;
-  on(event: 'acquire', listener: (connection: PoolConnection) => any): this;
-  on(event: 'release', listener: (connection: PoolConnection) => any): this;
+  on(event: 'connection', listener: (connection: CorePoolConnection) => any): this;
+  on(event: 'acquire', listener: (connection: CorePoolConnection) => any): this;
+  on(event: 'release', listener: (connection: CorePoolConnection) => any): this;
   on(event: 'enqueue', listener: () => any): this;
 
   end(): Promise<void>;
--- a/typings/mysql/lib/Connection.d.ts
+++ b/typings/mysql/lib/Connection.d.ts
@@ -335,7 +335,7 @@ export interface ConnectionOptions {
   jsonStrings?: boolean;
 }
 
-declare class Connection extends QueryableBase(ExecutableBase(EventEmitter)) {
+declare class BaseConnection extends QueryableBase(ExecutableBase(EventEmitter)) {
   config: ConnectionOptions;
 
   threadId: number;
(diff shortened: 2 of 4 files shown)


typings/mysql/lib/Connection.d.ts
```diff
@@ -431,4 +431,7 @@ declare class Connection extends QueryableBase(ExecutableBase(EventEmitter)) {
   sequenceId: number;
 }
 
-export { Connection };
+declare class Connection extends BaseConnection {
```
ashleybartlett, line 434: Changed this to fix the pool connection inheritance giving a type error on reimplementing the promise

promise.d.ts
```diff
@@ -78,17 +82,16 @@ export interface Connection extends QueryableAndExecutableBase {
...
-export interface Pool extends Connection {
+export interface Pool extends QueryableAndExecutableBase, Pick<CorePool, 'escape' | 'escapeId' | 'format'> {
   getConnection(): Promise<PoolConnection>;
 
   releaseConnection(connection: PoolConnection): void;
 
-  on(event: 'connection', listener: (connection: PoolConnection) => any): this;
-  on(event: 'acquire', listener: (connection: PoolConnection) => any): this;
-  on(event: 'release', listener: (connection: PoolConnection) => any): this;
+  on(event: 'connection', listener: (connection: CorePoolConnection) => any): this;
```
ashleybartlett, line 92: This is reflective of the actual implementation.

promise.d.ts
```diff
@@ -78,17 +82,16 @@ export interface Connection extends QueryableAndExecutableBase {
 
 export interface PoolConnection extends Connection {
   release(): void;
-  connection: Connection;
 }
 
-export interface Pool extends Connection {
+export interface Pool extends QueryableAndExecutableBase, Pick<CorePool, 'escape' | 'escapeId' | 'format'> {
```
ashleybartlett, line 87: Pool doesn't extend Connection in code, so this better reflects the actual implementation

promise.d.ts
```diff
@@ -42,6 +44,8 @@ export interface PreparedStatementInfo {
 export interface Connection extends QueryableAndExecutableBase {
   config: ConnectionOptions;
 
+  connection: CoreConnection;
```
ashleybartlett, line 47: This connection field is defined on the pool Connection, so this is now available everywhere pool connection is used.

typings/mysql/lib/Pool.d.ts
```diff
@@ -63,6 +63,13 @@ declare class Pool extends QueryableBase(ExecutableBase(EventEmitter)) {
 
   promise(promiseImpl?: PromiseConstructor): PromisePool;
 
+  format(sql: string, values?: any | any[] | { [param: string]: any }): string;
```
ashleybartlett, line 66: Copied from connection type, as these exist here in code.

typings/mysql/lib/PoolConnection.d.ts
```diff
@@ -1,7 +1,7 @@
-import { Connection } from './Connection.js';
+import { Connection, BaseConnection } from './Connection.js';
 import { Pool as PromisePool } from '../../../promise.js';
 
-declare class PoolConnection extends Connection {
+declare class PoolConnection extends BaseConnection {
```
ashleybartlett, line 4: Fixed inheritance type error for the promise function implementation.

wellwelwel:
Thanks, @ashleybartlett!

The inheritance from base connections after #3081 makes perfect sense, but I'll keep the #3273 issue as related due to a unexpected breaking change when using `instanceof`.

I'm planning to run some regression tests in your changes just to make sure we wouldn't imply an additional breaking change.

ashleybartlett:
I think that breaking change makes a lot of sense, as it was done under the hood, but i'm not sure if it's something that can be fixed properly.

Given it wasn't breaking existing checks, I can revert that change for the base type for now, and it can be saved for a breaking change release.

ashleybartlett:
> I'll keep the #3273 issue as related due to an unexpected breaking change when using `instanceof`.

Revisiting this, I realised that `instanceof` would be a runtime check, not a typescript one. For this, I would ensure Typescript is accurately describing the Javascript code it's referencing.

klingenm:
@wellwelwel What investigation is it you are looking for? Changing the types can not have any runtime effect and the PR currently only fixes a type error to match the actual implementation.

I was about to open a PR with the same fix, as I want to use the Promise api for connection handling but I want to access the base connection so I can use streaming responses. Now I have to typecast the promise connection type to the base connection type.

wellwelwel:
Hey @ashleybartlett, the problem is not with this PR itself, but with issue #3273. Changing the types from `Connection` to `Base...Connection` or `Core...Connection` will force a breaking change for everyone who performs any checks or depends on the `Connection` type.

The mistake is probably in the class names in the base code and needs to be fixed there first, before going back to this _PR_.

I can't confirm if it would be so simple to resolve by just renaming the classes. It would be necessary to check the instances to address the problem at its root (that's the real investigation).

Issue #3028, opened 2024-09-09
Improve ConnectionOptions ssl documentation


jrr:
I struggled a bit today connecting to an Azure mysql server. I wanted to use TLS, but didn't want to supply certificates manually. (counting on the host's OS to already trust the remote server)

I got it working, but in hindsight I have a couple suggestions to flatten the speed bump and help folks in the future:
 - Allow `ssl: true` for people like me that want to turn it on but don't need to customize it
 - Make it really clear in the tooltip that:
   - An empty object `ssl: {}` means "Use SSL".
   - An absent field (no `ssl:`) means "Don't use SSL".

pkuczynski:
Since I was struggling myself, I put up a short documentation: #3384. @jrr any feedback?

If this goes through I might work on a patch with `ssl: true`.

PR #3387, opened 2025-02-08
feat: allow to specify `ssl: true` in connection options

pkuczynski:
Follow up for #3384, inspired by #3028


wellwelwel:
Thanks, @pkuczynski!

I'll review your _PR_ as soon as possible. @sidorares, feel free to review it first 🤝

---

@pkuczynski, the website has some customized components. Normally I don't suggest it to the contributors not to make it complicated _(I put it in a personal queue to adjust it later)_. But to prevent confusion, I'm adding them as optional tips here 🙋🏻‍♂️

### ExternalCodeEmbed

> https://sidorares.github.io/node-mysql2/docs/contributing/website#externalcodeembed

> [!TIP]
> With this component, plus the `y` tip in https://github.com/sidorares/node-mysql2/commit/a446facc754bb1c62ab01e12ffd8021b5621eb8e#commitcomment-152332163, we can then dynamically demonstrate all the options in the documentation.
> 
> You can see a practical example here:
> - https://sidorares.github.io/node-mysql2/docs/examples/connections/create-connection#connectionoptions

### History

> https://sidorares.github.io/node-mysql2/docs/contributing/website#history

> [!TIP]
> Here we can refer to the next `semver` version ([**3.13.0**](https://github.com/sidorares/node-mysql2/pull/3377)), which made it possible to use the `ssl` property with boolean values.

### Stability

> https://sidorares.github.io/node-mysql2/docs/contributing/website#stability

> [!TIP]
> In the documentation, instead of `SSL Profile (deprecated)`, we can remove the parenthesis and replace it with a stability badge, for example: `<Stability level={0} />`.
> 
> You can see a practical example in:
> - https://sidorares.github.io/node-mysql2/docs/documentation/typescript-examples#okpacket

All these suggestions are optional and you can find the documentation contribution guide at [sidorares.github.io/node-mysql2/docs/contributing/website](https://sidorares.github.io/node-mysql2/docs/contributing/website) 💡

wellwelwel:
@pkuczynski, sorry for the delay.

The `ssl` value is set in:

https://github.com/sidorares/node-mysql2/blob/36349eb70d6a4c900b958dfc55c025453d264d9e/lib/connection_config.js#L144-L147

I think it would be beneficial to keep the logic contained in the same place _(especially as it's easier to maintain in the long-term)_, what do you think?

PR #3451, opened 2025-03-04
Add RegExp support to PoolCluster

### LINKED ISSUE #503: node-mysql pool cluster regex and wildcard support
How soon can support for wildcards and regexp be added to this module.

see https://github.com/mysqljs/mysql/pull/1572

my code depends on it

### LINKED ISSUE #3450: PoolCluster.getConnection(id) can return incorrect PoolCluster
When calling `PoolCluster.getConnection(id)` with an invalid `id` it may return a connection that does not match exactly. 

The last character of `id` is removed, and any connection ids that start with the new `id-last character` will be returned. The expected behavior (and the behavior in mysqljs/mysql) is that a `POOL_NOEXIST` error should be thrown.

```javascript
import mysql from "mysql2/promise";

(async () => {
  const poolCluster = mysql.createPoolCluster();
  poolCluster.add("129", {
    user: "root",
    database: "exampledb",
  });
  poolCluster.add("12", {
    user: "root",
    database: "exampledb",
  });

  let connection = await poolCluster.getConnection("125");
  // connection may be ID 129 or ID 12. Should be POOL_NOEXIST error.
  await connection.query("SELECT 1");

  poolCluster.end();
})();
```

uPaymeiFixit:
mysqljs/mysql has been updated to support regular expressions in PoolCluster. This brings most of that code to mysql2. 

Closes #3450 
Closes #503

### DIFF
--- a/lib/pool_cluster.js
+++ b/lib/pool_cluster.js
@@ -37,6 +37,23 @@ const getMonotonicMilliseconds = function () {
   return Math.floor(ms);
 };
 
+const isRegExp = function (val) {
+  return typeof val === 'object'
+    && Object.prototype.toString.call(val) === '[object RegExp]';
+}
+
+const patternRegExp = function (pattern) {
+  if (isRegExp(pattern)) {
+    return pattern;
+  }
+
+  var source = pattern
+    .replace(/([.+?^=!:${}()|\[\]\/\\])/g, '\\$1')
+    .replace(/\*/g, '.*');
+
+  return new RegExp('^' + source + '$');
+}
+
 class PoolNamespace {
   constructor(cluster, pattern, selector) {
     this._cluster = cluster;
@@ -256,20 +273,12 @@ class PoolCluster extends EventEmitter {
     let currentTime = 0;
     let foundNodeIds = this._findCaches[pattern];
 
-    if (typeof this._findCaches[pattern] === 'undefined') {
-      if (pattern === '*') {
-        // all
-        foundNodeIds = this._serviceableNodeIds;
-      } else if (this._serviceableNodeIds.indexOf(pattern) !== -1) {
-        // one
-        foundNodeIds = [pattern];
-      } else {
-        // wild matching
-        const keyword = pattern.substring(pattern.length - 1, 0);
-        foundNodeIds = this._serviceableNodeIds.filter((id) =>
-          id.startsWith(keyword)
-        );
-      }
+    if (foundNodeIds === undefined) {
+      let expression = patternRegExp(pattern);
+
+      foundNodeIds = this._serviceableNodeIds.filter((id) =>
+        id.match(expression)
+      );
     }
 
     this._findCaches[pattern] = foundNodeIds;


lib/pool_cluster.js
```diff
@@ -37,6 +37,23 @@ const getMonotonicMilliseconds = function () {
...
 
+const isRegExp = function (val) {
+  return typeof val === 'object'
+    && Object.prototype.toString.call(val) === '[object RegExp]';
+}
+
+const patternRegExp = function (pattern) {
+  if (isRegExp(pattern)) {
+    return pattern;
+  }
```
wellwelwel, line 48:
```suggestion
const patternRegExp = function (pattern) {
  if (pattern instanceof RegExp) {
    return pattern;
  }
```

uPaymeiFixit: No. I thought about it, but in the moment wanted it to be similar to mysqljs/mysql. I prefer instanceof, let's do it.

wellwelwel:
Thanks, @uPaymeiFixit 🙋🏻‍♂️

Do you think it would be possible to include a failing test without your changes? I would try an alternative similar to the original (without using `RegExp`) before creating regular expressions dynamically.

wellwelwel:
About the lint errors, you can run `npm run lint:fix` 🙋🏻‍♂️

uPaymeiFixit:
@wellwelwel I'm not sure if I understand what you mean by a failing test without my changes. 

Maybe a useful test would be checking that if you run `PoolCluster.add('a', poolConfig)` and then run `PoolCluster.getConnection('aa')` a `POOL_NOEXIST` error is thrown?

wellwelwel:
> I'm not sure if I understand what you mean by a failing test without my changes.

This _PR_ includes both a feature (support for `RexExp` selectors on Pool Cluster) and a fix (ensure the start/end of a pattern).

It would be great to include a test covering the `RegExp` pattern and other one ensuring the selector respect the correct pattern, similar to what you did in #3450:

```js
import mysql from "mysql2/promise";

(async () => {
  const poolCluster = mysql.createPoolCluster();
  poolCluster.add("129", {
    user: "root",
    database: "exampledb",
  });
  poolCluster.add("12", {
    user: "root",
    database: "exampledb",
  });

  let connection = await poolCluster.getConnection("125");
  // connection may be ID 129 or ID 12. Should be POOL_NOEXIST error.
  await connection.query("SELECT 1");

  poolCluster.end();
})();
```

- But using an test approach _(assert)_.

---

> Maybe a useful test would be checking that if you run `PoolCluster.add('a', poolConfig)` and then run `PoolCluster.getConnection('aa')` a `POOL_NOEXIST` error is thrown?

That's a good idea 🙋🏻‍♂️

uPaymeiFixit:
@wellwelwel I think I've got the fix-related test down, but I'm struggling a bit with the RegEx test. 

I've created a new file, `test/unit/pool-cluster/test-regex-pattern.test.cjs` with this contents: 

```javascript
'use strict';

const { assert } = require('poku');
const common = require('../../common.test.cjs');
const process = require('node:process');

// TODO: config poolCluster to work with MYSQL_CONNECTION_URL run
if (`${process.env.MYSQL_CONNECTION_URL}`.includes('pscale_pw_')) {
  console.log('skipping test for planetscale');
  process.exit(0);
}

const cluster = common.createPoolCluster();

const poolConfig = common.getConfig();
cluster.add('SLAVE1', poolConfig);

cluster.of(/SLAVE[12]/).getConnection((error, connection) => {
  assert.equal(connection._clusterId, 'SLAVE1', 'should match regex');
});

cluster.end();
```

And it's throwing this error: 

```
  uncaughtException Error: Can't add new command when connection is in closed state
      at PoolConnection._addCommandClosedState (/Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/base/connection.js:159:17)
      at PoolConnection.end [as _realEnd] (/Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/base/connection.js:904:26)
      at Pool.end (/Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/base/pool.js:124:18)
      at PoolCluster._removeNode (/Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/pool_cluster.js:360:17)
      at PoolCluster._increaseErrorCount (/Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/pool_cluster.js:318:10)
      at /Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/pool_cluster.js:344:14
      at /Users/joshuagibbs/git/uPaymeiFixit/node-mysql2/lib/base/pool.js:37:37
      at process.processTicksAndRejections (node:internal/process/task_queues:77:11) {
    fatal: true
  }
```

Any ideas?

wellwelwel:
> And it's throwing this error:

This error is due to the callback not waiting for execution to end the connection. You can end the connection before starting the assertions or use a promise-based connection 🙋🏻‍♂️

uPaymeiFixit:
I've added what I think are valid tests. I'm not familiar with poku or the folder structure for these tests, so it's entirely possible they're in the wrong place. Please let me know if they could be better organized.

wellwelwel:
@uPaymeiFixit, `assert.fails` trigger `process.exit(1)` instead of throw an error.

> I'm not familiar with poku or the folder structure for these tests
 
**Poku** API follows the same usage as it is for **Node.js** (https://poku.io/docs/documentation/assert), for example:

- **Node.js**

```js
import assert from 'node:assert';
import test from 'node:test';

test('My Test', () => {
  assert(true);
  assert.ok(true);
});
```

- **Poku** 🐷

```js
import { test, assert } from 'poku';

test('My Test', () => {
  assert(true);
  assert.ok(true);
});
```

wellwelwel:
I've seen that `mysqljs/mysql` uses this approach:

```js
  const source = pattern
    .replace(/([.+?^=!:${}()|[\]/\\])/g, '\\$1')
    .replace(/\*/g, '.*');

  return new RegExp(`^${source}$`);
```

But I'd like to test first if it's possible to get the same result without building `RegExp` dynamically. However, the current state of this _PR_ looks good to me 🙋🏻‍♂️

uPaymeiFixit:
Off the top of my head I can't think of a clean way to achieve the same result without building the regular expression manually. This comes to mind, but it's not fully compatible with mysqljs/mysql. Are you more worried about performance, or clarity? 

```patch
--- lib/pool_cluster.js
+++ lib/pool_cluster.js
@@ -269,7 +269,10 @@ class PoolCluster extends EventEmitter {
     let foundNodeIds = this._findCaches[pattern];
 
     if (foundNodeIds === undefined) {
-      const expression = patternRegExp(pattern);
+      let expression = pattern;
+      if (typeof pattern === 'string' && pattern.at(-1) === '*') {
+        expression = patternRegExp(pattern);
+      }
 
       foundNodeIds = this._serviceableNodeIds.filter((id) =>
         id.match(expression)
```

wellwelwel:
> Are you more worried about performance, or clarity?

Performance is a consideration, but my concern is particularly about security against **ReDoS**.

In this project you can see how I usually handle recursions to create safe dynamics "matches" in a similar result of a dynamic `RegExp`:

https://github.com/wellwelwel/jsonc.min/blob/main/src/index.ts

But it's not a requirement, it's more of a curiosity. I'll try something and I'll come back with some idea or make sure there's no **ReDoS** possibility.

wellwelwel:
@uPaymeiFixit. I tried something and, performance-wise, I suppose it's better, but complexity without using `RexExp` dynamically definitely doesn't seem to be a good option:

```js
_findNodeIds(pattern, includeOffline) {
  let currentTime = 0;
  let foundNodeIds = this._findCaches[pattern];

  if (foundNodeIds !== undefined) {
    if (includeOffline) {
      return foundNodeIds;
    }

    return foundNodeIds.filter((nodeId) => {
      const node = this._getNode(nodeId);
      if (!node._offlineUntil) return true;

      currentTime = currentTime || getMonotonicMilliseconds();
      return node._offlineUntil <= currentTime;
    });
  }

  if (pattern instanceof RegExp) {
    foundNodeIds = this._serviceableNodeIds.filter((id) => pattern.test(id));
  } else if (pattern === '*') {
    foundNodeIds = this._serviceableNodeIds;
  } else if (this._serviceableNodeIds.indexOf(pattern) !== -1) {
    foundNodeIds = [pattern];
  } else if (pattern.indexOf('*') !== -1) {
    const parts = pattern.split('*');
    foundNodeIds = this._serviceableNodeIds.filter((id) => {
      let idToCheck = id;

      if (parts[0] && !idToCheck.startsWith(parts[0])) {
        return false;
      }

      idToCheck = idToCheck.slice(parts[0].length);

      for (let i = 1; i < parts.length - 1; i++) {
        const part = parts[i];
        if (part === '') continue;

        const index = idToCheck.indexOf(part);
        if (index === -1) return false;

        idToCheck = idToCheck.slice(index + part.length);
      }

      const lastPart = parts[parts.length - 1];
      if (lastPart && !idToCheck.endsWith(lastPart)) {
        return false;
      }

      return true;
    });
  } else {
    foundNodeIds = [];
  }

  this._findCaches[pattern] = foundNodeIds;

  if (includeOffline) {
    return foundNodeIds;
  }

  return foundNodeIds.filter((nodeId) => {
    const node = this._getNode(nodeId);

    if (!node._offlineUntil) {
      return true;
    }

    if (!currentTime) {
      currentTime = getMonotonicMilliseconds();
    }

    return node._offlineUntil <= currentTime;
  });
}
```

On the other hand, this sanitization has been used for years by `mysqljs/mysql` without problems, so I think it's a safe alternative (and much simpler).

Please, let me know what do you think. I'm fine to merge in the current state.

- I personally don't think the complexity is worth the cost of performance in this particular case.

uPaymeiFixit:
I think the only way ReDoS is possible is if the user provides an instanceof RegExp as the pattern (because if they provide a string all of the regex special characters are escaped), in which case in my opinion it's up to them to sanitize the expression. 

I can appreciate the effort that went into experimenting with that alternative, but I agree it feels a little complex. _Maybe_ it could be a little easier if some of that logic gets broken out into another function, but I think my vote is still dynamic regular expression.

wellwelwel:
Coming back, the concern about security remains. The problem is not the use of Regex as an instance. In this case, the responsibility for sanitizing and validating the regular expression lies with the user.

My concern is with the process of dynamically creating strings into regular expressions, especially considering users rely on this value being assigned to a string and have never been introduced to pre-validating the value as a `RegExp`, as this is a new feature being included:

- https://github.com/sidorares/node-mysql2/blob/3bcd394187f6f9f2871e765667230f8bfabd5504/typings/mysql/lib/PoolCluster.d.ts#L65-L79

Still on sanitization, the current logic only normalizes `RegExp`'s _"magic"_ characters to guarantee their functionality and doesn't actually sanitize them:

```js
pattern
    .replace(/([.+?^=!:${}()|[\]/\\])/g, '\\$1')
    .replace(/\*/g, '.*')
```

- The normalization step leads **CodeQL** misunderstanding that the data is in fact being sanitized.

---

> in which case in my opinion it's up to them to sanitize the expression.

I understand it's unusual, but here are two RCEs fixed with gravity `9.8` because of a situation which, in theory, we expect to be the responsibility not of the final user, but of the developer:

- #2572
- #2608

uPaymeiFixit:
> The normalization step leads CodeQL misunderstanding that the data is in fact being sanitized.

If that's the case, I agree. Because it certainly has me fooled. 

Just to help me understand, can give an example of a string which would _not_ be properly sanitized by `.replace(/([.+?^=!:${}()|[\]/\\])/g, '\\$1')`? I wonder if there's a regular expression which _could_ properly sanitize it?

wellwelwel:
> If that's the case, I agree. Because it certainly has me fooled.

My previous comment indirectly implied that normalization is unsafe, but what I meant was about nullifying the effects of all special characters. Sorry, that wasn't the intention 🙋🏻‍♂️

In fact, normalization is almost complete. I assume it would be enough to add the `*` characters to the escaped characters and removing the `.*` replacement. This way, we avoid unexpected behavior by those who used a selector as a string:

wellwelwel:
@sidorares, do you know why [mysqljs/mysql](https://github.com/mysqljs/mysql/blob/dc9c152a87ec51a1f647447268917243d2eab1fd/lib/PoolCluster.js#L277) allows the `.*` for string conversion?

PR #3456, opened 2025-03-05
refactor: upgrade ESLint and migrate its plugins to `v9`

wellwelwel:
> Here's the popular _PR_ with thousands of changed lines 😅

There is **no change** in the actual files, except for **ESLint** applying the auto fixes, including **Prettier**'s rules _(nothing else)_, while migrating **ESLint** and its respective plugins to their latest versions.

For years, `lint-stage` never performed formatting before commits, which is why so many files were affected. I'll open a separate PR removing this dependency since it's not actually used in practice.

Also, you can follow all breaking changes from **ESLint** in [**Migrate to v9.x**](https://eslint.org/docs/latest/use/migrate-to-9.0.0) documentation.

---

### CodeQL

It's interesting to note that **CodeQL** will notify about two warnings, since a comma has been removed from the line in question for each warning.

If this is indeed a problem to be addressed, I don't see this _PR_ as the ideal place for it, since it's exclusively for _lint_ and _format_ rules and its apply.

<blockquote>

However, **Copilot** suggested it:

```js
const charMap = {
  '<': '\\u003C',
  '>': '\\u003E',
  '/': '\\u002F',
  '\\': '\\\\',
  '\b': '\\b',
  '\f': '\\f',
  '\n': '\\n',
  '\r': '\\r',
  '\t': '\\t',
  '\0': '\\0',
  '\u2028': '\\u2028',
  '\u2029': '\\u2029'
};

function escapeUnsafeChars(str) {
  return str.replace(/[<>\b\f\n\r\t\0\u2028\u2029]/g, x => charMap[x]);
}
```

- It might be worth checking a dedicated PR for this.

</blockquote>


wellwelwel:
I also took care to check each PR opened individually:

From 49 valid PRs _(not bots)_:

- [38 inevitably need rebasing](https://github.com/sidorares/node-mysql2/pulls?q=is:open+is:pr+label:%22needs+rebase%22) due to various conflicts.
- In addition, [7 requires investigation](https://github.com/sidorares/node-mysql2/pulls?q=is:open+is:pr+label:%22needs+investigation%22) and [2 requires discussion of ideias and implementations](https://github.com/sidorares/node-mysql2/pulls?q=is:open+is:pr+label:%22needs+discussion%22+).
- Most of them are inactive for years.

I carefully create a label for each of them, which is where each link above leads.

PR #3464, opened 2025-03-10
Create webpack.yml

Hosampor:


wellwelwel:
Hi, @Hosampor. We don't use **Webpack** on **MySQL2**. However, in cases like these, we use dependency installations directly in `package.json`.

PR #3465, opened 2025-03-10
Create datadog-synthetics.yml

Hosampor:


wellwelwel:
Hi, @Hosampor. The same point of https://github.com/wellwelwel/poku/pull/964#issuecomment-2706861503 fits here.

Issue #3699, opened 2025-07-09
Invalid date using sequelize bind queries


steveetm:
When using bind, the binary parser fails to forward the `columnType` `readDateTimeString`. Sequelize assumes a correct datetime being returned and append timezone to a date, which results in an invalid date.

Standalone repro, using latest mysql2 and sequelize@^6

Proposed fix at https://github.com/sidorares/node-mysql2/blob/9dd020203b861d4d7f2d81ad48dd73e5c5016d0b/lib/parsers/binary_parser.js#L114

```js
return packet.readDateTimeString(parseInt(field.decimals, 10), ' ', field.columnType);
```
```json
"dependencies": {
    "sequelize": "^6.37.7",
    "mysql2": "^3.14.1"
  }
```
```js
const { Sequelize, DataTypes } = require('sequelize');

const DB_NAME = 'test';
const DB_USER = 'root';
const DB_PASSWORD = '';
const DB_HOST = '';
const DB_PORT = 3306;

const sequelize = new Sequelize(DB_NAME, DB_USER, DB_PASSWORD, {
  host: DB_HOST,
  port: DB_PORT,
  dialect: 'mysql',
  logging: console.log,
  logQueryParameters: true,
});

const TestModel = sequelize.define('TestModel', {
  id: {
    type: DataTypes.INTEGER,
    primaryKey: true,
    autoIncrement: true,
  },
  dateWithTime: {
    type: DataTypes.DATE,
    allowNull: false,
  },
  dateWithoutTime: {
    type: DataTypes.DATE,
    allowNull: false,
  }
}, {
  tableName: 'test_date_fields',
  timestamps: false,
});

async function runTest() {
  try {
    await sequelize.authenticate();
    console.log('Connection to the database has been established successfully.');

    await TestModel.sync({ force: true });
    console.log('Table created successfully.');

    const dateWithTime = '2025-08-01 12:34:56.789';

    const dateWithoutTime = '2025-08-01';

    const createdRecord = await TestModel.create({
      dateWithTime,
      dateWithoutTime,
    });
    console.log('Record created with ID:', createdRecord.id);

    const result1 = await sequelize.query('SELECT * FROM test_date_fields WHERE id = $id', {
      bind: { id: createdRecord.id },
      type: sequelize.QueryTypes.SELECT,
    });

    const result2 = await sequelize.query(`SELECT * FROM test_date_fields WHERE id = ${createdRecord.id}`, {
[...]

PR #3700, opened 2025-07-09
fix: pass columnType to readDateTimeString

steveetm:
See issue: https://github.com/sidorares/node-mysql2/issues/3699

### DIFF
--- a/lib/parsers/static_binary_parser.js
+++ b/lib/parsers/static_binary_parser.js
@@ -146,7 +146,9 @@ function getBinaryParser(fields, _options, config) {
                         ].includes(field.columnType)
                       ) {
                         return packet.readDateTimeString(
-                          parseInt(field.decimals, 10)
+                          parseInt(field.decimals, 10),
+                          ' ',
+                          field.columnType
                         );
                       }
 


steveetm:
I could not add test or run those as locally running them always stuck after a few seconds. If you can help me sorting that out I would like to add some tests as well.

wellwelwel:
Curiously, some scenarios wait for the datetime and others pass with just the date:

```
1) test/esm/integration/parsers/typecast-field-datetime.test.mjs

  🔬 typeCast field.datetime
  ✘ DATETIME
        File test/esm/integration/parsers/typecast-field-datetime.test.mjs:37:8
        Code ERR_ASSERTION
    Operator ==
    Actual:
    2025-07-09
    Expected:
    2025-07-09 00:00:00
```

steveetm:
> Curiously, some scenarios wait for the datetime and others pass with just the date:

The expection there is to `.query` and `.execute` results should match.
This failure is the exact same what I get prior this PR, and no matter what I try locally(tried different node, mysql versions and configurations) it is always passes with the PR(and fails without it, obviously)

I don't know how, but it seems like the change is not applied for those tests, but that shouldn't be possible.

wellwelwel:
@steveetm, thanks a lot! Today marks exactly one year since this error was reported on **Sequelize** (https://github.com/sequelize/sequelize/issues/17141#issuecomment-2219810865).

Issue #3596, opened 2025-05-22
Streaming pipeline errors block connections


CarlOlson:
This issue was written for `mysql`, but the problem also exists in `mysql2`. The `mysql` repository no long accepts issues from non-contributors... The patch attached is for `mysql`, but I believe the same changes can be made to `query.js` in `mysql2`.

I'll look into making a PR with tests when I have the time.

---

Running this script will hang the process when `pool.end` is called:

<details><summary>test.mjs</summary>

```js
import { createPool } from 'mysql'; // mysql or mysql2
import { pipeline } from 'node:stream/promises';

const pool = createPool({
  host: process.env.MYSQL_HOST,
  port: Number(process.env.MYSQL_PORT),
  user: process.env.MYSQL_USER,
  password: process.env.MYSQL_PASSWORD,
  database: process.env.MYSQL_DATABASE,
  poolConnectionLimit: 1,
  multipleStatements: true,
});

const promisify = (fn) =>
  new Promise((resolve, reject) =>
    fn((exn, conn) => (exn ? reject(exn) : resolve(conn)))
  );

const conn = await promisify((cb) => pool.getConnection(cb));

try {
  const stream = conn.query('select 1; select sleep(1);').stream({});
  await pipeline(stream, async function* (source) {
    for await (const chunk of source) {
      console.log('chunk:', chunk);
      throw new Error();
    }
  }).catch((err) => console.log('caught:', err));
} finally {
  conn.release();
  console.log('released');
}

// also hangs here
// console.log(await promisify((cb) => conn.query('select 1', cb)));

await promisify((cb) => pool.end(cb));

console.log('done');
```

</details> 

This was noticed during test cleanup from `pool.end`, but it will also prevent additional queries after `release` is called.

The issue is that the connection gets paused, but is never resumed. This prevents the protocol sequence from ever processing `Quit` and the necessary callbacks from `end`.

Here is a yarn patch I've made for **mysql** to fix the problem:

<details><summary>Details</summary>

```diff
diff --git a/lib/protocol/sequences/Query.js b/lib/protocol/sequences/Query.js
index b7632959b6b18eff0a239456c67b2696a5b931bd..d085e63dde4d320ac34a2b39d74386f661f66e7a 100644
--- a/lib/protocol/sequences/Query.js
+++ b/lib/protocol/sequences/Query.js
@@ -190,37 +190,50 @@ Query.prototype._sendLocalDataFile = function(path) {
 };
 
[...]

PR #3769, opened 2025-08-26
fix: destroy connection when stream errors

fenichelar:
When a stream errors (due to break, throw, etc.) the connection can remain paused indefinitely.

To reproduce, create table `test` with 500 rows and run the below. The script will never exit.

```ts
import { createPool } from 'mysql2';

const pool = createPool({
  database: 'test',
  host: '127.0.0.1',
  port: 3306,
  user: 'user',
});

const results = pool.query('select * from test;').stream();
for await (const result of results) {
  break;
}

pool.end();
```

See https://github.com/sidorares/node-mysql2/issues/3596

### DIFF
--- a/test/integration/connection/test-stream-error-destroy-connection.test.cjs
+++ b/test/integration/connection/test-stream-error-destroy-connection.test.cjs
@@ -1,42 +1,45 @@
 'use strict';
 
+const process = require('node:process');
+const { test, skip } = require('poku');
 const common = require('../../common.test.cjs');
 
-const connection = common.createConnection();
+if (process.env.MYSQL_USE_TLS === '1') skip('Skipping for SSL=1');
 
-connection.query(
-  [
-    'CREATE TEMPORARY TABLE `items` (',
-    '`id` int(11) NOT NULL AUTO_INCREMENT,',
-    '`text` varchar(255) DEFAULT NULL,',
-    'PRIMARY KEY (`id`)',
-    ') ENGINE=InnoDB DEFAULT CHARSET=utf8',
-  ].join('\n'),
-  (err) => {
-    if (err) {
-      throw err;
+test('Ensure stream ends in case of error', async () => {
+  const connection = common.createConnection();
+
+  connection.query(
+    [
+      'CREATE TEMPORARY TABLE `items` (',
+      '`id` int(11) NOT NULL AUTO_INCREMENT,',
+      '`text` varchar(255) DEFAULT NULL,',
+      'PRIMARY KEY (`id`)',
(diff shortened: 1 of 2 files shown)


wellwelwel:
Thanks, @fenichelar! I'll test locally by forcing an error and check the feasibility of including your example as a test 🙋🏻‍♂️

fenichelar:
@wellwelwel test added.

test/integration/connection/test-stream-error-destroy-connection.test.cjs
```diff
@@ -0,0 +1,42 @@
...
+  for await (const row of rows) {
+    break;
+  }
+  setTimeout(() => {
+    throw new Error('Connection remains open after stream error');
+  }, 1000).unref();
+  connection.end();
+}
+
+run();
```
wellwelwel, line 42:
```suggestion
'use strict';

const process = require('node:process');
const { test, skip } = require('poku');
const common = require('../../common.test.cjs');

if (process.env.MYSQL_USE_TLS === '1') skip('Skipping for SSL=1');

test('Ensure stream ends in case of error', async () => {
  const connection = common.createConnection();

  connection.query(
    [
      'CREATE TEMPORARY TABLE `items` (',
      '`id` int(11) NOT NULL AUTO_INCREMENT,',
      '`text` varchar(255) DEFAULT NULL,',
      'PRIMARY KEY (`id`)',
      ') ENGINE=InnoDB DEFAULT CHARSET=utf8',
    ].join('\n'),
    (err) => {
      if (err) {
        throw err;
      }
    }
  );

  for (let i = 0; i < 100; i++) {
    connection.execute('INSERT INTO items(text) VALUES(?)', ['test'], (err) => {
      if (err) {
        throw err;
      }
    });
  }

  const rows = connection.query('SELECT * FROM items').stream();

  // eslint-disable-next-line no-unused-vars
  for await (const _ of rows) break;

  setTimeout(() => {
    throw new Error('Connection remains open after stream error');
  }, 1000).unref();

  connection.end();
});
```

wellwelwel (changes requested):

test/integration/connection/test-stream-error-destroy-connection.test.cjs
wellwelwel: I tested it with and without your change and confirmed the wrong behavior. I just made some adjustments to the test 🙋🏻‍♂️

wellwelwel:
Thanks again, @fenichelar! It's already available in `mysql2@canary` and soon in the `latest` version too 🤝

PR #3775, opened 2025-08-27
fix(stream): resume connection when stream errors or is destroyed

### LINKED ISSUE #1168: Query streamed with .stream() does not close when the stream has an error
When streaming a query with pool.query.stream(), if .destroy() is called on the stream, the 'close' event does not propagate upstream and close the underlying connection, leaving dead connections in the pool which are still executing.

As such it seems best to forcibly close the connection as I do not think mysql has a way to cancel a query while leaving the connection open.

Please see quick fix below, I have not had time yet to write a test or to confirm that this is the proper way to close the connection, or whether it might have other side effects:
https://github.com/CloudQuote/node-mysql2/commit/ac9236c7638974c100c421a79ec516f5df3fa156

### LINKED ISSUE #1847: Connection errors are not propagated to the query stream

### LINKED ISSUE #3596: Streaming pipeline errors block connections

wellwelwel:
This _PR_ continues #3769 work and fixes #3596. It also seems to resolve #1847 and resolve #1168, ensuring that `destroy` and error events are properly issued (not using _`try-catch`_, but _`stream.on('error', () => {})`_ approach, which is the expected pattern for callbacks).

---

In #3769, it destroys the connection on error to ensure that the stack is not blocked, but this approach can be aggressive, as a query error shouldn't destroy the connection.

In this _PR_, the approach is to destroy the current stream itself on error and emits the appropriate events, ensuring its functionality and allowing users to catch errors that were not previously emitted, as reported in #1847.

---

### Before this _PR_

The test that checks the events is not reached and the script is silently ended prematurely:

Screenshot 2025-08-27 at 05 37 42

### After this _PR_

Now all tests are reached:

Screenshot 2025-08-27 at 05 35 54

---

> [!NOTE]
>
> Thanks @fenichelar, @CarlOlson, and @dsech. Your examples and solutions helped us achieve a better understanding of the error 🤝

### DIFF
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -290,7 +290,6 @@ class Query extends Command {
         this.removeListener('end', onEnd);
         this.removeListener('error', onError);
         this.removeListener('error', onError);
-
         /* eslint-enable no-use-before-define */
 
         cb(err);
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -271,33 +271,51 @@ class Query extends Command {
   }
 
   stream(options) {
-    options = options || {};
+    options = options || Object.create(null);
     options.objectMode = true;
-    const stream = new Readable(options);
-    stream._read = () => {
-      this._connection && this._connection.resume();
-    };
-    this.on('result', (row, resultSetIndex) => {
-      if (!stream.push(row)) {
-        this._connection.pause();
-      }
-      stream.emit('result', row, resultSetIndex); // replicate old emitter
+
+    const stream = new Readable({
+      ...options,
+      read: () => {
+        this._connection && this._connection.resume();
+      },
+      destroy: (err, cb) => {
+        this._connection && this._connection.resume();
+
+        cb(err); // Pass on any errors
+      },
     });
-    this.on('error', (err) => {
-      stream.emit('error', err); // Pass on any errors
-    });
-    this.on('end', () => {
-      stream.push(null); // pushing null, indicating EOF
-    });
-    this.on('fields', (fields) => {
-      stream.emit('fields', fields); // replicate old emitter
-    });
-    stream.on('end', () => {
+
+    stream.once('end', () => {
       stream.emit('close');
     });
-    stream.on('error', () => {
-      this._connection && this._connection.destroy();
+
+    this.on('result', (row, index) => {
+      if (stream.destroyed) return;
+
+      if (!stream.push(row)) {
+        this._connection && this._connection.pause();
+      }
+
+      stream.emit('result', row, index); // replicate old emitter
     });
+
+    this.on('fields', (fields) => {
+      if (stream.destroyed) return;
+
+      stream.emit('fields', fields); // replicate old emitter
+    });
+
+    this.on('end', () => {
+      if (stream.destroyed) return;
+
+      stream.push(null); // pushing null, indicating EOF
+    });
+
+    this.on('error', (err) => {
+      stream.destroy(err);
+    });
+
     return stream;
   }
 
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -282,6 +282,13 @@ class Query extends Command {
       destroy: (err, cb) => {
         this._connection && this._connection.resume();
 
+        /* eslint-disable no-use-before-define */
+        this.removeListener('result', onResult);
+        this.removeListener('fields', onFields);
+        this.removeListener('end', onEnd);
+        this.removeListener('error', onError);
+        /* eslint-enable no-use-before-define */
+
         cb(err); // Pass on any errors
       },
     });
@@ -290,7 +297,7 @@ class Query extends Command {
       stream.emit('close');
     });
 
-    this.on('result', (row, index) => {
+    const onResult = (row, index) => {
       if (stream.destroyed) return;
 
       if (!stream.push(row)) {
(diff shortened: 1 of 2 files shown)


lib/commands/query.js
```diff
@@ -271,33 +271,68 @@ class Query extends Command {
...
+      },
+      destroy: (err, cb) => {
+        this._connection && this._connection.resume();
+
+        /* eslint-disable no-use-before-define */
+        this.removeListener('result', onResult);
+        this.removeListener('fields', onFields);
+        this.removeListener('end', onEnd);
+        this.removeListener('error', onError);
+        /* eslint-enable no-use-before-define */
```
CarlOlson, line 292: I think this is a valid complaint from `eslint`. These are not `var` and will not be hoisted.

wellwelwel: > I think this is a valid complaint from `eslint`. These are not `var` and will not be hoisted.

In this case, if we declare the event methods later, the stream is declared earlier, and vice versa 😂

Regarding the hoist issue, I believe it is not applicable since the `destroy` method will be called after the variables are declared, but a possible approach would be to pass the stream as a parameter to solve this concern.

wellwelwel: Even when passing the stream as a parameter, the concept remains the same: the references conflict when declared before or after in the same way. For example:

```js
const onError = (stream) => (err) => {
  stream.destroy(err);
};

const stream = new Readable({
  destroy: (err, cb) => {
    this.removeListener('error', onErrorHandler);
  },
});

const onErrorHandler = onError(stream);

this.on('error', onErrorHandler);
```

Screenshot 2025-08-27 at 07 23 16

wellwelwel: Solved! 🌞✨

I didn't want to overwrite the private stream methods, but it was the most viable and practical alternative.

Issue #3148, opened 2024-10-22
Pool connections are not gracefully closed once idle connection cleanup occurs.


dtimofeev:
Hi,
When Pool._removeIdleTimeoutConnections() is executed and an idle connection is selected for removal (regardless of the conditions) the following code is executed:
https://github.com/sidorares/node-mysql2/blob/master/lib/pool.js#L208
This in turn removes the connection from the pool and directly (non-gracefully) closes it by calling PoolConnection.destroy() instead of PoolConnection.end().
https://github.com/sidorares/node-mysql2/blob/master/lib/pool_connection.js#L50
As a result the connection is closed without notifying the server in any way.

Because of the above the SQL(MariaDB) server is flooded with warnings like:
`Aborted connection 123 to db: 'dbname' user: 'dbuser' host: '127.0.0.1' (Got an error reading communication packets)`

A possible solution might be to change the current PoolConnection.end() implementation with:
```
end() {
    this._removeFromPool();
    super.end();
}
```
but I'm not sure if this solves the issue correctly.

Looking forward to your thoughts on this.

dtimofeev:
Reproduce script:
```
const mysql = require("mysql2/promise");

(function () {
  const pool = mysql.createPool({
    host: "localhost",
    user: "root",
    password: "random",
    database: "tests",
    idleTimeout: 1000,
    connectionLimit: 5,
    // this has to be lower than connectionLimit in order for cleanup process to start at all
    maxIdle: 4,
  });

  setInterval(() => {
    pool.query("SELECT 1");
  }, 1000);
})();
```

dtimofeev:
The issue also occurs in latest MySQL:
```
docker run  --env=MYSQL_ROOT_PASSWORD=random -p 3306:3306 -d mysql:latest --log_error_verbosity=3
docker logs -f <ID from the above command>
```

```
[Note] [MY-010914] [Server] Aborted connection 8 to db: 'unconnected' user: 'root' host: '172.17.0.1' (Got an error reading communication packets).
```

dstankovd:
Hello,

I have submitted a new pull request (https://github.com/sidorares/node-mysql2/pull/3180) that addresses this issue with minimal changes.

In my view, an improved long-term solution would involve removing the deprecation warning, as it has been in place for some time, and implementing proper end functionality in pool_connection without relying on _realEnd. If you would like me to work on this enhancement, I would be happy to update the pull request accordingly.

Best regards,
Dimitar

[...]

PR #3180, opened 2024-11-04
fix: gracefully close pool connections sidorares#3148

dstankovd:

PR #3776, opened 2025-08-28
feat: gracefully end pool connections #3148

dstankovd:
Related to:  https://github.com/sidorares/node-mysql2/issues/3148

When Pool._removeIdleTimeoutConnections() is executed and an idle connection is selected for removal (regardless of the conditions) the stream is just ended without notifying the MySQL server. This leads to warnings being logged in the SQL error log.

This pull requests add additional config option that would gracefully close the pool connection.


wellwelwel:
Thanks, @dstankovd! After the `test/integration/test-pool-end.test.cjs` test, the connection is not being closed. By order of execution, it's probably the [`test-pool-release-idle-connection-replicate.test.cjs`](https://github.com/sidorares/node-mysql2/blob/master/test/integration/test-pool-release-idle-connection-replicate.test.cjs) test.

dstankovd:
> Thanks, @dstankovd! After the `test/integration/test-pool-end.test.cjs` test, the connection is not being closed. By order of execution, it's probably the [`test-pool-release-idle-connection-replicate.test.cjs`](https://github.com/sidorares/node-mysql2/blob/master/test/integration/test-pool-release-idle-connection-replicate.test.cjs) test.

I'm trying to figure out the hanging test runs, but I'm having trouble reproducing this issue.

On my local machine, the full test suite runs successfully against MySQL v8.3.0. I also tried using deno with the setup from this [failing job](https://github.com/sidorares/node-mysql2/actions/runs/17290491059/job/49076674708?pr=3776), and all tests completed without hanging. Since the tests pass in a replicated environment, do you have any thoughts on what could be causing this flakiness in the CI?

As a side note, I did find some tests that hangwhen using MariaDB 11.4. I believe this is an unrelated issue that's not connected to this PR, and I'll look into it separately.

trungdq88:
@dstankovd @wellwelwel first of all thank you for your time in maintaining this project!

After reading on this issue, I think the previous PR https://github.com/sidorares/node-mysql2/pull/3180/files is a much better solution. Much cleaner and avoid another confusing `gracefulEnd` flag (especially the flag changes the behavior of the `end` method as [described here](https://github.com/sidorares/node-mysql2/pull/3776/files#diff-2fbd1c14f5a2c9a5cb8d6a17d8e2f4e83afd75d62d8f5aa481aa82d8a9628484R37-R38)).

Any chance we can get @sidorares approval on that instead? My database log is also flooded with the "aborted connection" error messages so I would love to have this issue fixed.

Thanks!

wellwelwel:
@dstankovd, just checking and confirming that the error is in the new test _(test/integration/test-pool-release-idle-connection-graceful-end.test.cjs)_.

@trungdq88, I'm fine with either solution, but since the previous one removes a depreciation, it's interesting to consider @sidorares's point, especially for architectural reasons. Whereas in this current solution, since it's a new flag, it is more flexible to be merged before.

dstankovd:
Hey @wellwelwel, finally managed to reproduce the issue locally and I've reorganized the new tests a bit and changed them to depend on process exit to validate the outcome, that appears to have fixed the test execution for me. Hopefully the CI will pass too. :)

As for the @trungdq88's suggestion - I also prefer the solution in the previous pull request but, if the compatibility with mysql is a must, then it's a no-go. Generally, I don't think that changing the `end` method in the pool_connection would have led to any negative impact since it has been deprecated for a few years now and it does come with a big warning, so it's unlikely that anyone uses it.

Either way, for now I'm keen to have the logs clear from the "aborted connection" warning so I'd gladly use the work-around implementation in this pull request. I can open a new issue to further discuss the possibility of actually changing the `end` method.

wellwelwel:
Updating just to bring in the #1752 changes.

Regarding changing the `end` method itself, we can revisit this point when a major version is under discussion.

wellwelwel:
Thanks again, @dstankovd! It's already available in the `mysql2@canary` version and will soon be available in the main version as well 🤝

PR #3784, opened 2025-09-06
fix(types): restrict StreamOptions.objectMode to true (#3686)

### LINKED ISSUE #3686: Query StreamOptions.objectMode has type `any`, should be `true`
The StreamOptions interface in mysql2/typings/mysql/lib/protocol/sequences/Query.js includes: `objectMode: any`. However, the library always overwrites the provided value for objectMode to `true`. So, objectMode can be removed from the interface as it doesn't do anything, or at least its value could be set to `true`.

minzunim:
fix(types): restrict StreamOptions.objectMode to true

As discussed in #3686, `objectMode` is always forced to true internally, so the current `any` type is misleading. This change updates the definition to `true` to better reflect the actual behavior.

Closes #3686

### DIFF
--- a/typings/mysql/lib/protocol/sequences/Query.d.ts
+++ b/typings/mysql/lib/protocol/sequences/Query.d.ts
@@ -94,7 +94,7 @@ export interface StreamOptions {
   /**
    * The object mode of the stream (Default: true)
    */
-  objectMode?: any;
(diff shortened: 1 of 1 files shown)


wellwelwel:
Thanks, @minzunim! 🚀

Issue #2888, opened 2024-07-24
`sha256_password` auth plugin doesn't work via TLS connection


viladimiru:
Hi! I found out that I am getting an `Access denied` error when I try to authenticate with the `sha256_password` plugin on `MySQL v8.0.35-27`. However, if I authenticate via the Linux `mysql-client` library, everything works fine.

As far as I know, this plugin has been deprecated since `MySQL v8`, but I can still authenticate with `mysql-client`.

Could you please continue to support this plugin in the modern version of MySQL? Our service is a client for MySQL databases, and our customers need to be able to work with this plugin if they prefer to use it.

sidorares:
> Can you please support this plugin for modern version of MySQL? Our service is a client for MySQL databases and our customers need to have ability to work with this plugin if they prefer to use it

I can, but this is few hours (or maybe days, who knows) of work for a functionality that only a small fraction of users need. Unfortunately this is quite low on my priorities list. Couple of ways this can realistically happen:
- do the change yourself. I'm happy to guide where I can and review & mere your work if it's acceptable quality ( at least some basic unit tests )
- buy me a time so that I can focus on that work ( via sponsorships or some other way )

viladimiru:
> > Can you please support this plugin for modern version of MySQL? Our service is a client for MySQL databases and our customers need to have ability to work with this plugin if they prefer to use it
> 
> I can, but this is few hours (or maybe days, who knows) of work for a functionality that only a small fraction of users need. Unfortunately this is quite low on my priorities list. Couple of ways this can realistically happen:
> 
> * do the change yourself. I'm happy to guide where I can and review & mere your work if it's acceptable quality ( at least some basic unit tests )
> * buy me a time so that I can focus on that work ( via sponsorships or some other way )

@sidorares hello! Thanks for the reply. I've made the fix in this pull request #3809 — would appreciate it if you could take a look!

After reading the MySQL documentation, I understood that the `sha256_password` plugin works partially correctly, but it did not account for the fact that manual encryption is not required when connecting via TLS.

[...]

PR #3809, opened 2025-09-22
fix: fix sha256_password to work correctly over a TLS connection

viladimiru:
After reading the MySQL documentation, I understood that the `sha256_password` plugin works partially correctly, but it did not account for the fact that manual encryption is not required when connecting via TLS.

[documentation](https://dev.mysql.com/doc/refman/8.4/en/sha256-pluggable-authentication.html#:~:text=If%20the%20connection%20is%20secure%2C%20an%20RSA%20key%20pair%20is%20unnecessary%20and%20is%20not%20used.%20This%20applies%20to%20connections%20encrypted%20using%20TLS.%20The%20password%20is%20sent%20as%20cleartext%20but%20cannot%20be%20snooped%20because%20the%20connection%20is%20secure.)

Follow up for https://github.com/sidorares/node-mysql2/issues/2888

### DIFF
--- a/lib/auth_plugins/sha256_password.js
+++ b/lib/auth_plugins/sha256_password.js
@@ -3,7 +3,7 @@
 const PLUGIN_NAME = 'sha256_password';
 const crypto = require('crypto');
 const { xorRotating } = require('../auth_41');
-const tls = require('tls');
+const Tls = require('tls');
 
 const REQUEST_SERVER_KEY_PACKET = Buffer.from([1]);
 
@@ -34,7 +34,7 @@ module.exports =
       switch (state) {
         case STATE_INITIAL:
           if (
-            connection.stream instanceof tls.TLSSocket &&
+            connection.stream instanceof Tls.TLSSocket &&
             connection.stream.encrypted === true
           ) {
             // We don't need to encrypt passwords over TLS connection


wellwelwel:
Thanks, @viladimiru! I didn't get to test it in practice, but just bringing the [quote](https://dev.mysql.com/doc/refman/8.4/en/sha256-pluggable-authentication.html#:~:text=If%20the%20connection%20is%20secure%2C%20an%20RSA%20key%20pair%20is%20unnecessary%20and%20is%20not%20used.%20This%20applies%20to%20connections%20encrypted%20using%20TLS.%20The%20password%20is%20sent%20as%20cleartext%20but%20cannot%20be%20snooped%20because%20the%20connection%20is%20secure.) in question here (the same applies to **MySQL** **8** and **9**):

> For clients that use the sha256_password plugin, passwords are never exposed as cleartext when connecting to the server. How password transmission occurs depends on whether a secure connection or RSA encryption is used:
> 
> - If the connection is secure, an RSA key pair is unnecessary and is not used. This applies to connections encrypted using TLS. The password is sent as cleartext but cannot be snooped because the connection is secure.
> 
> - If the connection is not secure, and an RSA key pair is available, the connection remains unencrypted. This applies to connections not encrypted using TLS. RSA is used only for password exchange between client and server, to prevent password snooping. When the server receives the encrypted password, it decrypts it. A scramble is used in the encryption to prevent repeat attacks.
> 
> - If a secure connection is not used and RSA encryption is not available, the connection attempt fails because the password cannot be sent without being exposed as cleartext.

Also, regarding the missing tests, it is not actually related to this _PR_: this part of the logic was never covered.

wellwelwel:
In `"If the connection is not secure, and an RSA key pair is available, the connection remains unencrypted"`, I'm unsure about someone using the `rejectUnauthorized` option.

@viladimiru, in this scenario, wouldn't it be appropriate to keep the password encrypted?

wellwelwel:
I believe that instead of relying on the configuration provided by the user _(`if (connection.config.ssl) ...`)_, we can ensure it through the connection itself, for example:

```js
const tls = require('tls');

// ...

if (
  connection.stream instanceof tls.TLSSocket &&
  connection.stream.encrypted === true
) {
// ...
```

- The tests that passed in the bellow screenshot are based on this approach 🙋🏻‍♂️

> [!NOTE]
> 
> Testing locally by forcing `sha256_password` and **SSL**, I was able to see both the error mentioned in #2888 and the solution working:
>
> Screenshot 2025-09-29 at 19 36 04

viladimiru:
> I believe that instead of relying on the configuration provided by the user (if (connection.config.ssl) ...), we can ensure it through the connection itself, for example:
> ```js
> const tls = require('tls');
> 
> // ...
> 
> if (
>   connection.stream instanceof tls.TLSSocket &&
>   connection.stream.encrypted === true
> ) {
> // ...
> ```

Hi @wellwelwel! Thank you for the suggestion!

Unfortunately, I'm not very knowledgeable about this particular issue, so if you think it's better this way, I will adjust the condition.

wellwelwel:
> Unfortunately, I'm not very knowledgeable about this particular issue [...]

When using `rejectUnauthorized`, the `connection.config.ssl` will be _truthy_ because it 'll be an object (`{}`), while the connection could be unencrypted. It is not an issue for **MySQL2** or **MySQL**, but I consider it worth covering this possibility.

By checking the connection itself, we can confirm that it is actually secure before sending data as plain text.

wellwelwel:
Thanks again, @viladimiru! It's already available in `mysql2@canary` and soon in the `latest` version too 🤝

PR #3819, opened 2025-09-24
fix(typings): missing callback to PoolCluster.end()

mex:
This aligns the TS type of the `PoolCluster.end()` method to match the JS implementation in [pool_cluster.js](https://github.com/sidorares/node-mysql2/blob/6bfad13a4ce98b4534acc7daae340d027db63479/lib/pool_cluster.js#L232-L265).


wellwelwel:
Thanks, @mex! **LGTM**.

Issue #2588, opened 2024-04-11
got warning when host is an IP address


avispeng:
Hello.

We use knex and mysql2 to connect to Google Cloud SQL from a local machine.
```
const knex = require('knex');
const fs = require('fs');

const params = {
  client: 'mysql2',
  connection: {
    user: "db user",
    password: "db password",
    database: "db name",
    timezone: 'Z',
    host: "public IP address of db",
    ssl: {
      ca: fs.readFileSync(__dirname + '/server-ca.pem'),
      cert: fs.readFileSync(__dirname + '/client-cert.pem'),
      key: fs.readFileSync(__dirname + '/client-key.pem'),
    }
  },
};

const knexInstance = knex(params);
```

The issue is, when the `host` is an IP address (provided by Google Cloud SQL), we get warning

> [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.

It seems that the cause is the `host` is passed as `servername` in the call of `tls.connect()`, and according to https://nodejs.org/api/tls.html#tlsconnectoptions-callback
> servername: Server name for the SNI (Server Name Indication) TLS extension. It is the name of the host being connected to, and must be a host name, and not an IP address.

I am wondering what would be the correct approach to resolve this issue? Thank you.

sidorares:
linking related #2560

ctindel:
I think it would be great to provide an option for the caller to set whatever string they want in the SNI field, separating the hostname/IP address you connect to from whatever is placed into SNI.

sidorares:
@ctindel agree. PRs welcome :)

PR #3835, opened 2025-09-30
fix: skip SNI for IP addresses in TLS connection

mex:
When connecting using TLS there is an optional field, `servername`, for setting the [Server Name Indication (SNI)](https://en.wikipedia.org/wiki/Server_Name_Indication) but this field only makes sense to set if we are using a hostname to connect and not an IP. Since v12 of Node.js, it's been deprecated to provide an IP address as the SNI so it will be ignored at some point anyway.

This PR prevents setting `servername` in the TLS config if the MySQL config `host` field is an IP which in turn quiets this deprecation warning:
```
(node:1) [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.
```

The logic for setting SNI to the host was originally added in https://github.com/sidorares/node-mysql2/pull/1554.

Fixes https://github.com/sidorares/node-mysql2/issues/2588

### DIFF
--- a/lib/base/connection.js
+++ b/lib/base/connection.js
@@ -345,7 +345,9 @@ class BaseConnection extends EventEmitter {
     });
     const rejectUnauthorized = this.config.ssl.rejectUnauthorized;
     const verifyIdentity = this.config.ssl.verifyIdentity;
-    const servername = this.config.host;
+    const servername = Net.isIP(this.config.host)
+      ? undefined
+      : this.config.host;
 
     let secureEstablished = false;
     this.stream.removeAllListeners('data');


bertho-zero:
It should be possible to choose the servername. I connect with an IP address, but the certificate is generated with a domain (the FQDN) in the alt_names that isn't on any public DNS.

By connecting with an IP address as host and a servername for the SSL configuration, this warning disappears.

mex:
> It should be possible to choose the servername. I connect with an IP address, but the certificate is generated with a domain (the FQDN) in the alt_names that isn't on any public DNS.
> 
> By connecting with an IP address as host and a servername for the SSL configuration, this warning disappears.

I think it makes a lot of sense to allow specifically setting the servername as you do in https://github.com/sidorares/node-mysql2/pull/3845. I considered adding it as part of this as well but I would classify it as a separate feature. If the SSL certificate does not have a FQDN/SAN, I will still need the change as proposed here.

wellwelwel:
Thanks, @mex! It's already available in `mysql2@canary` and soon in the `latest` version too 🤝

mex:
Perfect, thanks!

Issue #2560, opened 2024-04-04
Is there any way to get this library to set the SNI field when establishing a TLS connection?


ctindel:
I don't see any options for having it set the TLS server name indication field when ssl attributes are part of the configuration. Is there a way to do this?

sidorares:
It's currently set to be `hostname` you are connecting to - https://github.com/sidorares/node-mysql2/blob/caa859893ca426cead740389ad6a2498125a456f/lib/connection.js#L358-L371

What's your scenario? Do you need to set it to a different value?

ctindel:
Maybe I'm doing something wrong, I couldn't get it working to connect using mTLS based authentication and I thought maybe it wasn't setting the SNI.

PR #3845, opened 2025-10-03
feat: allow to override ssl servername

### LINKED ISSUE #2588: got warning when host is an IP address

bertho-zero:
Ideally, you would be able to choose the server name used for SSL verification.

Another way to fix #2588 and #3835

### DIFF
--- a/lib/base/connection.js
+++ b/lib/base/connection.js
@@ -345,7 +345,7 @@ class BaseConnection extends EventEmitter {
     });
     const rejectUnauthorized = this.config.ssl.rejectUnauthorized;
     const verifyIdentity = this.config.ssl.verifyIdentity;
-    const servername = this.config.host;
+    const servername = this.config.ssl.servername || this.config.host;
 
     let secureEstablished = false;
     this.stream.removeAllListeners('data');


wellwelwel:
Thanks for the idea, @bertho-zero. I believe that this isn't an alternative, but a continuation of #3835 and resolves #2560.

When enabling this, it's a good idea to align the types and add the usage to the documentation (for example at https://sidorares.github.io/node-mysql2/docs/documentation/extras).

> [!NOTE]
> 
> I'll keep the _PR_ as a draft for now. Feel free to continue the idea after #3835 _(already merged)_.

PR #3855, opened 2025-10-09
fix: add missing charset encoding for UTF8MB4_0900_BIN

kangsorang:
# Fix charset encoding mapping for UTF8MB4_0900_BIN

## Problem
MySQL2 throws `Encoding not recognized: 'undefined'` error when connecting to MySQL 8.0+ servers using `utf8mb4_0900_bin` collation.

## Root Cause
Mismatch between `charsets.js` and `charset_encodings.js`:
- `charsets.js` defines `UTF8MB4_0900_BIN = 309`
- `charset_encodings.js` array has length 309 (indices 0-308)
- `charset_encodings[309]` returns `undefined`
- `Iconv.getDecoder(undefined)` throws encoding error

## Solution
Add missing encoding mapping for index 309 in `charset_encodings.js`.

## Changes
- **File**: `lib/constants/charset_encodings.js`
- **Change**: Add `'utf8'` entry at index 309 for `UTF8MB4_0900_BIN`

## Impact
- Fixes connection issues with MySQL 8.0+ `utf8mb4_0900_bin` collation
- Resolves `Encoding not recognized: 'undefined'` errors
- Fix : #3856

### DIFF
--- a/lib/constants/charset_encodings.js
+++ b/lib/constants/charset_encodings.js
@@ -313,4 +313,5 @@ module.exports = [
   'utf8',
   'utf8',
   'utf8',
+  'utf8',
 ];


wellwelwel:
Thanks, @kangsorang ✨

Do you think it's possible to have a test that covers this case (reproducing the failure if it doesn't have your modification)? Perhaps a temporary table with this encoding would solve it.

wellwelwel:
Thanks again, @kangsorang! I'll include the following test in a new _PR_ to ensure consistency between A and B to prevent that kind of error:

```ts
import { assert, describe, it } from 'poku';
import mysql from '../../../../index.js';

describe('Charset to Encoding coverage', () => {
  const charsets = mysql.Charsets;
  const encodings = mysql.CharsetToEncoding;

  it('every charset ID should have a defined encoding', () => {
    const missing: string[] = [];
    const seen = new Set<number>();

    for (const [name, id] of Object.entries(charsets)) {
      if (typeof id !== 'number' || seen.has(id)) continue;

      seen.add(id);
      if (encodings[id] === undefined) missing.push(`${name} (id=${id})`);
    }

    assert.deepStrictEqual(
      missing,
      [],
      `Missing encoding mappings: ${missing.join(', ')}`
    );
  });
});
```

> Screenshot 2026-02-20 at 08 07 17

Issue #3859, opened 2025-10-11
Tests can hang indefinitely when connection resources aren’t cleaned up after assertion failures


HappyZombies:
### **TL;DR: The tests don’t reliably clean up after themselves on failures, failed assertions can leave connections, pools, or prepared statements open, causing the process to hang. The poku testing module doesn't seem to have an option to "fix" these tests, so certain tests must be manually fixed.**

Hey there! I wrote an issue a few days ago and found time to sit down and look into what the problem was, and I believe I have found it! 

While running the mysql2 test suite locally, I noticed that several unit tests can hang randomly if a connection, pool, or prepared statement isn’t properly closed when a test fails.

I initially discovered this because my test database wasn’t named `test` (I used `nodetest`). That caused some assertions to fail, since some of the checks are hardcoded (for example, [here](https://github.com/sidorares/node-mysql2/blob/master/test/esm/integration/connection/test-column-inspect.test.mjs#L96)) but instead of the suite exiting cleanly, it would hang indefinitely.

After digging, I found that some test files wrap test() calls in async IIFEs and/or place cleanup code (e.g. connection.end() or pool.end()) in a stop that can occur after a failing assert. So, when the assertion throws, that teardown line never runs, leaving sockets or prepared statements open. I _think_ that Poku waits for the event loop to drain, but the leaked handle keeps it alive forever.

I have ran this problem into the following files, but I have seen it/sure that there are way more.

* https://github.com/sidorares/node-mysql2/blob/master/test/esm/integration/connection/test-column-inspect.test.mjs
* https://github.com/sidorares/node-mysql2/blob/master/test/esm/unit/check-extensions.test.mjs

Note that it's not always something that is wrapped in an async IIFEs, but a majority of the hangs are.

Integration test [`test-column-inspect.test.mjs`](https://github.com/sidorares/node-mysql2/blob/master/test/esm/integration/connection/test-column-inspect.test.mjs) is the easiest place to see this occur -- simply make that last assert fail in any way (add a typo/bad field or create a database who's name isn't `test`) and the test won't ever continue/close/finish do the hanging connection.

Or just run this simple unit test some test directory.

```js
import { test, assert } from 'poku';
import util from 'node:util';
import { createRequire } from 'node:module';

const require = createRequire(import.meta.url);
const common = require('../../../common.test.cjs');

(async () => {
  const connection = common.createConnection().promise();

  test(async () => {
    const [, columns] = await connection.query(
[...]

Issue #3854, opened 2025-10-08
Can't seem to run unit tests nor coverage tests locally with MySQL in a docker container, tests seems to randomly hang/stall.


HappyZombies:
Hey there, I am looking to contribute to this package =)

I started by reading the contribution guide and it said how to run the tests. So I decide to run the unit tests to make sure all is green before I start adding/changing stuff lol

But I ran into a problem running them locally, I see that they definelty all do run [according to this action run.](https://github.com/sidorares/node-mysql2/actions/runs/18338736295/job/52228744967)

But running them locally I am not having success, they seem to intermiately halt/stall at random points during the test run.

Repo Steps:

1. Docker Compose

```yml
services:
  mysql:
    image: mysql:8.0
    # platform: linux/amd64  --- add if on M1
    container_name: mysql-v8.0
    environment:
      MYSQL_ROOT_PASSWORD: "somepassword"
      MYSQL_DATABASE: test
    ports:
      - "3306:3306"
    volumes:
      - mysql_data:/var/lib/mysql
```

Run `docker compose up -d`

2. Made a launch.json for VSCode

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node",
            "request": "launch",
            "name": "Debug npm test (poku)",
            "runtimeExecutable": "npm",
            "runtimeArgs": [
                "run-script",
                "test"
            ],
            "port": 9229,
            "env": {
                "CI": "1",
                "MYSQL_HOST": "127.0.0.1",
                "MYSQL_USER": "root",
                "MYSQL_PASSWORD": "somepassword",
                "MYSQL_DATABASE": "node_test"
            },
            "console": "integratedTerminal",
            "internalConsoleOptions": "neverOpen"
        }
    ]
}
```

Or simply just exported it 

```sh
export CI=1
export MYSQL_HOST=127.0.0.1
export MYSQL_USER=root
export MYSQL_PASSWORD=somepassword
export MYSQL_DATABASE=node_test
```

NOTE: I noticed that the tests continue down a lot further IF I set MYSQL_HOST to 127.0.0.1, if I use localhost or 0.0.0.0, they stop a lot earlier at `test/esm/unit/check-extensions.test.mjs`, always.

3. Run `npm run test`

**Expected**:

All tests should run and complete.

**Actual**:

The test runner sometimes hangs or stops at different spots. Sometimes only a few tests run, sometimes more, but it never completes. This only happens locally; in CI all tests pass. 

[...]

PR #3871, opened 2025-10-17
refactor(docs,test): small clarification update on contribution guidelines and fix URI-encoding + hardcoded-db tests

HappyZombies:
This PR does the following:

* Test: Encodes MYSQL username/password when constructing the config URI in `common.test.cjs` to prevent malformed URIs with reserved characters. This is to prevent the issue where a username and/or password could contain special characters and it not get parsed properly (e.g., username is "engineer:frontend", "engineer" would be passed as the username, not "engineer:frontend).

* Test: Adds a targeted `.catch()` in some test files, this ensures teardown (e.g., connection.end()) runs reliably.

* Docs: Update Contributing guide to clearly state the database name MUST be exactly "test" when running the suite.

Please NOTE:

The added .catch() in these two files do not cover all other test suites with similar patterns (e.g., if they fail, the tests will hang). It aims to address the most common pitfall (at least that I believe), which are failures caused by using the wrong DB name when running test suites, due to some hard coded "test" assertion tests.

I believe that a comprehensive fix would require a broader refactor across all tests to unify error handling and proper resource teardown (e.g., actually calling connection.end()).

So point is, this is a temporary for the test files with the `.catch()` on it, so I am not opposed if they want to be removed/not added for a better refactor down the line. :)

Address: #3859 and #3854


wellwelwel (changes requested):
Thanks, @HappyZombies!

Using `catch` for errors raises some concerns for me, especially because it ignores errors when they actually need to be flagged.

We can use the `describe` + `it` set instead, for example:

- Before:

```js
await connect();

assert(false); // throws

await disconnect(); // this will not be performed 
```

- New:

```js
await describe('...', async () => {
  await connect();

  it('...', () => {
    assert(false); // throws within the `it` scope
  });

   await disconnect(); // this will be performed
});
```

In practice, you can change from this [_(a behavior particular to **Poku**)_](https://poku.io/docs/documentation/helpers/describe#using-as-test-titles):

```js
describe('...', common.describeOptions);
```

To this [_(widely adopted standard in testing)_](https://poku.io/docs/documentation/helpers/describe#grouping-tests-usual):

```js
await describe('...', async () => {
  // ...
});
```

> [!TIP]
> 
> - In **ES Module** tests, there is no need for `(async () => {})()` 🙋🏻‍♂️
> - **Related:**
>   - https://poku.io/docs/documentation/helpers/describe
>   - https://poku.io/docs/documentation/helpers/it

dannyreg:
> 
> ```js
> await describe('...', async () => {
>   // ...
> });
> ```
> 

@wellwelwel this pattern still doesn't seem to work under this particular context, looking at `test-column-inspect-test`, I refactored it to be the following.

```js
import { test, assert, describe, beforeEach, it } from 'poku';
import util from 'node:util';
import { createRequire } from 'node:module';

const require = createRequire(import.meta.url);
const common = require('../../../common.test.cjs');

await describe('Custom inspect for column definition', async () => {
  const connection = common.createConnection().promise();

  it('maps fields to schema-like description when depth > 1', async () => {
    const schema = `
        id INT NOT NULL AUTO_INCREMENT,
        weight INT(2) UNSIGNED ZEROFILL,
        usignedInt INT UNSIGNED NOT NULL,
        signedInt INT NOT NULL,
        unsignedShort SMALLINT UNSIGNED NOT NULL,
        signedShort SMALLINT NOT NULL,
        tinyIntUnsigned TINYINT UNSIGNED NOT NULL,
        tinyIntSigned TINYINT NOT NULL,
        mediumIntUnsigned MEDIUMINT UNSIGNED NOT NULL,
        mediumIntSigned MEDIUMINT NOT NULL,
        bigIntSigned BIGINT NOT NULL,
        bigIntUnsigned BIGINT UNSIGNED NOT NULL,
        longText_ LONGTEXT NOT NULL,
        mediumText_ MEDIUMTEXT NOT NULL,
        text_ TEXT NOT NULL,
        tinyText_ TINYTEXT NOT NULL,
        varString_1000 VARCHAR(1000) NOT NULL,
        decimalDefault DECIMAL,
        decimal13_10 DECIMAL(13,10),
        floatDefault FLOAT,
        float11_7 FLOAT(11,7),
        dummyLastFieldToAllowForTrailingComma INT,
      `;

    await connection.query(
      `CREATE TEMPORARY TABLE test_fields (${schema} PRIMARY KEY (id))`
    );

    const [, columns] = await connection.query('select * from test_fields');
    const inspectResults = util.inspect(columns);
    const schemaArray = schema
      .split('\n')
      .map((line) => line.trim())
      .filter((line) => line.length > 0)
      .map((line) => {
        const words = line.split(' ');
        const name = `\`${words[0]}\``;
        return [name, ...words.slice(1)].join(' ');
      });

    const normalizedInspectResults = inspectResults
      .split('\n')
      .slice(1, -2) // remove "[" and "]" lines and also last dummy field
      .map((line) => line.trim())
      // remove primary key - it's not in the schema explicitly but worth having in inspect
      .map((line) => line.split('PRIMARY KEY ').join(''));

    for (let l = 0; l < normalizedInspectResults.length; l++) {
      const inspectLine = normalizedInspectResults[l];
      const schemaLine = schemaArray[l];

      assert.equal(
        inspectLine,
        schemaLine,
        'Loop: Should map fields to a schema-like description when depth is > 1'
      );
    }
  });

  it('shows detailed description when depth < 1', async () => {
    await connection.query(`
        CREATE TEMPORARY TABLE test_fields2 (
            id INT,
            decimal13_10 DECIMAL(13,10) UNSIGNED NOT NULL,
            PRIMARY KEY (id)
        )
      `);

    const [, columns] = await connection.query('select * from test_fields2');
    const inspectResults = util.inspect(columns[1]);

    assert.deepEqual(
      inspectResults,
      util.inspect({
        catalog: 'def',
        schema: 'test',
        name: 'decimal13_10',
        orgName: 'decimal13_10',
        table: 'test_fields2',
        orgTable: 'test_fields2',
        characterSet: 63,
        encoding: 'binary',
        columnLength: 14,
        type: 246,
        flags: ['NOT NULL'],
        decimals: 10,
        typeName: 'NEWDECIMAL',
      }),
      'should show detailed description when depth is < 1'
    );
  });

  await connection.end();
});

```

Notice that the callback for the `it` method is an async query, due to this, the await connection.end() gets called first and causes a race case, that being the connection terminates first before the async callback is ran, this results in the following error:

```txt
unhandledRejection Error: Can't add new command when connection is in closed state
```

Now adding await in front of the `it` method fixes this problem, but it re-introduces the .connection.end() method never being called. 

```js
await connect();

assert(false); // throws

await disconnect(); // this will not be performed 
```

It seems like when the throw occurs inside the it scope (when await is added in the front) it never logs the error since the connection is still alive. 

Maybe we can add some way in the commons file to register the connections and disconnect if an unhandledRejection occurs? reason why I added the catch since that is what is happening, an uncaught or unhandledRejection is occurring and not being caught since the connection to the DB is alive/not releasing the event loop.

Does Poku have something like this so I can try? I don't think it does -- idea here would be to simply connect and disconnect after ever run, but then again idk iit would really solve the problem.

```js
 await beforeAll(async () => {
    connection = common.createConnection().promise();
    await connection.connect?.(); // fully wait to be connected first
  });

  await afterAll(async () => {
    await connection.end();
  });
  ```
  
 I wonder how other test frameworks have solved this issue if at all? Or maybe your test framework has a sense of a timeout so if any error throws it can move on? Idk, point is, I can't get this to work how you want cuz it doesn't work! haha, I really think I need try/catches here!

**Comment by wellwelwel** [comment]
> unhandledRejection Error: Can't add new command when connection is in closed state

I believe the condition you mentioned is because you are not waiting for the result of the `it` methods (`async/await`).

> It seems like when the throw occurs inside the it scope (when await is added in the front) it never logs the error since the connection is still alive.

I took your last example and ensured the `await` in the `it` methods and also used the `beforeEach` and `afterEach` _hooks_ to clean up and close the connection after each test _(`it`)_ cycle. @HappyZombies, can you try this way?

```js
import { assert, describe, afterEach, beforeEach, it } from 'poku';
import util from 'node:util';
import { createRequire } from 'node:module';

const require = createRequire(import.meta.url);
const common = require('../../../common.test.cjs');

await describe('Custom inspect for column definition', async () => {
  let connection;

  beforeEach(async () => {
    connection = common.createConnection().promise();
    await connection.query(`DROP TABLE IF EXISTS test_fields`);
  });

  afterEach(async () => {
    await connection.end();
  });

  await it('maps fields to schema-like description when depth > 1', async () => {
    const schema = `
        id INT NOT NULL AUTO_INCREMENT,
        weight INT(2) UNSIGNED ZEROFILL,
        usignedInt INT UNSIGNED NOT NULL,
        signedInt INT NOT NULL,
        unsignedShort SMALLINT UNSIGNED NOT NULL,
        signedShort SMALLINT NOT NULL,
        tinyIntUnsigned TINYINT UNSIGNED NOT NULL,
        tinyIntSigned TINYINT NOT NULL,
        mediumIntUnsigned MEDIUMINT UNSIGNED NOT NULL,
        mediumIntSigned MEDIUMINT NOT NULL,
        bigIntSigned BIGINT NOT NULL,
        bigIntUnsigned BIGINT UNSIGNED NOT NULL,
        longText_ LONGTEXT NOT NULL,
        mediumText_ MEDIUMTEXT NOT NULL,
        text_ TEXT NOT NULL,
        tinyText_ TINYTEXT NOT NULL,
        varString_1000 VARCHAR(1000) NOT NULL,
        decimalDefault DECIMAL,
        decimal13_10 DECIMAL(13,10),
        floatDefault FLOAT,
        float11_7 FLOAT(11,7),
        dummyLastFieldToAllowForTrailingComma INT,
      `;

    await connection.query(
      `CREATE TEMPORARY TABLE test_fields (${schema} PRIMARY KEY (id))`
    );

    const [, columns] = await connection.query('select * from test_fields');
    const inspectResults = util.inspect(columns);
    const schemaArray = schema
      .split('\n')
      .map((line) => line.trim())
      .filter((line) => line.length > 0)
      .map((line) => {
        const words = line.split(' ');
        const name = `\`${words[0]}\``;
        return [name, ...words.slice(1)].join(' ');
      });

    const normalizedInspectResults = inspectResults
      .split('\n')
      .slice(1, -2) // remove "[" and "]" lines and also last dummy field
      .map((line) => line.trim())
      // remove primary key - it's not in the schema explicitly but worth having in inspect
      .map((line) => line.split('PRIMARY KEY ').join(''));

    for (let l = 0; l < normalizedInspectResults.length; l++) {
      const inspectLine = normalizedInspectResults[l];
      const schemaLine = schemaArray[l];

      assert.equal(
        inspectLine,
        schemaLine,
        'Loop: Should map fields to a schema-like description when depth is > 1'
      );
    }
  });

  await it('shows detailed description when depth < 1', async () => {
    await connection.query(`
        CREATE TEMPORARY TABLE test_fields2 (
            id INT,
            decimal13_10 DECIMAL(13,10) UNSIGNED NOT NULL,
            PRIMARY KEY (id)
        )
      `);

    const [, columns] = await connection.query('select * from test_fields2');
    const inspectResults = util.inspect(columns[1]);

    assert.deepEqual(
      inspectResults,
      util.inspect({
        catalog: 'def',
        schema: 'test',
        name: 'decimal13_10',
        orgName: 'decimal13_10',
        table: 'test_fields2',
        orgTable: 'test_fields2',
        characterSet: 63,
        encoding: 'binary',
        columnLength: 14,
        type: 246,
        flags: ['NOT NULL'],
        decimals: 10,
        typeName: 'NEWDECIMAL',
      }),
      'should show detailed description when depth is < 1'
    );
  });
});
```

<table>
<tr>
<th>
Success Example
</th>
<th>
Failure Example
</th>
</tr>
<tr>
<td>
Screenshot 2025-10-20 at 16 10 18
</td>
<td>
Screenshot 2025-10-20 at 16 11 20
</td>
</tr>
</table>

dannyreg:
@wellwelwel mhm yup that seems to work, the key thing here was using adding the after each hook, not just calling .end() at the end of the tests.

```js
  afterEach(async () => {
    await connection.end();
  });
```

And of course, updating the beforeEach() to connect again. Ok this will work then, guess I was really hellbent on keeping the test as it was.

I think a lot more unit tests could benefit from this, however for this PR I will start with these two that I added and look at fixing others in another PR.

wellwelwel:
Thanks again, @HappyZombies! ✨

Issue #3764, opened 2025-08-23
Request: Public API for connection validation


mercmobily:
## 

Knex.js needs to validate mysql2 connections before use in connection pools. Currently, we're forced to use private properties that could break with any update.

### Current Situation
```javascript
// Knex's mysql2 dialect - relies on private properties
validateConnection(connection) {
  return (
    connection &&
    !connection._fatalError &&      // Private
    !connection._protocolError &&   // Private
    !connection._closing &&         // Private
    !connection.stream.destroyed
  );
}
```

For comparison, the original `mysql` driver provides public API:
```javascript
// mysql driver - uses public 'state' property
validateConnection(connection) {
  return connection.state === 'connected' || connection.state === 'authenticated';
}
```

### Background
In 2017, you kindly offered to add public API for this ([knex#2175](https://github.com/knex/knex/pull/2175#issuecomment-350890643)):
> I'm happy to add anything that would work for you, ideally in a way compatible with mysqljs/mysql
> what if I wrap `_fatalError` into a getter function and "bless" it to be public api?

This would still be incredibly helpful for connection pool management.

### Request
Could mysql2 provide one of these public methods?

**Option A: Simple boolean check**
```javascript
connection.isValid()  // returns true/false
```

**Option B: State property (mysql compatible)**
```javascript
connection.state  // 'connected' | 'disconnected' | 'protocol_error' | 'closing'
```

**Option C: Expose existing checks**
```javascript
connection.hasFatalError()     // or
connection.hasProtocolError()  // or
connection.isClosing()
```

### Related Issues
- knex/knex#6255 - MySQL2 relies on private API for connection validation
- knex/knex#1798 - Original connection handling issue
- knex/knex#2175 - Previous attempt to improve validation

Would you consider adding any of these public methods? Option B would maintain compatibility with the original mysql driver, making migration easier for users.

Thank you for maintaining this excellent library!

dannyreg:
Adding option B, note that the original mysql module does not have a "closing" state, so will not be adding this.

mercmobily:
@HappyZombies Fantastic, thank you very much!

The mysql driver never exposed a “closing” state either, and Knex’s mysql validation has worked fine by accepting only state === 'connected' || state  === 'authenticated'
 
For mysql2, we can prefer the new connection.state and still keep a  conservative check that the underlying stream is not destroyed. That covers the common “being torn down” case without needing a dedicated “closing”  state.

[...]

PR #3885, opened 2025-10-23
Add state property to BaseConnection to match original mysql package

HappyZombies:
This PR intends to address [#3764](https://github.com/sidorares/node-mysql2/issues/3764) by introducing a `state` property on the connection object, aligning behavior with the [original mysql package](https://github.com/mysqljs/mysql/blob/master/lib/Connection.js#L21).

The new property reflects the current connection status; it only includes the following three states: disconnected, connected, or protocol_error.

The state flow includes:

Default: 'disconnected'

Updated to 'connected' upon successful handshake.

Updated to 'disconnected' after .end() or .close() is called (note that failures can eventually call this method).

Updated to 'protocol_error' on fatal or unexpected protocol errors.

Tests include:

Connects successfully and sets state to connected.

Closes cleanly and sets state to disconnected.

Incorrect credentials result in state protocol_error.

Let me know what you think, and tell me if any changes should be made -- along if we need more tests as well :)

### DIFF
--- a/test/unit/connection/test-connection-state.test.cjs
+++ b/test/unit/connection/test-connection-state.test.cjs
@@ -1,0 +1,98 @@
+'use strict';
+const { assert, describe, it, beforeEach, afterEach } = require('poku');
+const common = require('../../common.test.cjs');
+
+(async () => {
+  await describe('Connection state', async () => {
+    let connection;
+
+    beforeEach(() => (connection = common.createConnection().promise()));
+
+    afterEach(async () => {
+      if (!connection) return;
+      try {
+        await connection.end();
+      } catch (e) {
+        // ignore teardown errors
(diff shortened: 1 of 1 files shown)


wellwelwel (changes requested):
A tip would be to use the `sleep` method instead of setting the `setTimeout` multiple times with a `Promise`, for example:

```js
import { sleep } from 'poku';

// ...
await sleep(20);
// ...
```

dannyreg:
I just realized I forgot to add the "authorized" state, so I will be adding that and also fixing the unit tests as per your recommendation :)

wellwelwel:
Hey, @HappyZombies. I'm closing this in favor of #3958. Thank you for working on it 🤝

PR #3916, opened 2025-11-16
docs: update custom stream SOCKS5 example to modern API (fixes outdated callback usage)

### LINKED ISSUE #3655: Pooling over a socks5 proxy does not work
Example given in documentation (https://sidorares.github.io/node-mysql2/docs/documentation/extras#connecting-using-custom-stream) no longer works:

```js
const SocksConnection = require('socksjs');
  database: 'test',
  user: 'foo',
  password: 'bar',
  stream: function (cb) {
    const newStream = new SocksConnection(
      { host: 'remote.host', port: 3306 },
      { host: 'localhost', port: 1080 }
    );
    cb(null, newStream);
  },
});

pool.query('SELECT 1')
```

gives:

```
Uncaught TypeError: cb is not a function
```

I couldn't figure out any way to pool over socks5 proxies, using either [socksjs](https://github.com/M2Ys4U/socksjs) lib mentioned in the example (which had its last release 11 years ago) or with the more popular [socks](https://github.com/JoshGlazebrook/socks) package.

Priyank-Gaur:
This PR updates the “Connecting using custom stream” documentation.

The old example used a callback-style:
stream: function (cb) { ... }

This pattern no longer works in current versions of mysql2 and results in the runtime error:
TypeError: cb is not a function

What changed

Replaced the outdated callback-based SOCKS5 example.
Added a modern example where stream returns a Duplex stream or a Promise that resolves to one.
Added a short note explaining that the callback-style API is no longer supported.

Why this change

The existing docs mislead users into using a deprecated pattern that fails on current mysql2 versions.
The updated version shows the correct API and prevents confusion/errors for anyone connecting over SOCKS5 or other custom transport streams.

Related Issue
Fixes #3655

### DIFF
--- a/website/docs/documentation/extras.mdx
+++ b/website/docs/documentation/extras.mdx
@@ -127,15 +127,15 @@ const pool = mysql.createPool({
       proxy: {
         host: 'localhost',
         port: 1080,
-        type: 5
+        type: 5,
       },
       destination: {
         host: 'remote.host',
-        port: 3306
-      }
+        port: 3306,
+      },
     });
     return socket; // must be a Duplex stream
-  }
+  },
 });
 ```
 


wellwelwel:
Thanks, @Priyank-Gaur! ✨

siddharthvp:
What versions of mysql2 and socks is this tested with?

With the latest ones (mysql2 3.15.3, socks 2.8.7):

Firstly, socks `createConnection` requires `command` to be passed. Without that it gives: 
```
SocksClientError: An invalid SOCKS command was provided. Valid options are connect, bind, and associate.
```

If we set command to any of those 3 values, the code gives:
```
    this.stream.on('error', this._handleNetworkError.bind(this));
                ^
TypeError: this.stream.on is not a function
 ```

when getting a connection from the pool.

Priyank-Gaur:
> What versions of mysql2 and socks is this tested with?
> 
> With the latest ones (mysql2 3.15.3, socks 2.8.7):
> 
> Firstly, socks `createConnection` requires `command` to be passed. Without that it gives:
> 
> ```
> SocksClientError: An invalid SOCKS command was provided. Valid options are connect, bind, and associate.
> ```
> 
> If we set command to any of those 3 values, the code gives:
> 
> ```
>     this.stream.on('error', this._handleNetworkError.bind(this));
>                 ^
> TypeError: this.stream.on is not a function
> ```
> 
> when getting a connection from the pool.

Thanks for checking this!

My PR focused on updating the old callback-style example to the modern stream-returning API since the previous example no longer worked with mysql2.

It looks like the SOCKS library (`socks` package) now requires an additional `command` field and may behave differently in the latest versions.

If needed, I can open a follow-up PR  or If you want you can open a PR to update the example with a working configuration for the latest socks version.

siddharthvp:
As I said in #3655, there is no working configuration I can come up with.

PR #3928, opened 2025-11-21
ci: Fix MySQL version detection for ORDER BY placeholder parameter count on 2052 unit test

### LINKED ISSUE #3927: Unit test 2052.test.mjs has Incorrect version detection in the method of hasIncorrectPrepareParameter for MySQL 8.4.x
The version-detection logic used in mysql2’s test suite for detecting the “incorrect parameter count in COM_STMT_PREPARE” is slightly flawed, and causes tests to fail on MySQL 8.4.x (including 8.4.7).

MySQL 8.4.x correctly reports parameter counts, but the test expects the old bugged behavior.

## Propose Fixed:

Refactor the method of `hasIncorrectPrepareParameter`

The method currently is a little hard to follow and it can also be simplified a bit.

Reproduce:

Have version 8.4 of MySQL, run the unit test 2052.test.mjs against it, due to #3859 , the tests will hang, but we incorrectly get the hasIncorrectPrepareParameter set to true.

https://github.com/sidorares/node-mysql2/blob/490502f9c0f9f0699256930fda04ea16a78c11df/test/esm/regressions/2052.test.mjs#L143

HappyZombies:
This MR updates the hasIncorrectPrepareParameter logic to correctly reflect which MySQL versions actually exhibit the COM_STMT_PREPARE “ORDER BY ?” parameter-count bug.

This MR will resolve #3927 

Changes include:

* Marking 8.1, 8.2, 8.3, and 8.4.0 as "has incorrect prepare parameter"

* Treating 8.0.38+, 8.4.1+, 5.7 and below, and 9+ as fixed

* Adding a comment clarifying that the < 8.0.38 check is a conservative safeguard

Testing:

* Verified versions of MySQL 5.7, 8.0.1, 8.0.38, 8.1, 8.4.0, 8.4.7 and 9.5.0 by running this unit test against them.

* Confirmed that tests pass for versions that are fixed and correctly fail for versions that still have the bug.

### DIFF
--- a/.github/workflows/ci-mysql.yml
+++ b/.github/workflows/ci-mysql.yml
@@ -24,7 +24,7 @@ jobs:
             'mysql:8.1',
             'mysql:8.2',
             # 'mysql:8.3', # Already tested in "ci-linux"
-            # 'mysql:8.4', # TODO: Tests never end
+            'mysql:8.4',
             # 'mysql:9.0', # Already tested in "ci-coverage"
             'mysql:9.1',
             'mysql:9.2',
--- a/test/esm/regressions/2052.test.mjs
+++ b/test/esm/regressions/2052.test.mjs
@@ -100,22 +100,10 @@ test(async () => {
   const mySqlVersion = await common.getMysqlVersion(connection);
 
   const hasIncorrectPrepareParameter = (() => {
+    const incorrectVersions = ['8.1.0', '8.2.0', '8.3.0', '8.4.0'];
     const { major, minor, patch } = mySqlVersion;
-    // > 9 fixed
-    if (major >= 9) return false;
-
-    // 8.0 fixed, but only if 8.0.38+
-    // NOTE: We assume < 8.0.38 might be buggy, even though some builds behave correctly (e.g., 8.0.1)
-    if (major === 8 && minor === 0) return patch < 38;
-
-    // 8.1, 8.2, 8.3 do not have the fix.
-    if (major === 8 && (minor === 1 || minor === 2 || minor === 3)) return true;
-
-    // 8.4 fixed, but only 8.4.1+ onward
-    if (major === 8 && minor === 4) return patch < 1;
-
-    // everything else (MySQL < 8) was always correct
-    return false;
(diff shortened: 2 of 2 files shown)


wellwelwel:
Hey @HappyZombies! 🙋🏻‍♂️

If there is a complex range of versions that have divergent results, wouldn't it be interesting to map only those that fail, simplifying the logic? Just as an example _(not tested)_:

```js
const hasIncorrectPrepareParameter = (() => {
  const incorrectVersions = [ /* '8.4.0', ... */ ];
  const { major, minor, patch } = mySqlVersion;

  return incorrectVersions.includes(`${major}.${minor}.${patch}`);
}
```

> [!NOTE]
> 
> Currently, **MySQL** version **8.4** is not tested with **GitHub Actions**:
> 
> https://github.com/sidorares/node-mysql2/blob/490502f9c0f9f0699256930fda04ea16a78c11df/.github/workflows/ci-mysql.yml#L27

dannyreg:
@wellwelwel that todo commenting out v8.4, due to tests running forever, is probably because of this method check! So this change should enable us to run 8.4! I will uncomment it out and see.

With that being said I think I will do the change as you suggest, 

I think right now just marking these as incorrect should suffice.

```js
    const incorrectVersions = ['8.1.0', '8.2.0', '8.3.0', '8.4.0'];
 ```
I will double check and make sure

**Comment by dannyreg** [comment]
Ok just checked. 

Version 8.0.0 doesn't have the bug, neither does 8.0.1, nor latest, which is 8.0.44

To me it seems it was introduced in 8.1.0, which would turn be 8.4.0, but was promptly fixed in 8.4.1

So the simple array check SHOULD be enough for this, I am confident in it.

```js
const incorrectVersions = ['8.1.0', '8.2.0', '8.3.0', '8.4.0'];
```
  

So in conclusion these versions really are the ONLY known versions that has this bug.

wellwelwel:
> So in conclusion these versions really are the ONLY known versions that has this bug.

If we notice additional versions, we can include them as needed. Thanks, @HappyZombies ✨

PR #3929, opened 2025-11-23
feat(pool): emit 'connectionQueueAcquired' event on connection dequeue fixes #3844

### LINKED ISSUE #3844: Feature Request - Queue Size Visibility
It would be very useful if it could provide an event or use the acquire event to trigger when a queued connection is pulled from the queue and acquired.
This will enable the determination of the current queue depth or size at any given moment, helping to diagnose connection pool bottlenecks and improve the observability of connection usage in real-time.

shubhanshu2103:
This PR introduces a new event, 'connectionQueueAcquired', on the connection pool. This event fires whenever a waiting request is pulled from the internal queue (_connectionQueue) and successfully acquires a free connection.

This addresses the request for better observability of connection pool bottlenecks by providing real-time data on the queue state at the exact moment a waiting request is served.

🌟 Key Changes

New Event Emission in lib/base/pool.js:

The BasePool.releaseConnection() method is modified to emit 'connectionQueueAcquired' when this._connectionQueue.length is greater than 0, after shifting the waiting callback but before calling it.

Event Payload:

The event payload provides the essential information needed for diagnostics:

{
  connection: connection, // The Connection object being handed to the request
  queueDepth: this._connectionQueue.length // The actual size of the queue *after* the request was dequeued
}

New Integration Test:

A new integration test file, test/integration/test-pool-queue-event.js, was added to fully validate the feature.

The test saturates the pool (limit 2), queues two additional requests (depth 2), and then releases the active connections, asserting that the event fires exactly twice with the correct descending queueDepth values of 1 and 0.

Implementation Details

The implementation in releaseConnection is minimal and targeted to avoid side effects:

} else if (this._connectionQueue.length) {
    cb = this._connectionQueue.shift();
    
    // New Event: Fired when request acquires connection from queue
    this.emit('connectionQueueAcquired', {
        connection: connection,
        queueDepth: this._connectionQueue.length 
    });
    
    process.nextTick(cb.bind(null, null, connection));
} else {
// ...

### DIFF
--- a/test/integration/test-pool-queue-event.js
+++ b/test/integration/test-pool-queue-event.js
@@ -1,0 +1,116 @@
+'use strict';
+
+const assert = require('assert');
+// Corrected path based on file structure: ../common.test.cjs
+// const {test} = require('../common.test.cjs'); 
+const { test } = require('poku'); 
+const mysql = require('../..');
+const process = require('process'); // Ensure process is available
+
+// Helper function to create a pool using test environment variables
+function createTestPool(options) {
+  const config = {
+    host: process.env.MYSQL_HOST,
+    user: process.env.MYSQL_USER,
+    password: process.env.MYSQL_PASSWORD,
+    database: process.env.MYSQL_DATABASE,
+    ...options,
+  };
+  return mysql.createPool(config);
+}
+
+// --- Test Case ---
+
+test('Pool emits connectionQueueAcquired event and reports correct queue depth', function (done) {
+  // 1. Setup Pool: Small limit to force queuing
+  const connectionLimit = 2;
+  const pool = createTestPool({
+    connectionLimit: connectionLimit,
+    waitForConnections: true,
+    queueLimit: 0, // No queue limit
+  });
+
+  let acquiredConnections = [];
+  let queuedAcquiredEvents = [];
+  const expectedQueueDepths = [1, 0]; // Expected depth after 1st dequeue, then 2nd
+
+  // 2. Listen for the new event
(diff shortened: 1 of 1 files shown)


shubhanshu2103:
@wellwelwel please let me know the improvements needed.

test/integration/test-pool-queue-event.js
```diff
@@ -0,0 +1,116 @@
+'use strict';
+
+const assert = require('assert');
```
wellwelwel: Are there a specific limitation or motivation for using **Node.js** assert in this test? For example:

```js
const { assert, test } = require('poku');
```

test/integration/test-pool-queue-event.js
```diff
@@ -0,0 +1,116 @@
...
+  .catch(err => {
+    pool.end(() => done(err));
+  });
+});
+
+// For the purposes of this test, we don't need the queued connections to be stored,
+// as they are handled and released inside the final Promise block via the event payload.
+function queuedConnectionCallback(err, conn) {
+    if (err) throw err
+}
```
wellwelwel: I recommend declaring functions/variables before they are used 🙋🏻‍♂️

wellwelwel (commented):
Hey, @shubhanshu2103! I've suggested some ideas in the added test, but I'd suggest implementing the feature itself before moving on to the tests (or discussing the feature implementation in issue #3844) 🤝

wellwelwel:
> [!NOTE]
> I'm marking this _PR_ as a draft until the feature is implemented 🚧

PR #3958, opened 2025-12-10
feat(BaseConnection): add state getter to track connection lifecycle

### LINKED ISSUE #3764: Request: Public API for connection validation

CHOIJEWON:
## Summary

Closes #3764

Adds a `state` getter to `Connection` for tracking connection lifecycle states

## Motivation

Users need a reliable way to monitor connection status for debugging and health checks. Previously, there was no standard property
to check the current connection state.

## Changes

- **New property**: `connection.state` (readonly)
- **States**: `disconnected` → `protocol_handshake` → `connected` → `authenticated` / `error`
- **TypeScript**: Added `ConnectionState` type definition
- **Tests**: 11 unit tests covering all states and priority rules

## Usage

```javascript
const connection = mysql.createConnection(config);

console.log(connection.state); // 'disconnected'

connection.connect((err) => {
  if (err) {
    console.log(connection.state); // 'error'
  } else {
    console.log(connection.state); // 'authenticated'
  }
});

State Priority

Error > Disconnected > Authenticated > Connected > Protocol Handshake

Test Results

✔ All 11 state transition tests passing

### DIFF
--- a/test/unit/connection/test-connection-state.test.cjs
+++ b/test/unit/connection/test-connection-state.test.cjs
@@ -1,5 +1,5 @@
 'use strict';
-const { assert } = require('poku');
+const { test, assert } = require('poku');
 const BaseConnection = require('../../../lib/base/connection.js');
 const ConnectionConfig = require('../../../lib/connection_config.js');
 const EventEmitter = require('events');
@@ -11,6 +11,7 @@ function createMockConnection() {
     user: 'test',
     password: 'test',
     database: 'test',
+    connectTimeout: 0,
   });
 
   // Create a minimal mock stream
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by wellwelwel** [comment]
Thanks, @CHOIJEWON! Sorry for the delay, I'm _swamped_ with end-of-year tasks 🙋🏻‍♂️

I'm following the mocks approach, I would probably just change it from a comment followed by a numbered variable (`conn1`, `2`, `3`...), to a `test('comment content', () => { const conn = ... })`.

I'm also noticing that the `"disconnected"` status isn't being reached by the tests. Probably because of this logic inside `assert`:

```js
initialState === 'disconnected' || initialState === 'protocol_handshake'
```

> [!TIP]
> Using [**esm**](https://github.com/sidorares/node-mysql2/tree/632fef827d0fb03d8d6609862faf4caf1a71263e/test/esm/unit) may simplify testing with `Promises`, as well as allowing `top-await`.

**Related:**

- #3885

CHOIJEWON:
@wellwelwel 

That makes sense. Thank you for the feedback, and I’ll make the updates as quickly as I can.

CHOIJEWON:
@wellwelwel 

I completely understand that things can be busy at the end of the year
Whenever you have time please feel free to take a look at the updated commit
Wishing you a wonderful end of the year 👍

wellwelwel:
Thanks again, @CHOIJEWON! I also wish you a great end of the year 🙋🏻‍♂️

PR #4011, opened 2026-01-15
fix(named-placeholders): improve handling of mixed/nested quotes in query parsing

Tushar-Banswal-spraxa:
## Upgrade `named-placeholders` from version 1.1.3 to 1.1.6

We are using the `mysql2` package in our application, and currently maintain a custom patch to address the following bug:

- https://github.com/mysqljs/named-placeholders/pull/39

This issue has been resolved in `named-placeholders` version **1.1.6**. If `mysql2` is updated to use this latest version, we will be able to remove our custom patch or resolution.

This is a **patch-level upgrade** and does **not introduce any breaking changes**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "iconv-lite": "^0.7.0",
     "long": "^5.2.1",
     "lru.min": "^1.0.0",
-    "named-placeholders": "^1.1.3",
+    "named-placeholders": "^1.1.6",
     "seq-queue": "^0.0.5",
     "sqlstring": "^2.3.2"
   },


wellwelwel:
Hi and thanks, @Tushar-Banswal-spraxa 🙋🏻‍♂️

The **named-placeholder** is already updated in the **package-lock.json**:

> https://github.com/sidorares/node-mysql2/blob/e58d520efbd9abad9a3f9592853cdc8ff1753a57/package-lock.json#L3777-L3779

For new installations, it already comes in the same version as above (`v1.1.6`), due to the `^` prefix.

For users who already have MySQL2 installed, updating sub-dependencies doesn't usually happen by itself and depends on `npm update` (or `npm update named-placeholders`, in this case).

---

On the other hand, it's interesting to consider that **named-placeholders** is not "just" a sub-dependency: it is exposed to the end user and its patches are directly reflected in **MySQL2** and its users.

I believe that **named-placeholder** new releases should simultaneously generate new releases for **MySQL2**, just like the development dependencies we use (e.g., **ESLint**, **Docusaurus**, **React**), and they probably could receive the same or similar commit messages, for example:

```
fix(named-placeholders): improve handling of mixed/nested quotes in query parsing
```

> [!NOTE]
> 
> Using `build(deps):` prefix won't force a new release.

Issue #4022, opened 2026-01-21
Fix hard-coded test database in assertions


HappyZombies:
**Situation:** 

[The test suite allows configuring and connecting to a specific database name](https://github.com/sidorares/node-mysql2/blob/master/test/common.test.cjs#L14) (with env variable of MYSQL_DATABASE), but some tests [still have assertions that are hard-coded to expect the database name to be test.](https://github.com/sidorares/node-mysql2/blob/master/test/esm/integration/connection/test-column-inspect.test.mjs#L98)

As a result, even when a different database is correctly provided and used, certain tests fail because the assertions do not reflect the configured database. This can create a mismatch between supported behavior and test expectations.

dannyreg:
I brought this up in [another issue](https://github.com/sidorares/node-mysql2/issues/3854), but decided to make this issue so that I can have an issue attached to it when I make an MR to fix it :)

PR #4030, opened 2026-01-22
ci: resolve issue where some unit tests were hard coding the database name when it is meant to be configurable

HappyZombies:
Resolves https://github.com/sidorares/node-mysql2/issues/4022, I honestly thought there were a lot more but there were only two unit tests that had this kind of check hard coded.


wellwelwel:
Thanks, @HappyZombies! 🙌🏻

Issue #4009, opened 2026-01-14
Setting connection flag of "DEPRECATE_EOF" causes MySQL Server v8.0.44 to throw an error?


HappyZombies:
This could just be that it's not supported, (or maybe I am just doing something wrong??) but I am messing around with some nuances around the difference of when MySQL includes an EOF packet or omits it + sends OK Packet instead.

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query_response_text_resultset.html

Anyways I noticed that by default the client capability of `DEPRECATE_EOF` is not set by default https://github.com/sidorares/node-mysql2/blob/master/lib/connection_config.js#L225

When I set this by using the flags property in the createConnection method, the MySQL Server sends back an error immediately after its first successful query. Looking at wireshark I don't see the client sending anything, so I am not entirely sure why the server sent the error back... but regardless setting the DEPRECATE_EOF flag does indeed switch/change the textresult response to include OK Packet instead.

Note that the changes DID in deed go into affect (OK Packet as final packet)
<img width="1283" height="528" alt="Image" src="https://github.com/user-attachments/assets/8637420d-2412-4dd0-b5d6-40e2bb14fa82" />

Code Example:

```js
/* eslint-disable no-console */
const mysql = require('mysql2/promise');

async function getConnection() {
    return mysql.createConnection({
        host: 'localhost',
        user: 'root',
        password: 'password',
        flags: ["DEPRECATE_EOF"]
    });
}

async function main() {
    console.log("Let's do this.");
    const connection = await getConnection();
    await connection.query(`SELECT 1 LIMIT 0, 200;`);
    console.log("Done.");
    await connection.end();
}

main().catch(console.error);

```

MySQL Version 8.0.44

It could be that this just isn't supported at all on the client yet?

Error from MySQL Server:

```text
MySQL Protocol - response ERROR
    Packet Length: 51
    Packet Number: 0
    Response Code: ERR Packet (0xff)
    Error Code: 1158
    SQL state: 08S01
    Error message: Got an error reading communication packets
```

Node Error (mysql client simply just caught the error):
```text
Error: Unexpected packet while no commands in the queue
    at Connection.protocolError

...
...
Emitted 'error' event on Connection instance at:
    at Connection.protocolError


 fatal: true,
  code: 'PROTOCOL_UNEXPECTED_PACKET'
```

dannyreg:
I have investigation updates:

[...]

PR #4033, opened 2026-01-25
fix(constants): Remove unsupported CLIENT_DEPRECATE_EOF flag from constants

HappyZombies:
This PR removes the CLIENT_DEPRECATE_EOF client capability (stored as `DEPRECATE_EOF ` in code), which is currently advertised but not supported and can cause protocol errors. It also adds a regression test to ensure only valid client capability flags are applied and unknown flags are ignored.

This PR resolves https://github.com/sidorares/node-mysql2/issues/4009, view the item for more details and the investigation done for this.


wellwelwel:
Hey, @HappyZombies! I can't see the need to include a test that checks for a non-existent variable, since it has been permanently removed. I think a test for this scenario would be a connection that fails when using the `DEPRECATE_EOF` flag, and when removing it, the test ends successfully.

For example, "capturing" this log from #4009:

```
MySQL Protocol - response ERROR
  Packet Length: 51
  Packet Number: 0
  Response Code: ERR Packet (0xff)
  Error Code: 1158
  SQL state: 08S01
  Error message: Got an error reading communication packets
```

Then, testing the same connection + query without the flag and this time, it should pass.

What do you think?

dannyreg:
@wellwelwel I'm good with that change, however this means I would need to leave the constant value of DEPRECATE_EOF in order to have test that checks for a failure on on/off, since the only way to do a test like that after removing the constant would be to manually modify the 'config.clientFlags', or other work arounds. 

I’m okay doing that if we think it’s the more valuable test, just wanted to call out that it would rely on directly modifying internal config rather than using a supported option. In which case I can just do this test since the current one isn't _that_ valuable 

Now of course unless you think it's beneficial to leave that non working constant of DEPRECATE_EOF flag ...

wellwelwel:
> Now of course unless you think it's beneficial to leave that non working constant of DEPRECATE_EOF flag

I would suggest removing the `test/integration/connection/test-flags.test.cjs` test before merging. The reproduction test I mentioned would be more beneficial for the issue itself (#4009), as proof of concept.

dannyreg:
@wellwelwel I removed the test, so simply removed the capability and now the common.test.js file allows for the flags property to be passed.

PR #4034, opened 2026-01-25
fix(types): Add missing ConnectionState type to Promise Connection interface

### LINKED ISSUE #4005: interface Connection missing state property in promise.d.ts
mysql2 3.16.0 add a new property named state on Connection class, but it is missed on promise.d.ts.

minzunim:
fix(types): Add missing state property to Promise Connection interface

In 3.16.0, a new `state` property was added to the callback-based Connection class to allow users to check the connection state. However, this property was missing from the Promise-based Connection interface type definition.

This change adds the `readonly state: ConnectionState` property to align the Promise Connection interface with the callback-based API.

Changes:
- Added ConnectionState import to promise.d.ts
- Added readonly state: ConnectionState property to Promise Connection interface
- Exported ConnectionState type from typings/mysql/index.d.ts for centralized type management

Closes #4005

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -10,7 +10,7 @@ import {
   PoolOptions,
   PoolClusterOptions,
   Pool as CorePool,
-  ConnectionState
+  ConnectionState,
 } from './index.js';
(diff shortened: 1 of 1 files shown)


wellwelwel:
Thanks, @minzunim! ✨

Issue #1176, opened 2020-07-02
[Question]  non replaceable question mark 


mogadanez:
How can I use question mark as literal in query?
for example 
```(LOCATE('?', REVERSE (url), 1)```
I  want to use it as real `?` string, not a parameter

sidorares:
can you use prepared statements? With `connection.execute(query, parameters)` query is prepared on the server side and thus parsed with sql engine. On the other hand, `connection.query(query, parameters)` parameters substitution is performed on the client and there is no plan to bundle full fledged sql parser with this driver ( you a free to BYO of course ).

sidorares:
Another solution is to put placeholder there but use `"?"` parameter value for that placeholder - `query("... (LOCATE(?, REVERSE (url), 1)...", [..., "?", ... ])`

mogadanez:
> Another solution is to put placeholder there but use "?" parameter value for that placeholder - query("... (LOCATE(?, REVERSE (url), 1)...", [..., "?", ... ])
yep already did this

doni-wicaksana:
Sorry my English is bad. But I found the issue is here:

node_modules/sqlstring/lib/SqlString.js
line 86: var placeholdersRegex = /\?+/g;
Should be:
line 86: var placeholdersRegex = /\?(?![^'"]*['"])/g;

PR #4054, opened 2026-02-04
fix(security): resolve a potential SQL injection bypass through objects

### LINKED ISSUE #1914: Object Injection in "prepared" statement.
I believe the current library is vulnerable to object injection. The library this was based off of explains it here: https://github.com/mysqljs/mysql#escaping-query-values, and the author's stance created a rather humorous git issue back in 2014.  https://github.com/mysqljs/mysql/issues/731.

I validated this against the example code in the library itself:
```js
// get the client

// create the connection to database
  host: 'localhost',
  user: 'root',
  password: 'password',
  database: 'injection'
});

// Imagine this is req.body.password
var password = {
  password: true
};

// with placeholder
connection.query(
  'SELECT * FROM `users` WHERE `username` = ? AND `password` = ?',
  ['admin', password],
  function(err, results) {
    console.log(results);
  }
);
```

This will convert the select query to: 
```
SELECT * FROM `users` WHERE `username` = 'admin' AND `password` = `password` = true,
```

For further validation, you can change true to false and you'll get no results.

### LINKED ISSUE #2528: insert can use Buffer, but not work with Uint8Array
when use Uint8Array , will cause error (the hash column)

```

Error: insert ignore into `v` (`css`, `hash`, `idLi`, `js`) values (151, {"0":114,"1":108,"2":86,"3":55,"4":10,"5":74,"6":187,"7":109,"8":165,"9":203,"10":90,"11":226,"12":165,"13":144,"14":143,"15":33,
"16":98,"17":122,"18":116,"19":22,"20":231,"21":192,"22":78,"23":37,"24":13,"25":131,"26":169,"27":153,"28":129,"29":125,"30":136,"31":157}, X'01010101010101010101010101010101010101010101010101010101010
1010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010
10101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101', 38) - Unknown column '0' in 'field list'
    at Packet.asError (/Volumes/d/i18n/node/s3dist/node_modules/.pnpm/mysql2@3.9.2/node_modules/mysql2/lib/packets/packet.js:728:17)
    at Query.execute (/Volumes/d/i18n/node/s3dist/node_modules/.pnpm/mysql2@3.9.2/node_modules/mysql2/lib/commands/command.js:29:26)
    at Connection.handlePacket (/Volumes/d/i18n/node/s3dist/node_modules/.pnpm/mysql2@3.9.2/node_modules/mysql2/lib/connection.js:481:34)
    at PacketParser.onPacket (/Volumes/d/i18n/node/s3dist/node_modules/.pnpm/mysql2@3.9.2/node_modules/mysql2/lib/connection.js:97:12)
    at PacketParser.executeStart (/Volumes/d/i18n/node/s3dist/node_modules/.pnpm/mysql2@3.9.2/node_modules/mysql2/lib/packet_parser.js:75:16)
    at Socket.<anonymous> (/Volumes/d/i18n/node/s3dist/node_modules/.pnpm/mysql2@3.9.2/node_modules/mysql2/lib/connection.js:104:25)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
  code: 'ER_BAD_FIELD_ERROR',
  errno: 1054,
  sqlState: '42S22',
  sqlMessage: "Unknown column '0' in 'field list'",
  sql: "insert ignore into `v` (`css`, `hash`, `idLi`, `js`) values (151, `0` = 114, `1` = 108, `2` = 86, `3` = 55, `4` = 10, `5` = 74, `6` = 187, `7` = 109, `8` = 165, `9` = 203, `10` = 90, `11` = 226,
 `12` = 165, `13` = 144, `14` = 143, `15` = 33, `16` = 98, `17` = 122, `18` = 116, `19` = 22, `20` = 231, `21` = 192, `22` = 78, `23` = 37, `24` = 13, `25` = 131, `26` = 169, `27` = 153, `28` = 129, `29
` = 125, `30` = 136, `31` = 157, X'01010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010
101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101', 38)"
}
```

### LINKED ISSUE #3481: prepared statements format error when filed is `JSON` type
```
let query='select  *  from ??  where (??=?) limit ?,? '
let params = ['certification', "cert->>'$.name'", 'myname', 0, 20];
let pool = mysql.createPool(poolOption);
let sql = pool.format(query, params);
//  the sql is not correct.
// correct query sql should be: select * from certification where cert->>'$.name' = ‘myname’ limit 0, 20
```

`pool.format` method will finally call `escapeId(val, forbidQualified)` in `SqlString`.
the sqlstring module not update now.
file: `node_modules\sqlstring\lib\SqlString.js`#18, code below:

```
SqlString.escapeId = function escapeId(val, forbidQualified) {
  if (Array.isArray(val)) {
    var sql = '';

    for (var i = 0; i < val.length; i++) {
      sql += (i === 0 ? '' : ', ') + SqlString.escapeId(val[i], forbidQualified);
    }

    return sql;
  } else if (forbidQualified) {
    return '`' + String(val).replace(ID_GLOBAL_REGEXP, '``') + '`';
  } else {
    return '`' + String(val).replace(ID_GLOBAL_REGEXP, '``').replace(QUAL_GLOBAL_REGEXP, '`.`') + '`';
  }
};
```

### LINKED ISSUE #4051: Security Issue: Change stringifyObjects default value to true
Alarm bells were raised about this in 2022 [1], 2024 [2] and 2025 [3]. The _stringifyObjects_ default is set to _false_, which makes literally thousands of applications susceptible to SQL Injection.

If the default value had been set to _true_ instead of the current default _false_, it would have protected all new applications from this vulnerability, which was introduced in _node-mysql_ and later replicated in _node-mysql2_.

All technical details can be found in [1] and [2]. It is alarming that, although this issue was raised, it has not been fixed since.

[1] https://flattsecurity.medium.com/finding-an-unseen-sql-injection-by-bypassing-escape-functions-in-mysqljs-mysql-90b27f6542b4
[2] https://github.com/sidorares/node-mysql2/issues/1247#issuecomment-1958043226
[3] https://blog.mantrainfosec.com/blog/18/prepared-statements-prepared-to-be-vulnerable

wellwelwel:
> [!IMPORTANT]
> 
> Normally, this would be handled with much more care, but considering that it is a vulnerability that has been exploited since 2022 in a completely public manner, I also decided to resolve it publicly.

Fixes #1914, Fixes #4051, Fixes #2528, Fixes #3481.

---

## Vulnerability Exploitation

- https://flattsecurity.medium.com/finding-an-unseen-sql-injection-by-bypassing-escape-functions-in-mysqljs-mysql-90b27f6542b4
- https://blog.mantrainfosec.com/blog/18/prepared-statements-prepared-to-be-vulnerable

---

## Changes

This fix doesn't change **MySQL2**'s behavior, nor does it introduce any kind of breaking change.

Unlike what's recommended in the two exploits _(forcing the use of `true` in `stringifyObjects`)_, this doesn't fix the reported vulnerability, it just "hides" it.

I researched several scenarios where it's actually necessary to convert values derived from objects, and there are the three safe/expected situations:

<blockquote>

#### ✅ UPDATE

```ts
const query = format('UPDATE users SET ?', [
  { name: 'foo', email: 'bar@test.com' },
]);
```

#### ✅ INSERT + SET

```ts
const query = format('INSERT INTO users SET ?', [
  { name: 'foo', email: 'bar@test.com' },
]);
```

#### ✅ INSERT ... KEY UPDATE

```ts
const query = format(
  'INSERT INTO users (name, email) VALUES (?, ?) ON DUPLICATE KEY UPDATE ?',
  ['foo', 'bar@test.com', { name: 'foo', email: 'bar@test.com' }]
);
```

</blockquote>

❗️ For **SELECT**, conventional **INSERT**, **DELETE**, etc., they will always cause a potential **SQL injection**. For example:

```ts
const query = format('SELECT * FROM users WHERE email = ?', [{ email: 1 }]);

// ❌ SELECT * FROM users WHERE email = `email` = 1
```

---

## The replacement of [sqlstring](https://github.com/mysqljs/sqlstring) (and why)

Mainly motivated by this comment: https://github.com/mysqljs/sqlstring/issues/49#issuecomment-571618184, **sqlstring** must be compatible with **Node.js v0.6**. This limits the natural improvement of new language features, such as the addition of `BigInt` (v10.4) and `Uint8Array` (v0.10).

---

### Security + Breaking Change Conflicts

I read the discussion around #1914 and evaluated the ideas:

- Remove `stringifyObjects` entirely, create an alternative feature such as `sqlstring.SET`, etc.

I thought carefully about the issue and the time that has passed without resolution (~3 years), probably due to the challenge of negotiating a breaking change that affects so many users and packages simultaneously.

---

### The solution

Instead of introducing a breaking change, I just handled the parameters properly, for example:

- Performing an **AST**-based approach, this way it captures both spaces, line breaks and tables, doesn't depend on `RegExp`, supports comments and understands when key values are used inside a string 🔥

Then, in short:

```ts
// Lazy: compute SET position only when we first encounter an object
if (setIndex === -2) setIndex = findSetKeyword(sql);

if (
  setIndex !== -1 &&
  setIndex < placeholderPosition &&
  !hasSqlString(currentValue) &&
  !Array.isArray(currentValue) &&
  !Buffer.isBuffer(currentValue) &&
  !isDate(currentValue) &&
  isRecord(currentValue)
) {
  escapedValue = objectToValues(currentValue, timezone);
  setIndex = -1;
}
```

This way, `SET` or `KEY UPDATE` provide the expected functionality, while `SELECT`, _common_ `INSERT`, etc., return the same behavior when `stringifyObjects` is enabled, for example `[Object ...]`. Unlike a breaking change, these queries should no longer process the object values in any case, after all, that is where the vulnerability lies.

---

### Long term support and pros

It will enable us to move forward with fixing issues #2528, #3481, and #1176, including support for `BigInt` (https://github.com/mysqljs/sqlstring/issues/49).

---

### ⚡️ Performance

Not only the language resource point, but also performance limitations, where **SQL Escaper** outperforms **sqlstring** up to **40%**:

<blockquote>

Each benchmark formats `10,000` queries using `format` with `100` mixed values (numbers, strings, `null`, and dates), comparing **SQL Escaper** against the original [**sqlstring**](https://github.com/mysqljs/sqlstring) through [**hyperfine**](https://github.com/sharkdp/hyperfine):

| Benchmark                                | sqlstring | SQL Escaper |       Difference |
| ---------------------------------------- | --------: | ----------: | ---------------: |
| Select 100 values                        |  248.8 ms |    178.7 ms | **1.39x faster** |
| Insert 100 values                        |  247.5 ms |    196.2 ms | **1.26x faster** |
| SET with 100 values                      |  257.5 ms |    205.2 ms | **1.26x faster** |
| SET with 100 objects                     |  348.3 ms |    250.5 ms | **1.39x faster** |
| ON DUPLICATE KEY UPDATE with 100 values  |  466.2 ms |    394.6 ms | **1.18x faster** |
| ON DUPLICATE KEY UPDATE with 100 objects |  558.2 ms |    433.9 ms | **1.29x faster** |

</blockquote>

---

## Considerations

I still consider that this problem should indeed be solved on the **sqlstring** side, but unfortunately I was unable to contribute due to the limitations of the environment (**Node.js v0.6**): `String.prototype.search()`, for example, requires version **0.10** 😅

I believe that **MySQL2** has evolved differently from **mysqljs/mysql**. Even if we ignore the security issue, I still believe that MySQL2 has a lot to gain from the level of performance that new approaches can offer.

**SQL Escaper** is also compatible with **mysqljs/mysql** as usual, but requires at least **Node.js v12**.

> [!TIP]
> 
> With this idea, for example, we can discuss different approaches, such as removing it entirely, etc., with the project already safely and available in the meantime.


wellwelwel:
> cc @sidorares 

If you like the solution, I would be happy to assign you as a maintainer, as was done with **AWS SSL Profiles**. I'm totally open to migrating the repository to **mysqljs** organization or what you think is best.

---

> cc @dougwilson 

It will be a pleasure to contribute to **sqlstring**, regardless of the direction this _PR_ takes. Also, from docs:

<blockquote>

## Acknowledgements

- **SQL Escaper** is adapted from [**sqlstring**](https://github.com/mysqljs/sqlstring) ([**MIT**](https://github.com/mysqljs/sqlstring/blob/master/LICENSE)), modernizing it with high performance, TypeScript support and multi-runtime compatibility.
- Special thanks to [**Douglas Wilson**](https://github.com/dougwilson) for the original **sqlstring** project and its [**contributors**](https://github.com/mysqljs/sqlstring/graphs/contributors).

</blockquote>

---

> @mantrainfosec and all the community

The proposed repository is: **https://github.com/mysqljs/sql-escaper**.

It has the same strictness as [**lru.min**](https://github.com/wellwelwel/lru.min), with 100% coverage of real tests, passing in all **sqlstring** tests (and including new ones too, specially considered for **MySQL2**), 100% typed, and fully documented too.

Please, feel free to review 💙

dougwilson:
I have not really been in Node.js land for some time now and am open to delegating the modules to someone who is / the mysqljs org too.

wellwelwel:
> I have not really been in Node.js land for some time now and am open to delegating the modules to someone who is / the mysqljs org too.

@dougwilson, although I'm not part of **mysqljs** organization, I'm open to helping in any way I can 🙋🏻‍♂️

In this specific case, I didn't simply think about creating a different project, I recognized that there is a really strong philosophy behind **sqlstring** and **mysqljs/mysql** that must be respected.

wellwelwel:
It's finally ready 🎉

Now it uses an **AST**-based approach to:

- Support multi lines, spaces and tables.
- Support SQL comments, including multi line comments.
- Distinguish when a keyword is used as value.
- Distinguish between multiple clauses/keywords in the same query.
- Distinguish when a column has a keyword name.

New features:

- Supports `Uint8Array` and `BigInt` ✨
- Preserves `JSON` path expressions

> As proof of concept, I fixed #2528 and #3481 too 🙋🏻‍♂️

---

For those concerned about this issue, you can use an overrides in your **package.json**:

```json
"dependencies": {
  "mysql2": "^3.16.3"
},
"overrides": {
  "sqlstring": "npm:sql-escaper"
}
```

- Then, clean the `node_modules` and run `npm i`.

Now, it will work normally for `UPDATE`, `INSERT...SET` and `INSERT...KEY UPDATE`, but it will not compromise queries as reported in https://github.com/sidorares/node-mysql2/issues/4051#issuecomment-3850596869.

> [!TIP]
> 
> The same works for **mysqljs/mysql**, but it requires at least **Node.js v12**.

sidorares:
@wellwelwel sorry just found time to get myself up to date with recent discussions

I'm pretty sure node versions can be listed a bit for sqlstring. We can stay reasonable conservative but v12 is probably quite reasonable

If you believe modernized code is 100% compatible with sqlstring surface api mo vote is to just replace it with your version
I invited you to mysqljs org - let me know if you need anything else

wellwelwel:
> If you believe modernized code is 100% compatible with sqlstring surface api mo vote is to just replace it with your version
> I invited you to mysqljs org - let me know if you need anything else

Thanks, @sidorares! All **sqlstring** tests pass, but if there is any unnoticed usage, I mentioned in the documentation that this must be treated as a bug and reported/fixed (except for potentially vulnerable behavior, naturally).

I'm polishing things up after the migration, before continuing here. Thanks again!

wellwelwel:
> I'm polishing things up after the migration, before continuing here. Thanks again!

Well _(not me)_, I went over every project detail, created even more tests for new use cases, and so far, I really don't see anything different from **sqlstring** for final users.

If any conflicts arise, I believe users will report them. From now on, it's ready to merge 🤝

ZainabAl-khabori:
I would like to report a case, if that's acceptable here.

- **This won't work:** format("set @isActive = true; insert into Branches set ?;", { id: "abcdef123456", location: "HQ" });
- **But this will:** format("insert into Branches set ?; set @isActive = true;", { id: "abcdef123456", location: "HQ" });

After some testing, I came to the conclusion that maybe the @ in the variable is throwing regex off for the placeholders that come after it, since this issue only appears if the variable came before a placeholder that substitutes an object.

wellwelwel:
Thanks @ZainabAl-khabori! This was fixed in https://github.com/mysqljs/sql-escaper/pull/34 and I included your examples to tests in https://github.com/mysqljs/sql-escaper/pull/36.

Please, feel free to open issues directly in [**SQL Escaper**](https://github.com/mysqljs/sql-escaper) if you encounter any errors. I'm revisiting issue #4126 before releasing this fix on **npm** 🙋🏻‍♂️

Issue #2050, opened 2023-06-07
read-only error on aws aurora mysql failover action


itjobs-levi:
When failover of aws aurora mysql occurs, "The MySQL server is running with the --read-only option so it cannot execute this statement" error occurs frequently.
I think it is a temporary phenomenon caused by the promotion of the read node.
However, the nodejs connection pool continuously tries to write to the read node.
It could be that the previous write node was accessed by aws aurora endpoint dns caching.
When aws aurora failover occurs, we have to restart the application to be called as a normal writing node.
Is there any workaround for this issue?

anton-zen:
I also encountered this. Seem like the jdbc friver for Java is doing a better job at this https://github.com/awslabs/aws-mysql-jdbc#the-aws-jdbc-driver-failover-process

Rayyan98:
@itjobs-levi did you find anything to get around the issue we just faced it twice in the last couple of days

itjobs-levi:
I haven't found a solution to this problem :(
Since we are using nodejs, the java jdbc driver did not provide insight.
We designed it so that failovers rarely occur and that when a failover occurs, the application can be recovered and run.

PR #4075, opened 2026-02-12
fix(pool): discard connection on error 1290 (Aurora read-only failure)

natetarrh:
## Problem

During an Aurora MySQL failover, an existing writable connection becomes read-only. Subsequent writes return error 1290 (`ER_OPTION_PREVENTS_STATEMENT`):

> The MySQL server is running with the --read-only option so it cannot execute this statement

The pool returns this stale read-only connection to the free list, so all subsequent callers receive the same broken connection.

Fixes https://github.com/sidorares/node-mysql2/issues/2050

## Prior Art

The Go MySQL driver handles this same scenario by closing the connection and returning `ErrBadConn` when it encounters errors 1290, 1792, or 1836 during failover. This causes Go's `database/sql` pool to discard the connection and create a new one:

https://github.com/go-sql-driver/mysql/blob/76c00e35a8d48f8f70f0e7dffe584692bd3fa612/packets.go#L599-L612

## Fix

`pool.query()` and `pool.execute()` now detect error 1290 and call `conn.destroy()` instead of `conn.release()`, forcing the pool to create a fresh connection for the next caller.

## Test

Adds an integration test (`test-aurora-failover-readonly.test.cjs`) that simulates the failover using a mock server:
1. Pool gets a connection, first query succeeds
2. Mock server starts returning error 1290 (simulating failover)
3. Asserts the pool discards the bad connection and provides a new one (different `threadId`)

### DIFF
--- a/test/integration/connection/test-aurora-failover-readonly.test.cjs
+++ b/test/integration/connection/test-aurora-failover-readonly.test.cjs
@@ -19,6 +19,12 @@ if (typeof Deno !== 'undefined') process.exit(0);
 let connId = 0;
 let failoverTriggered = false;
 
+const readOnlyError = {
+  code: 1290,
+  message:
+    'The MySQL server is running with the --read-only option so it cannot execute this statement',
+};
+
 const server = mysql.createServer((conn) => {
   conn.on('error', () => {});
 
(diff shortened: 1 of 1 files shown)


wellwelwel:
Thanks, @natetarrh! Going forward, we can cover all read-only cases mentioned by **Go-MySQL-Driver**, for example:

```js
// Source: https://github.com/go-sql-driver/mysql/blob/76c00e35a8d48f8f70f0e7dffe584692bd3fa612/packets.go#L598-L613
function isReadOnlyError(err) {
  if (!err || !err.errno) {
    return false;
  }

  // 1792: ER_CANT_EXECUTE_IN_READ_ONLY_TRANSACTION
  // 1290: ER_OPTION_PREVENTS_STATEMENT (returned by Aurora during failover)
  // 1836: ER_READ_ONLY_MODE
  return (
    err.errno === Errors.ER_OPTION_PREVENTS_STATEMENT ||
    err.errno === Errors.ER_CANT_EXECUTE_IN_READ_ONLY_TRANSACTION ||
    err.errno === Errors.ER_READ_ONLY_MODE
  );
}
```

PR #4082, opened 2026-02-16
fix(pool): handle all read-only errors during Aurora failover

wellwelwel:
Continue with the work of #4075 by extending error `1290` handling to also cover errors `1792` and `1836`, matching **Go-MySQL-Driver** implementation for **Aurora** failover.

For reference:

- https://github.com/go-sql-driver/mysql/blob/76c00e35a8d48f8f70f0e7dffe584692bd3fa612/packets.go#L598-L613

### DIFF
--- a/lib/base/pool.js
+++ b/lib/base/pool.js
@@ -8,6 +8,21 @@ const Queue = require('denque');
 const BaseConnection = require('./connection.js');
 const Errors = require('../constants/errors.js');
 
+// Source: https://github.com/go-sql-driver/mysql/blob/76c00e35a8d48f8f70f0e7dffe584692bd3fa612/packets.go#L598-L613
+function isReadOnlyError(err) {
+  if (!err || !err.errno) {
+    return false;
+  }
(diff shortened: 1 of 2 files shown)


natetarrh:
Thanks for the follow-up! For posterity, I’ve only seen the 1290 error code in this situation (and the inline comments in the Go MySQL driver confirm this). My understanding is that Aurora read replicas set the [innodb_read_only](https://dev.mysql.com/doc/refman/8.4/en/innodb-parameters.html#sysvar_innodb_read_only) system variable and after a failover a write query to a preexisting connection to the old primary (which now has this variable set to `ON`) will fail with error code 1290 as a result. The other error codes are for similar situations (making a write query using a connection to a database that’s read-only) but implemented differently (not sure which situations result in the other two error codes here). Personally I still think it’s a good idea to handle all three. Note that the Go driver makes this behavior configurable–I’m not sure whether there’s a reasonable use case for opting out of reconnecting in these cases (maybe third-party code is trying to make write queries against a database that will always be read-only and we want to avoid new connection overhead?) but just want to mention it.

PR #3985, opened 2025-12-19
fix: types for query values

uPaymeiFixit:
Fixes https://github.com/sidorares/node-mysql2/issues/3982

Issue #3695, opened 2025-07-08
Improve contributors and users experience


wellwelwel:
Based on various contributions (including **Discussions**, **Issues** and **Pull Requests** — _also in private messages asking for help_), I would like to improve a few things where I feel the contributors have the most difficulty _(or that I notice, even when nothing is said)_.

These are my ideas:

### Contributing

- [ ] Including a template when opening **Issues** and **Pull Requests**.
- [ ] In addition to the [Contribution Guide](https://github.com/sidorares/node-mysql2/blob/master/Contributing.md), I believe it's really useful to have tips on how to create and modify tests for features and fixes.

### Automations

Several issues remain open because they have no means of reproduction. For some time now, I have categorized them by adding a `label' to identify them, but this doesn't send a notification to the person who opened the Issue or a "formal request".

- [ ] I think an automation to create a default comment on Issues that need reproduction after adding this label would be of great help and, as it is triggered manually, it wouldn't cause any spam.

### Testing

I think that reworking the test flow to be more practical is a good starting point:

- [ ] Unite all tests to `/test/esm/` (just `/test/` — no more mixed tests between **CJS**/`require` and **ESM**/`import`).

### Bun and Deno

- [ ] It's interesting to include the installation guide for these runtimes too, as well as highlighting some possible differences in use between them (e.g., [**@types/node** conflict for **Deno** with **TypeScript**](https://github.com/sidorares/node-mysql2/issues/3136)).

### Documentation

- [ ] Refer to the [**mysqljs/mysql**](https://github.com/mysqljs/mysql) documentation for sections and features that aren't present in the **MySQL2** documentation, suggesting contributions for their inclusion.
  - [ ] This also includes opening a dedicated Issue for unimplemented features, if found.
- [ ] I also consider that we could include a [/**llms.txt**](https://llmstxt.org/) with instructions in the website on where critical functions are and their use, directing the documentation with an improved **NLP** for **AI** users and editors.

sidorares:
all are really good points. For "contibuting" - would be good to include high level overview of the components and connection between them ( this should also help ai editor users to generate better code )

jeengbe:
Do you have plans for porting the entire lib/ code base to TypeScript? That was my single one greatest worry writing my PR. The project scope is manageable and I would be happy to help

wellwelwel:
[...]

Issue #2803, opened 2024-06-22
Typescript conversion


evert:
Hi @sidorares !

I was curious if there's an interest in a total typescript conversion of this project. I started this as an exercise/experiment and noticed the code is hella clean which makes this relatively easy (just a big grind, no big refactors).

If this is a goal of this project to eventually make this switch, I can do the conversion with a future pull request in mind, but if the intent is to keep this a Javascript project I will just keep my fork and make more dramatic changes.

wellwelwel:
I believe it would be interesting to list both positive and negative points 🙋🏻‍♂️

**A question:** Would the tests/tools also be converted or just the main code? 

Before the `v3.2.1` (where it was not possible to compile projects made with **TypeScript** using **MySQL2** without `skipLibCheck`), I started a fork with this goal in mind, but ended up turning it into an independent project (_ORM_).

---

_Some notes for this discussion:_

- Currently, there are **48** _PRs_ that would need to be converted to **TypeScript**.
- At this moment, the issue #2667 is the only one directly related to typings.
- Direct dependencies that have no typings:
  - [generate-function](https://www.npmjs.com/package/generate-function)
  - [named-placeholders](https://www.npmjs.com/package/named-placeholders)
  - [seq-queue](https://www.npmjs.com/package/seq-queue)
- Dev dependencies that have no typings:
  - [assert-diff](https://www.npmjs.com/package/assert-diff)

evert:
I may not be the best person to list all the pros and cons. I think the benefit for me is that I'm intending to write a new user-facing API for MySQL using `Symbol.iterator`, and I want to try my hand in analyzing SQL strings and generating types based on the database schema and sql queries to try to bring raw SQL queries into the type system.

Typescript gives me a very degree of confidence during refactoring, that plain Javascript does not give me so I figured since I'm doing this I might as well see if this project would benefit my grind ;)

I suppose another benefit is that it's much harder for the typings and code to get out of sync, leading to issues like #2667.

> A question: Would the tests/tools also be converted or just the main code?

I did not intend to touch the tests, because this gives me a great check ensuring that my conversion didn't break anything. I personally like having my tests in Javascript because so many tests check edge-cases and Typescript tends to complain a lot.

sidorares:
Good point about backlog of PRs / in progress branches that someone will also have to convert after big refactor

In general I don't mind this but definitely not high on my priority list.

[...]

PR #4086, opened 2026-02-17
ci: transcribe `esm` tests to TypeScript

wellwelwel:
The intention is, once 100% of the tests are transcribed to **TypeScript** in the future, to completely remove the **./test/tsc-build** directory.

---

### Advantages

- Easier to identify inconsistencies between **JavaScript** code and separately created types (`d.ts`) naturally while tests are being created.
- More than **20** typing inconsistencies were found just in **./test/esm** that went unnoticed by **./test/tsc-build** 🔥
- Makes it easier to review contributions such as #3321 and #3985.
- Looking ahead, it avoids test redundancy/mirroring due to types (one directory to test types without execution, another to test **JavaScript**).

---

> [!IMPORTANT]
> No files inside the core (**lib**, **index.js**, etc.) were modified and remain totally untouched 🙋🏻‍♂️

---

> [!NOTE]
> Since for **Poku** there is no difference between running **CJS**, **ESM**, or **TypeScript** files in the same test suite (regardless of the runtime or coverage), there is no need to make any adaptations to the source code (**lib**), while we can take the first step towards #2803.
> 
> The legacy version of **Deno** that used a _polyfill_ has been removed from the workflow since it is incompatible with `require` + dynamic typings.

---

### Related

- #3695

### DIFF
--- a/test/esm/common.test.mts
+++ b/test/esm/common.test.mts
@@ -1,3 +1,10 @@
+import type { Connection as PromiseConnection } from '../../promise.js';
+import type {
+  ConnectionOptions,
+  PoolOptions,
+  PoolClusterOptions,
+  Connection,
+} from '../../index.js';
 import { createRequire } from 'node:module';
 import fs from 'node:fs';
 import path from 'node:path';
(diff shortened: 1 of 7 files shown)


wellwelwel:
All type inconsistencies are marked with: `// @ts-expect-error: TODO: implement typings`. Some of them have already been reported, but without reproduction until now. For example:

- #3964
- #2667

From here on, types are no longer a compilation test and become an integral part of the tests that are actually executed ✨

PR #4091, opened 2026-02-17
ci: transcribe `cjs` tests to ES Modules and TypeScript

wellwelwel:
Continues the work from #4086 by migrating all tests to **ES Modules** + **TypeScript** 🎉

Since this is so intensive, I'll break it down into stages. This _PR_ will be exclusive for migrating the tests, not for refactoring them or deleting the `.cjs` files.

☔️ To be able to "navigate" through so many files and ensure there were no losses, I relied on coverage ratio.

---

As with #4086, no files outside tests were modified. Until the third stage, `.cjs` files will be kept (without any use) to assist in refactoring and to provide an accurate comparison to the original tests.

- There are **157** missing types mapped that went unnoticed by **./test/tsc-build** 😵‍💫
  - By searching for `// @ts-expect-error: TODO: implement typings` throughout the tests, fixes should be applied directly to the actual **MySQL2** types.

---

### Related

- #3695

---

> _Off the record, I need to say that this really took a lot of work_ 🥲

### DIFF
--- a/test/esm/integration/connection/test-quit.test.mts
+++ b/test/esm/integration/connection/test-quit.test.mts
@@ -27,7 +27,8 @@ const server = createServer(
       host: 'localhost',
       // @ts-expect-error: internal access
       port: server._port,
-      ssl: undefined,
+      // @ts-expect-error: TODO: implement typings
+      ssl: false,
     });
 
     connection.query(queryCli, (err, _rows, _fields) => {


wellwelwel:
In the third step, I'll remove all unused files (`.cjs`) and include a dedicated section for tests in the documentation (probably in [**Contributing.md**](https://github.com/sidorares/node-mysql2/blob/master/Contributing.md)).

PR #3937, opened 2025-11-30
fix: fix `PoolConnection.end` callback and promise resolution

CodeVoyager-bit:
This PR fixes an issue where calling [end()](cci:1://file:///Users/krish/Desktop/ap_alone/node-mysql2/node-mysql2/lib/promise/connection.js:62:2-66:3) on a [PoolConnection](cci:2://file:///Users/krish/Desktop/ap_alone/node-mysql2/node-mysql2/lib/base/pool_connection.js:4:0-66:1) would not execute the provided callback or resolve the promise when using the Promise API.

Previously, `PoolConnection.end()` did not accept any arguments, causing the callback to be ignored and the Promise wrapper to hang indefinitely.

## Changes
- Updated `PoolConnection.prototype.end` to accept a [callback](cci:1://file:///Users/krish/Desktop/ap_alone/node-mysql2/node-mysql2/lib/base/connection.js:813:4-824:5) parameter.
- Ensure [callback](cci:1://file:///Users/krish/Desktop/ap_alone/node-mysql2/node-mysql2/lib/base/connection.js:813:4-824:5) is executed after releasing the connection back to the pool (when `gracefulEnd` is false).
- Pass [callback](cci:1://file:///Users/krish/Desktop/ap_alone/node-mysql2/node-mysql2/lib/base/connection.js:813:4-824:5) to `super.end()` when `gracefulEnd` is true.

## Related Issue
Fixes #3915

Issue #3137, opened 2024-10-21
old unmaintained package with no license (seq-queue@0.0.5)


andres645:
Hi,
I noticed that the MySQL2 project relies on the package seq-queue@0.0.5, which has not been maintained for over a decade and lacks a license. The absence of a license poses potential legal and compliance risks.

Given the importance of this issue, I would like to suggest either:

Replacing seq-queue with a maintained and licensed alternative, or
Working with the original maintainers of seq-queue (if possible) to update the package with a proper license.

This issue can potentially impact the ability to use this package in a compliant manner, so addressing it would be beneficial for the broader community.

Looking forward to your thoughts and any steps to resolve this.

Thank you for maintaining MySQL2!

wellwelwel:
Hi, @andres645 🙋🏻‍♂️

12 years ago, [MIT license](https://github.com/changchang/seq-queue/blob/master/LICENSE) was added, but it hasn't been uploaded to **NPM**.

Do you have a suggestion? It's currently in use here:

https://github.com/sidorares/node-mysql2/blob/c06d2aea35cf2e45386bae5804d712f8902652ad/lib/compressed_protocol.js#L120-L122

wellwelwel:
> Working with the original maintainers of seq-queue (if possible) to update the package with a proper license.

Also, see:

- https://github.com/changchang/seq-queue/pull/3

wellwelwel:
I think this issue can be closed as unplanned, but I'll keep it open for @sidorares 🙋🏻‍♂️

andres645:
Thanks for the answer :)

sidorares:
I'll close it as "If it ain't broke, don't fix it" :)
In theory we could migrate to a similar library, or copy the code ( it's just one small file and license allows that ) or rewrite part of the library that needs it to handle control manually, but as @wellwelwel pointed out there is no license conflict with myql2 MIT and no known bugs related to seq-queue, at least in a context of our usage.

PR #4095, opened 2026-02-19
chore: replace seq-queue with simple queue class

jens-ox:
I'm currently building a license list for a project and came across `seq-queue`, which doesn't have a `license` field.

The package seems to be unmaintained and the author isn't responsive (see [issue](https://github.com/changchang/seq-queue/issues/5)).

This PR replaces the dependency with a simple inline queue implementation. While this might not be a very important change, it reduces the transitive complexity of both the code and the supply chain a bit.


jens-ox:
@wellwelwel - I've [added a very basic unit test](https://github.com/sidorares/node-mysql2/pull/4095/commits/019ca9b14e68a40e3f7a62870647d4d2e49c9364) because Codecov was previously complaining. This requires exporting the Queue from `compressed_protocol.js`, which isn't strictly necessary. If unit tests aren't needed here, I can revert `019ca9b`.

wellwelwel:
Thanks, @jens-ox! Regarding **seq-queue** license, it is actually in both the repository and **npm**, but not in **package.json**:

- https://github.com/changchang/seq-queue/blob/1f00953b3ac538af86603377c4aad39e98b2c176/LICENSE
- https://cdn.jsdelivr.net/npm/seq-queue@0.0.5/LICENSE


> @wellwelwel - I've [added a very basic unit test](https://github.com/sidorares/node-mysql2/pull/4095/commits/019ca9b14e68a40e3f7a62870647d4d2e49c9364) because Codecov was previously complaining. This requires exporting the Queue from `compressed_protocol.js`, which isn't strictly necessary. If unit tests aren't needed here, I can revert `019ca9b`.

About the tests, the `.cjs` files are temporarily in the repository so as not to cause conflict with open **PRs** in progress (#3937, specifically), but they are no longer executed.

The tests have been moved to **./test/esm** for now 🙋🏻‍♂️

wellwelwel:
I usually leave the decision on dependency changes to @sidorares, but in this case it's not a replacement, it's a removal for a really simpler local solution.

I believe that the only point of attention now is check for performance against **seq-queue**. Except for the benchmark analysis, **LGTM**!

> I just triggered the latest tests to run 🙋🏻‍♂️

jens-ox:
@wellwelwel I've move the tests to the esm subdirectory (I've put it in the root there, lmk if I should move it to a subdirectory there).

> Regarding seq-queue license, it is actually in both the repository and npm, but not in package.json

I know - I'm currently trying to figure out how license checking should look like. I'd like to generate a list of license texts for all the packages I'm using in a project, but this turned out to be a real nightmare - the license field isn't enforced and the actual license text is often not available or at a non-standard location (with the worst offenders being binary packages shipping only a huge compiled binary and a bare-bones `package.json` without repository or license info - see [example](https://www.npmjs.com/package/sst-win32-x86?activeTab=code), see [fix](https://github.com/anomalyco/sst/pull/6423)).
However, this endeavour turned out to uncover many old abandoned projects in the supply chain 😅

wellwelwel:
I believe the only difference between the `Class` is the timeout used by **seq-queue**. But in terms of performance, [**hyperfine**](https://github.com/sharkdp/hyperfine) performed a _little little_ better with the current minimalist implementation, than when applying this functionality:

> Simplified replica of the default **seq-queue** behavior (we can update it, if needed).

```js
class Queue {
  constructor() {
    this._pending = [];
    this._active = false;
    this._taskTimeout = 3000;
    this._timerId = null;
  }

  push(fn) {
    this._pending.push({
      fn,
      taskTimeout: this._taskTimeout,
    });

    if (!this._active) {
      this._active = true;
      process.nextTick(() => this._processNext());
    }
  }

  _processNext() {
    if (this._timerId) {
      clearTimeout(this._timerId);
      this._timerId = null;
    }

    const task = this._pending.shift();
    if (!task) {
      this._active = false;
      return;
    }

    let settled = false;
    this._timerId = setTimeout(() => {
      if (!settled) {
        settled = true;
        process.nextTick(() => this._processNext());
      }
    }, task.taskTimeout);

    task.fn({
      done: () => {
        if (!settled) {
          settled = true;
          process.nextTick(() => this._processNext());
        }
      },
    });
  }
}
```

---

And just a test to see how it works with massive processes _(not to be implemented or anything like that)_:

> 1.000.000 rows (10.000 por batch x 100 batches).

```ts
import type { RowDataPacket } from '../../../index.js';
import { assert, describe, it } from 'poku';
import { createConnection } from '../common.test.mjs';

await describe('Queue', async () => {
  const connection = createConnection({ compress: true }).promise();

  await it('should handle bulk queries with compression enabled', async () => {
    const BATCH_SIZE = 10_000;
    const TOTAL_BATCHES = 100;

    await connection.query(`
      CREATE TEMPORARY TABLE _test_queue_bulk (
        id INT AUTO_INCREMENT PRIMARY KEY,
        payload TEXT NOT NULL,
        blob_data BLOB NOT NULL
      )
    `);

    for (let batch = 0; batch < TOTAL_BATCHES; batch++) {
      const values: string[] = [];
      const params: (string | Buffer)[] = [];

      for (let i = 0; i < BATCH_SIZE; i++) {
        values.push('(?, ?)');
        params.push(
          JSON.stringify({
            batch,
            index: i,
            data: `row_${batch}_${i}_${'x'.repeat(200)}`,
          })
        );
        params.push(Buffer.alloc(256, (batch * BATCH_SIZE + i) & 0xff));
      }

      await connection.query(
        `INSERT INTO _test_queue_bulk (payload, blob_data) VALUES ${values.join(', ')}`,
        params
      );
    }

    const [rows] = await connection.query<RowDataPacket[]>(
      'SELECT COUNT(*) AS total FROM _test_queue_bulk'
    );
    assert.strictEqual(rows[0].total, BATCH_SIZE * TOTAL_BATCHES);

    const [allRows] = await connection.query<RowDataPacket[]>(
      'SELECT * FROM _test_queue_bulk ORDER BY id'
    );
    assert.strictEqual(allRows.length, BATCH_SIZE * TOTAL_BATCHES);

    const first = JSON.parse(allRows[0].payload);
    assert.strictEqual(first.batch, 0);
    assert.strictEqual(first.index, 0);

    const last = JSON.parse(allRows[allRows.length - 1].payload);
    assert.strictEqual(last.batch, TOTAL_BATCHES - 1);
    assert.strictEqual(last.index, BATCH_SIZE - 1);
  });

  await connection.end();
});
```

> Screenshot 2026-02-19 at 09 54 11

wellwelwel:
Thanks again, @jens-ox! 🤝

### Related

- #3137

Issue #1087, opened 2019-12-18
What happens when a connection is released?


joeyhub:
As far as I can understand a pool connection when released will at least make it available to any waiting connection. This means it's a persistent connection.

Does this perform any kind of cleanup? What happens if it is released with a transaction still in play?

joeyhub:
Is is possible to have a pool that instead keeps connections open in advance?

sidorares:
this is the code for `releaseConnection` - https://github.com/sidorares/node-mysql2/blob/2824df36a3fb7113ae44320871ae574ca56f61a7/lib/pool.js#L79-L94

In short, it's just returned to list of "free" connection and is up for grabs for next `.getConnection` call. You are responsible for cleaning up connection state.

joeyhub:
If I destroy the connection instead such as using end does the pool know to create a new connection?

For transactional workloads first and foremost I usually have the connection die on any error at all with the default for a lost connection always being rollback. I tend to assume the worst rather than risk committing a partially complete transaction.

This also applies to connection persistence in those cases. Unless a clean state can be guaranteed I wouldn't use it.

MySQL transactions offer various guarantees and I feel like the documentation where possible should be very clear about whether or not it maintains those.

The simplest way to be able to maintain them is if rather than release the connection can be ended and the pool will make a new one. I have a feeling there may be a command or will be to reset the current connection but I'm not sure.

I would not recommend implementing an attempt to clean connections by hand because it gets very complicated very quickly with transaction state, variable state, temp table state, etc.

I believe traditionally this is meant for reusing connections safely in a pool:

https://dev.mysql.com/doc/refman/8.0/en/mysql-reset-connection.html

joeyhub:
https://github.com/mysql/mysql-server/blob/91a17cedb1ee880fe7915fb14cfd74c04e8d6588/libmysql/libmysql.cc

joeyhub:
In the mean time a work around is crucial or use of pool is a no-no for guarantee based dev unless DIY.

I think ensuring and documenting end is fine to use instead of release for safety is crucial.

Aside from that full transaction support is really complex. I've never seen an application library do it properly.

[...]

Issue #1469, opened 2021-12-06
poolConnection.changeUser changing for the whole Pool.getConnection()


TheVoid-0:
This may be intentional, but if it is, i'd need some help.
Basically what happens is that when you get a pool connection and then change it's user (i use this because i need to change the database) and a error occurs, like bad credentials for the new user, instead of only compromising the current connection that i've got, it compromises the whole pool, whenever i try to get another connection from the pool, the pool tries to use the same wrong credentials that i've passed to the connection.changeUser. This makes the pool unusable in my code because i can never get a connection again to be able to change user, in my understanding, changing the user on the connection should only affect that pooledConnection and not the whole pool.

I have one pool per server, and that server has many dbs, that's why i need to change the connection user from that specific pool, when some error occurs that makes that connection unusable, i use connection.destroy(), but this don't prevent the pool to keep trying to connect the new pool.getConnection() with the same wrong credentials i passed to the one i destroyed.

I would like to get some info if this behaviour is intentional, or if it really has this behaviour at all, maybe i'm fuc**ing things up somewhere... Thanks for helping in advance!

TheVoid-0:
Here is a [repo](https://github.com/TheVoid-0/mysql2-changeuser-pool-issue) to try reproduce the issue . I've already fix my problem by releasing the connection instead of destroying it. But i think this shouldn't matter because i may want to actually destroy a connection that got access denied instead of releasing it.

diovannaschell:
I'm also having this problem and would like to know if this is the expected behavior. In my current test, when you get two or more connections from the same pool and change the user using the connection.changeUser() in any of them, it changes for every connection on that pool, even the ones already in use, that IMO does not make much sense, if I wanted to change for every connection I would look for a method to do this on the pool itself, the connection should only affect itself. Is this really intended? Like @TheVoid-0  I also want to use multiple databases on the same pool and is currently unviable

sidorares:
@diovannaschell @TheVoid-0 hopefully https://github.com/sidorares/node-mysql2/pull/1437#issuecomment-954475737 should help

sidorares:
I maybe I'm reading this wrong @TheVoid-0 . So you thing `changeUser` is updating the pool config and next time connection is created it's using that config instead of the original one?

```js
const pool = createPool({ user: 'A', db: 'B' });
const connA = await pool.getConnection();
await connA.changeUser({ user: 'B', db: 'C' });
[...]

PR #4148, opened 2026-03-03
feat: implement COM_RESET_CONNECTION with pool integration

sidorares:
This PR implements MySQL's `COM_RESET_CONNECTION` command (introduced in MySQL 5.7.3) and integrates it with connection pools via a new `resetOnRelease` pool option.

This addresses **7-year-old issue #934** and resolves issues #1087, #1328, and #1469.

### Problem

When connections are reused from a pool, session state persists between different users:

- User variables remain set
- Temporary tables persist
- Session variables stay modified
- Locks (GET_LOCK) aren't released
- Active transactions continue

This creates:

- **Security issues**: Session state leakage between users
- **Data integrity issues**: Unexpected state in "fresh" connections
- **Hard-to-debug bugs**: State pollution across requests

### Current Workarounds (Suboptimal)

1. **Call `changeUser()` with no params**: Requires full re-authentication (slow, extra roundtrip)
2. **Destroy and recreate connections**: Wasteful, defeats pool purpose
3. **Manually clean state**: Error-prone, easy to miss something
4. **Avoid pools for safety**: Defeats the purpose of pooling

## Solution

Implements `COM_RESET_CONNECTION` (0x1F) - a lightweight command that:

- Resets all session state (variables, temp tables, locks, transactions, etc.)
- Does NOT re-authenticate (faster than changeUser)
- Does NOT close TCP connection
- **~3-5x faster** than changeUser for state reset

### 1. Core Command Implementation

- Add `ResetConnection` command class
- Add `ResetConnection` packet class
- Add `connection.reset(callback)` method
- Add promise wrapper `await connection.reset()`
- Clear prepared statements cache on reset
- TypeScript definitions

### 2. Pool Integration

- Add `resetOnRelease` pool config option (**default: `false`**)
- Modify `releaseConnection()` to call `reset()` when enabled
- Handle reset failures gracefully (destroy connection, create new one)
- Add `_handleSuccessfulRelease()` helper method
- TypeScript pool option definition

### 3. Comprehensive Tests

- **6 connection reset tests** (`test/integration/connection/test-reset-connection.test.mts`):
  - Basic reset functionality
  - User variable cleanup
  - Temporary table cleanup
  - Prepared statements cache clearing
  - Transaction rollback
  - Promise wrapper support

- **7 pool integration tests** (`test/integration/pool/test-pool-reset-on-release.test.mts`):
  - Explicit opt-in behavior (resetOnRelease: true)
  - Opt-out behavior (resetOnRelease: false)
  - Error handling for failed resets
  - Prepared statements cleanup
  - Promise pool support
  - Queued connection reset behavior

### Connection API (added)

```js
// Callback style
connection.reset((err) => {
  if (err) throw err;
  console.log('Connection state reset');
});

// Promise style
await connection.reset();
```

### Pool API (enhanced)

```js
// Default: no reset on release (preserves existing behavior)
  host: 'localhost',
  user: 'root',
  password: 'password',
  database: 'test',
  // resetOnRelease defaults to false
});

// Opt-in: automatic reset on release (RECOMMENDED for security)
const securePool = mysql.createPool({
  // ... config
  resetOnRelease: true,
});
```

## Why `resetOnRelease` defaults to `false`

While `resetOnRelease: true` is the more correct and secure behavior for connection pools, we default to `false` for this initial release to avoid a **semver-breaking change**. Existing applications may rely on session state (user variables, temporary tables, etc.) persisting across pool connection reuse. Changing this default silently would break those applications.

**Future plan**: In the next major version, `resetOnRelease` will default to `true`, as this is the safer and recommended behavior for production applications. We encourage all users to explicitly opt in to `resetOnRelease: true` now.

## What Gets Reset

When `connection.reset()` is called (or triggered by pool with `resetOnRelease: true`):

| State | Cleared? | Details |
|---|---|---|
| User variables | Yes | All `@variable` values cleared |
| Temporary tables | Yes | All temp tables dropped |
| Prepared statements | Yes | Server invalidates, client clears cache |
| Session variables | Yes | Reset to global defaults |
| Locks (GET_LOCK) | Yes | All named locks released |
| Active transaction | Yes | Rolled back if active |
| User/Database | No | Unchanged (use changeUser for this) |

### Benchmarks (preliminary)

| Scenario | Time | vs. changeUser |
|---|---|---|
| changeUser() | ~5-10ms | baseline |
| reset() | ~1-2ms | **3-5x faster** |
| No reset | ~0ms | n/a |

## MySQL Version Compatibility

| Version | Support |
|---|---|
| MySQL 5.7.3+ | Full support |
| MySQL 5.7.0 - 5.7.2 | Command not available |
| MySQL 5.6 and older | Command not available |
| MariaDB 10.2.4+ | Full support |

**Error handling**: If reset fails (e.g., old MySQL version), the connection is destroyed and removed from pool, and a new connection is created for the next request.

## Issues Resolved

This PR directly addresses:

- **#1328** (May 2021): Feature request for mysql_reset_connection
- **#934** (March 2019): Connection state is not reset on return to pool (7 years old!)
- **#1087** (Oct 2023): What happens when a connection is released?
- **#1469** (Dec 2021): poolConnection.changeUser changing for the whole Pool

Also helps with:

- **#1103** (June 2023): Recommendation on connection recycling
- **#586** (Jan 2021): Event on release connections?

## Related PRs

- **#1437** (Oct 2021): Implement CMD_RESET_CONNECTION (draft by @testn) - This PR incorporates and completes that work

## Backward Compatibility

**Fully backward compatible** - no breaking changes.

- `resetOnRelease` defaults to `false`, preserving existing pool behavior
- Direct `connection.reset()` calls are new API, don't affect existing code
- Users can opt in to `resetOnRelease: true` at their discretion

## Checklist

- [x] Implementation complete
- [x] Tests added and passing
- [x] TypeScript definitions added
- [x] Linting passes
- [x] No breaking changes to existing code paths
- [ ] Documentation updated
- [ ] Changelog entry added
- [ ] Approved by maintainers

## Credits

- Original PR #1437 by @testn
- Issue reports and discussion by @aikar, @joeyhub, @danielgindi, @TheVoid-0
- MySQL protocol documentation

### DIFF
--- a/website/docs/documentation/00-index.mdx
+++ b/website/docs/documentation/00-index.mdx
@@ -21,6 +21,7 @@ Not only **MySQL2** offers better performance over [Node MySQL][node-mysql], we 
 
 - [Prepared Statements](/docs/documentation/prepared-statements)
 - [Query Attributes](/docs/documentation/query-attributes)
+- [Reset Connection](/docs/documentation/reset-connection)
 - [Promise Wrapper](/docs/documentation/promise-wrapper)
 - [Authentication Switch](/docs/documentation/authentication-switch)
 - [More Features](/docs/documentation/extras)
--- a/website/docs/documentation/prepared-statements.mdx
+++ b/website/docs/documentation/prepared-statements.mdx
@@ -41,7 +41,7 @@ connection.prepare('select ? + ? as tests', (err, statement) => {
 });
 ```
 
-Note that you should not use statement after connection reset (`changeUser()` or disconnect). Statement scope is connection, you need to prepare statement for each new connection in order to use it.
+Note that you should not use statement after connection reset (`changeUser()`, [`reset()`](/docs/documentation/reset-connection) or disconnect). Statement scope is connection, you need to prepare statement for each new connection in order to use it.
 
 ## Configuration
 
--- a/website/docs/documentation/reset-connection.mdx
+++ b/website/docs/documentation/reset-connection.mdx
@@ -1,0 +1,194 @@
+---
+tags: [reset, resetOnRelease, pool, COM_RESET_CONNECTION]
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Reset Connection
+
+MySQL's `COM_RESET_CONNECTION` command (available since MySQL 5.7.3 and MariaDB 10.2.4) resets a connection's session state without closing the underlying TCP connection or re-authenticating. It is significantly faster than `changeUser()` for clearing session state.
+
+## What gets reset
+
+| State               | Cleared? | Details                                 |
+| ------------------- | -------- | --------------------------------------- |
+| User variables      | Yes      | All `@variable` values cleared          |
+| Temporary tables    | Yes      | All temp tables dropped                 |
+| Prepared statements | Yes      | Server invalidates, client clears cache |
+| Session variables   | Yes      | Reset to global defaults                |
+| Locks (GET_LOCK)    | Yes      | All named locks released                |
+| Active transaction  | Yes      | Rolled back if active                   |
+| User / Database     | No       | Unchanged (use `changeUser` for this)   |
+
+<hr />
+
+## connection.reset()
+
+> **reset(callback?: (err: Error | null) => void): void**
+
+Resets the connection session state. The connection remains open and authenticated.
+
+<Tabs>
+  <TabItem value='promise.js' default>
+
+```js
+import mysql from 'mysql2/promise';
+
+const connection = await mysql.createConnection({
+  host: 'localhost',
+  user: 'root',
+  database: 'test',
+});
+
+await connection.query('SET @user_id = 123');
+
+// highlight-next-line
+await connection.reset();
+
+// @user_id is now NULL
+const [rows] = await connection.query('SELECT @user_id as val');
+console.log(rows[0].val); // null
+
+await connection.end();
+```
+
+  </TabItem>
+  <TabItem value='callback.js'>
+
+```js
+const mysql = require('mysql2');
+
+const connection = mysql.createConnection({
+  host: 'localhost',
+  user: 'root',
+  database: 'test',
+});
+
+connection.query('SET @user_id = 123', (err) => {
+  if (err) throw err;
+
+  // highlight-next-line
+  connection.reset((err) => {
+    if (err) throw err;
+
+    // @user_id is now NULL
+    connection.query('SELECT @user_id as val', (err, rows) => {
+      if (err) throw err;
+      console.log(rows[0].val); // null
+      connection.end();
+    });
+  });
+});
+```
+
+  </TabItem>
+</Tabs>
+
+:::tip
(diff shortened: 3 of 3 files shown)


wellwelwel:
Hey @sidorares, for the tests that never finish, please see the examples in PR #4094's description 🙋🏻‍♂️

sidorares:
## Addressed Review Feedback ✅

Thanks @wellwelwel for the excellent suggestions! I've applied all of them in commit `6572f059`.

### Changes Made:

**1. Replaced complex type assertions with `@ts-expect-error`** ✅
- Much cleaner and more maintainable
- Easier to track for future typing improvements
- Changed from: `(conn as unknown as { _statements?: { size: number } })._statements`
- Changed to: `// @ts-expect-error: internal access\n conn._statements`

**2. Added try/finally blocks for connection cleanup** ✅
- All connection tests now use try/finally pattern
- Prevents hanging tests when assertions fail
- Connection.end() guaranteed to be called even on errors

**3. Buffer.alloc() instead of Buffer.allocUnsafe()** (Copilot suggestion) ✅
- Prevents potential memory leak of uninitialized buffer contents
- Negligible performance cost for 5-byte packet

**4. Pass null explicitly to callback** (Copilot suggestion) ✅
- Changed: `onResult.bind(this)` → `onResult.bind(this, null)`
- Aligns with Node.js callback convention: `callback(err, result)`
- Callback gets `null` on success, not `undefined`

### Regarding Test Hangs:
Following your guidance and PR #4094 examples, I've refactored all tests to use the outer-scope pattern for connection lifecycle management. This ensures connections are properly cleaned up even when tests fail.

All review feedback has been addressed! 🎉

wellwelwel:
Bringing #4154, #4170, #4185, and #4188 here.

By forcing a timeout directly in the tests, each test that exceeds the limit now fails and displays the errors instead of hanging indefinitely.

This will help to see reports that were not previously displayed since the processes were never closed (the tests continue to wait for an error (`assert`) at the same level as connections end or try to encapsulate the assertions in a `try` block instead of defining the connections and they end in separate scopes) 🙋🏻‍♂️

Real case: https://github.com/sidorares/node-mysql2/actions/runs/22732532589/job/65925146713?pr=4148

sidorares:
@wellwelwel I'm thinking to make 'resetOnRelease' default off, at least for this change
while this is probably a correct behavior it would require major version release as it could potentially break existing behaviour

wellwelwel:
> I'm thinking to make 'resetOnRelease' default off, at least for this change while this is probably a correct behavior it would require major version release as it could potentially break existing behaviour

Just yesterday I fixed an unexpected breaking change conflict from 2024 caused by a change in the instance source of a private class from `PoolConnection`:

- #4183

So I agree with the idea of not breaking the behavior without a major semver 🙋🏻‍♂️

sidorares:
> @sidorares, do you think it's worth including this option/flag in the documentation (a small paragraph in [Extra Features](https://sidorares.github.io/node-mysql2/docs/documentation/extras), for example)? Otherwise, **LGTM** 🙋🏻‍♂️

definitely, I thought I added it, will do now

PR #4150, opened 2026-03-03
ci: ensure timeout to OSX workflow

wellwelwel:
Just as the title says 🙋🏻‍♂️


wellwelwel:
All **GitHub**-hosted runners with label `macos-latest` are busy:

> Screenshot 2026-03-03 at 12 06 56

PR #4159, opened 2026-03-08
fix: validate buffer bounds in geometry parser

wellwelwel:
`parseGeometryValue` now checks remaining buffer length before each read.

Malformed geometry payloads with inflated point/ring counts previously caused uncaught `RangeError` crashes and now return `null`.

- Just a **TOC**: I resolved all `// eslint-disable-next-line no-case-declarations` comments in `parseGeometryValue` 🙋🏻‍♂️

> cc @peaktwilight, please see also https://github.com/sidorares/node-mysql2/commit/7c2ae002a5e1708d0a0a757bedd00c7e16ae2eee 🤝


peaktwilight:
@wellwelwel awesome thanks for fixing it! If you need help with fixing the rest the findings I mentioned by email, I'll gladly help out :D

PR #4161, opened 2026-03-08
fix: bound `null`-terminated string read to packet end

wellwelwel:
Previously, the `null` byte scan could read beyond the current packet into adjacent data in the same backing Buffer. `readNullTerminatedString` now stops at the logical packet boundary.

**Credits:** The fix used was totally made by @peaktwilight:

```js
while (end < this.end && this.buffer[end] !== 0x00) {
  end = end + 1;
}
```

- My work here was basically confirming the bug and creating tests to reproduce the behaviors and cover it.

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -406,8 +406,8 @@ class Packet {
   readNullTerminatedString(encoding) {
     const start = this.offset;
     let end = this.offset;
-    while (this.buffer[end]) {
-      end = end + 1; // TODO: handle OOB check
+    while (end < this.end && this.buffer[end] !== 0x00) {
+      end = end + 1;
     }
     this.offset = end + 1;
     return StringParser.decode(this.buffer, encoding, start, end);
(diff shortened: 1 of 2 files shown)


peaktwilight:
Thanks for your work!

wellwelwel:
> Thanks for your work!

Thanks too, @peaktwilight!

"Finding 1" has some more complex concepts to consider when fixing, such as multiple ways to approach corrections, including avoiding breaking changes. I'll think about this a little more calmly before submitting a _PR_ 🙋🏻‍♂️

> [!NOTE]
> Both this and #4159 fixes are already available in the `canary` version, and you can see the credits in #4160 🤝

peaktwilight:
Awesome thanks so much and lmk if I can help with that one too. Also is it worth doing an advisory for this or not? Tbh not sure :D

wellwelwel:
> Awesome thanks so much and lmk if I can help with that one too. Also is it worth doing an advisory for this or not? Tbh not sure :D

I do believe it is valid _(all tests prove the points)_. Just if possible, wait for #4160 to be merged, submitting the complete report already including the fixes 🤝

peaktwilight:
Yup will do for sure, I'll also wait on the release to be out to be safe before publishing any kind of writeup. Gn!

PR #4162, opened 2026-03-08
fix: prevent query param override of URL-defined connection options

wellwelwel:
Query parameters in `parseUrl` could override connection options already extracted from the URL structure (e.g., `host`, `port`, `user`, `password`, and `database`):

> https://github.com/sidorares/node-mysql2/blob/91c5229dff2293953635b93f753b45bff31deac4/lib/connection_config.js#L273-L279

Now, keys already present in the parsed options object are skipped during query parameter processing.

Dynamic options like `ssl`, `multipleStatements`, etc. remain configurable via query parameters since they are not part of the URL structure and can/are commonly used in **DSN** strings, it also prevent a potential breaking change to this fix, since whoever controls the **DSN**, usually already controls the host and credentials.

> cc @peaktwilight, please see also https://github.com/sidorares/node-mysql2/commit/3123b4e686e4e7c3893b20773376aff2c31840f7 🤝

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -277,7 +277,10 @@ class ConnectionConfig {
       user: decodeURIComponent(parsedUrl.username),
       password: decodeURIComponent(parsedUrl.password),
     };
-    parsedUrl.searchParams.forEach((value, key) => {
+    for (const [key, value] of parsedUrl.searchParams) {
+      if (key in options) {
+        continue;
+      }
       try {
         // Try to parse this as a JSON expression first
         options[key] = JSON.parse(value);
@@ -285,7 +288,7 @@ class ConnectionConfig {
         // Otherwise assume it is a plain string
         options[key] = value;
       }
-    });
+    }
     return options;
   }
 }
--- a/test/unit/connection/test-parseurl-option-override.test.mts
+++ b/test/unit/connection/test-parseurl-option-override.test.mts
@@ -1,0 +1,53 @@
+import { describe, it, strict } from 'poku';
+import ConnectionConfig from '../../../lib/connection_config.js';
+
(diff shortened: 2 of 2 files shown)


wellwelwel:
This _PR_ also indirectly fixes the possibility of prototype pollution (manipulating the `__proto__` key). @peaktwilight, I believe it would be interesting to complement the report/advisory with this detail.

peaktwilight:
@wellwelwel perfect, will add it to the advisory and writeup when published, thanks for the heads up!

In the meantime, there'a slight edge case that might also be worth looking at while we're already at it:

The bounds checks before each read prevent the RangeError crash, but the loop count itself (num/numPoints/numRings) is still taken at face value from the buffer. For the multi-geometry types (cases 4–7), a payload with numGeometries = 0xFFFFFFFF and no actual geometry data causes a bunch of no-op iterations of parseGeometry() > each call hits the bounds check and returns null, but the loop still spins.

I tested this against the canary build:

  Count                     Time (ms)
  1,000                     0.18ms
  100,000                   1.88ms
  1,000,000                 10ms
  10,000,000                91ms
  100,000,000               734ms
  0xFFFFFFFF (extrapolated)  ~37s

A single malformed geometry response could freeze the event loop for more than half a minute.

Same approach would work for numPoints in WKBLineString and numRings in WKBPolygon. Something like:

// after reading num/numPoints/numRings:
if (num > (bufferLength - offset) / 9) return null;

9 bytes = minimum sub-geometry size (1 byte order + 4 wkbType + 4 count). For points it'd be /16 (two doubles), and for rings /4 (minimum ring header).

Obviously minor compared to the other fixes, but same preconditions as the original finding (attacker-controlled server data). Just figured I'd flag it while the code is fresh :D

PR #4164, opened 2026-03-09
fix: handle malformed geometry payloads

wellwelwel:
Now `parseGeometryValue` validates counts against remaining buffer bytes before entering loops from malformed geometry payloads.

Related:

- https://github.com/sidorares/node-mysql2/pull/4162#issuecomment-4021495280

> cc @peaktwilight


peaktwilight:
Awesome 👍 !

Issue #3726, opened 2025-07-23
Drop prepared statements for cached queries if connection gets swapped due to operations like ZDP


JasonShin:
Cloud database providers like AWS Aurora offer Zero Downtime Patching (ZDP), which maintains connection availability during maintenance windows. However, ZDP has a critical limitation: prepared statements don't survive the underlying connection replacement that occurs during the patch process.
This creates a mismatch between the client-side cache (node-mysql2's LRU cache of prepared statements) and the server-side reality. The client believes cached prepared statements are still valid, but the database has discarded them after the ZDP operation. When the application attempts to execute a cached prepared statement, it receives a "prepared statement not found" error.

**Proposed Solution**
Implement resilient prepared statement handling that can recover from ZDP scenarios:

- Error Detection: Catch prepared statement errors that indicate the statement no longer exists on the server
- Cache Invalidation: Remove the invalid entry from the LRU cache
- Transparent Recovery: Automatically re-prepare the statement and retry the original query
- Seamless Experience: Handle this recovery transparently so the application layer doesn't need to manage ZDP-related failures

This approach would make the prepared statement cache self-healing, automatically adapting when cloud infrastructure operations invalidate server-side prepared statements while maintaining client-side cache consistency.

The key insight is treating prepared statement "not found" errors as cache invalidation signals rather than terminal failures, enabling graceful recovery from infrastructure-level operations like ZDP.

PR #4199, opened 2026-03-20
feat: auto purge stale prepared statements from cache and retry query

Invader444:
Even with proper tuning of maxPreparedStatements, it is possible that the automatic LRU cache contains stale prepared statements that have been released on the server from time to time.

Normally, with cached prepared statements, a user could catch the ER_UNKNOWN_STMT_HANDLER error, purge from their cache, and try again with a newly prepared statement. However, to do that with the internal LRU cache, a user would need to touch internals that are not meant to be public APIs... which could break at any time.

This handles that invalidation and retry logic internally so the user doesn't have to touch private/internal implementation details.

This may help issue #805


Invader444:
Added additional coverage for the EventEmitter path.

Invader444:
Ok, this time I think it's ready; apologies for all the noise 😅

Invader444:
Should be at 100% coverage now, and also cleaned up the "emit 'end' twice" that was happening when the internal prepare statement failed.

AlekseyLkh:
I think this PR does not cover all possible cases where a prepared statement gets unexpectedly invalidated.

There are essentially two scenarios:

- Unexpected caching when using `prepare()` manually, and consequently unexpected invalidation in the case of eviction (I ran into this while studying the library's behavior manually).
- Invalidation during ZDP (Zero Downtime Patching), for example on AWS RDS. (I encountered this problem recently when AWS was patching RDS servers and it is the reason why I found this PR. Below is what I came up with after Investigate.)

The first case, in my view, should be fixed by removing the cache from the manual mechanism. If someone is manually using prepare(), stmt.execute(), stmt.close(), it should not interact with the cache at all. The caching mechanism should remain exclusively on the execute method side.

Regarding the second case: During ZDP, the session is essentially destroyed and `ER_UNKNOWN_STMT_HANDLER` errors start flooding in. It would be logical to remove statements from the cache one by one in this situation, but that approach is problematic due to MySQL's internal behavior. The server generates prepared statement IDs starting from zero, and during ZDP the counter is apparently reset — so newly created statements may receive the same ID as one already present in the cache. This leads to ER_WRONG_ARGUMENTS errors rather than ER_UNKNOWN_STMT_HANDLER errors. Therefore, I believe it is more practical to completely clear the cache whenever the caching mechanism encounters an ER_UNKNOWN_STMT_HANDLER error.

sidorares:
> The first case, in my view, should be fixed by removing the cache from the manual mechanism. If someone is manually using prepare(), stmt.execute(), stmt.close(), it should not interact with the cache at all. The caching mechanism should remain exclusively on the execute method side.
> 

agree.

> Regarding the second case: During ZDP, the session is essentially destroyed and `ER_UNKNOWN_STMT_HANDLER` errors start flooding in. It would be logical to remove statements from the cache one by one in this situation, but that approach is problematic due to MySQL's internal behavior. The server generates prepared statement IDs starting from zero, and during ZDP the counter is apparently reset — so newly created statements may receive the same ID as one already present in the cache. This leads to ER_WRONG_ARGUMENTS errors rather than ER_UNKNOWN_STMT_HANDLER errors. Therefore, I believe it is more practical to completely clear the cache whenever the caching mechanism encounters an ER_UNKNOWN_STMT_HANDLER error.

interesting. I wonder if there is any way to get notified about statement counter reset / ZDP. If that's not possible, clearing cache on ER_UNKNOWN_STMT_HANDLER seems like a logical thing to do ( still, potentially unexpected behaviour when the query that is actiually executed might not match to what was in `execute()` callsite

Invader444:
https://github.com/sidorares/node-mysql2/issues/3726 mentions this...

lib/base/connection.js
```diff
@@ -837,7 +876,48 @@ class BaseConnection extends EventEmitter {
...
+          firstArg &&
+          firstArg.errno === ER_UNKNOWN_STMT_HANDLER
+        ) {
+          executeCommand.emit = (eventName, ...rest) => {
+            if (eventName === 'end') {
+              // Intercept the 'end' event that will be emitted after this 'error' event is emitted
+              executeCommand.emit = origEmit;
+              return false;
+            }
+            /* c8 ignore next 3 */
```
wellwelwel:
```suggestion
```

I recommend not ignoring coverage alerts so we can always keep track of what the tests actually cover (and what they don't) 🙋🏻‍♂️

lib/base/connection.js
```diff
@@ -814,13 +816,50 @@ class BaseConnection extends EventEmitter {
...
+            executeCommand.onResult = (err, ...rest) => {
+              if (err && err.errno === ER_UNKNOWN_STMT_HANDLER) {
+                const origEmit = executeCommand.emit.bind(executeCommand);
+                executeCommand.emit = (eventName, ...rest) => {
+                  if (eventName === 'end') {
+                    // Intercept the 'end' event that will be emitted after this 'error' event is emitted
+                    executeCommand.emit = origEmit;
+                    return false;
+                  }
+                  /* c8 ignore next 3 */
```
wellwelwel:
```suggestion
```

wellwelwel (changes requested):

Invader444:
I think for the ZDP scenario where statement IDs might point to the wrong statement... a Map of statement ID to statement key would allow us, during prepare response, to detect that this has happened and remove the old statement that had the same ID... thoughts?

This would need to include all prepared statements, even if they weren't automatically created via execute... though we should also stop caching if not coming from execute in the first place.

sidorares:
I was thinking if changing prepare() / unprepare() caching behavior needs a major version release, but reading https://github.com/sidorares/node-mysql2/issues/805#issuecomment-447871245 I'm more inclined to release it as a bugfix as its currently already quite broken and likely almost never used. I think we can merge this change and have a follow up PR which removes caching from manual prepare()

Invader444:
Awesome, but the ZDP thing for me (and I think that's what bit me too which prompted me to write this), highlights a bigger problem with mismatched query IDs to queries... admittedly unlikely on a given connection to SELECT * FROM widgets WHERE id = ?, and the server thinks that prepared statement is DELETE FROM widgets WHERE id = ?, but possible?

sidorares:
I initially thought it's possible, but maybe not. After ZDP happens, the very next `.execute()` command results in either ER_UNKNOWN_STMT_HANDLER if it was previously cached or in id conflict which we can also detect. I haven't read you changes carefully yet @Invader444 but I assumed your PR handles both scenarios? 

ER_UNKNOWN_STMT_HANDLER -> reset cache, retry prepare + execute again
server returns id that we already have -> reset cache, save new statement, continue

Invader444:
I thought it would be execute -> prepare/cache store -> execute it, execute again -> cache fetch succeed -> ZDP -> another out of band prepare makes the same query id for a delete with same number and type of arguments -> second execute performs the wrong query.

Invader444:
But if you are using the pool correctly that's not possible... so maybe not a real issue 🤔

Invader444:
Unclear on how id conflict would work, I don't think the id is part of the cache key so how would we know without a full cache scan (which sounds like a bad thing to do...)

Invader444:
That leads to my map of query id to cache key.
prepare will always store query id to cache key map. Releasing a prepared statement would remove from that map.
Cache lookup would fetch the key from the cache, then check if the id to key map contains the query id. If it doesn't, or it does and points to the same key, use the cache. Otherwise, delete from cache and execute a new prepare.

sidorares:
@Invader444 yes, unless we add statement id -> cache key map and reset connection on the first conflict the following is possible now:

```
-> client calls execute('select 1 where ? = ?', [1, 1])
-> client does not have it cached yet
-> sends COM_STMT_PREPARE "select 1 where ? = ?"
<- server sends statement id 0
-> client caches: "select 1 where ? = ?" -> statement id 0
-> sends COM_STMT_EXECUTE id 0 with params [1, 1]
<- server returns result for "select 1 where ? = ?"
--- server ZDP / session reset happens ---
   server-side prepared statements are gone
   client cache still says:
   "select 1 where ? = ?" -> statement id 0
--- some new prepares happen on that same logical connection before client retries this query ---
-> another prepare happens for "delete from t where a = ? and b = ?"
<- server sends statement id 0
   because statement ids restarted from 0 after reset
-> client later calls execute('select 1 where ? = ?', [1, 1])
-> client looks in cache and finds statement id 0
-> sends COM_STMT_EXECUTE id 0 with params [1, 1]
<- server executes the statement currently bound to id 0 which is now "delete from t where a = ? and b = ?"
```

sidorares:
Reviewed the retry mechanics against MySQL 8.3 (branch as of the merge with master) — the new tests pass, but I found two confirmed defects, verified with repros.

### Confirmed defects

**1. The retry is appended to the tail of the command queue, so commands queued behind the failing execute run before the retried one.** Inside the error handler, `prepareAndExecute` goes through `addCommand`, which pushes behind everything already queued. Repro: queue a stale `execute` followed by a `query`, and the query completes first:

```
Completion order:
  query(after)   -> [{"second":2}]
  execute(retry) -> [{"test":101}]
```

A connection normally guarantees serialized, in-order execution, and code relies on that — think `execute('UPDATE …')` (stale) followed by `query('COMMIT')`: the UPDATE would silently run *after* the COMMIT. The retry's prepare+execute pair needs to jump to the front of `_commands` (an unshift-style insert) rather than the back.

**2. Retry colliding with `end()` double-invokes the user callback and masks the real error.** With `conn.execute(staleStmt, cb); conn.end()`, `addCommand` has already been swapped to `_addCommandClosedState` when the retry fires, so both re-added commands synchronously error. On master the callback fires **once** with the real `ER_UNKNOWN_STMT_HANDLER`; on this branch it fires **twice** with `"Can't add new command when connection is in closed state"`:

```
master:      callback #1: err=Unknown prepared statement handler (1) given to mysql_stmt_precheck
this branch: callback #1: err=Can't add new command when connection is in closed state
             callback #2: err=Can't add new command when connection is in closed state
```

The double invocation happens because the wrapper restores `executeCommand.onResult = wrappedCb` before re-adding, so the prepare's `errorCb` fires it once and `_addCommandClosedState` fires it again for the execute command. In EventEmitter mode the same collision is worse: `executeCommand` never emits `'end'`, so anyone awaiting it hangs. If the connection can no longer accept commands, the retry should be abandoned and the original error delivered once.

### Design and smaller points

- **`cacheHasKey` is snapshotted when `execute()` is called, but the actual cache hit happens later when the queued `Prepare` runs.** If the entry is evicted in between (LRU at capacity), the run-time prepare is real and fresh, yet the interceptor is still armed — and on a genuine mid-flight reset it would then patch the *fresh* statement's `close` to a no-op and drop it from the cache, leaking a server-side handle. The `Prepare` command is the one that knows whether it cache-hit; deciding there would remove the whole class of divergence.
- **The emit-swapping logic is duplicated nearly verbatim in the callback and EventEmitter branches**, and it needs several comments to explain itself ("leaving this here in case that changes in the future…"). Extracting one helper (or handling the retry at the `Command` level instead of monkey-patching `emit` on a live emitter) would cut the duplication and most of the narration.
- On retry, `_setTimeout` runs again, so `options.timeout` effectively doubles for a retried statement. Probably acceptable, worth being deliberate about.
- Tests: the `addCommand` interception keyed on "the third call" is brittle — if command scheduling ever changes, the test corrupts the wrong command's query and fails confusingly. Asserting the intercepted command is actually a `Prepare` (e.g. check `command.query`/constructor) before mutating would make it self-diagnosing. Also `strict(x.size === 1)` gives a useless failure message compared to `strict.equal(x.size, 1)`.
- **No docs update.** Automatic retry is user-observable behavior (errors users previously caught and handled themselves now self-heal) — a short note on the prepared-statements docs page under `website/docs/` would be worth adding.

PR #4257, opened 2026-04-10
perf: defer Error object creation to error handlers in promise wrappers

sidorares:
## Summary

- Moves `new Error()` instantiation from function scope into error handling blocks across all promise wrapper modules, so Error objects are only created when actual errors occur
- Eliminates unnecessary CPU and GC overhead from stack trace capture on every successful call in high-throughput scenarios
- Updates `test-async-stack` to verify error propagation without asserting on caller-frame line numbers (which are no longer captured at the call site)

### Trade-off

The previous pattern captured a stack trace at the call site on *every* invocation so that, on error, the rejected Error would include the caller's frame. This is useful for debugging but expensive: in benchmarks over 280k calls the eager `new Error()` was a measurable CPU hot-spot. After this change, stack traces on errors still exist but originate from the internal callback rather than the caller.

### Changes

| File | Change |
|---|---|
| `lib/promise/make_done_cb.js` | Remove `localErr` parameter; create `Error` inside `if (err)` |
| `lib/promise/connection.js` | Remove eager `new Error()` from all 10 methods |
| `lib/promise/pool.js` | Same for `query`, `execute`, `end` |
| `lib/promise/pool_cluster.js` | Same for `query`, `execute` |
| `lib/promise/prepared_statement_info.js` | Same for `execute` |
| `promise.js` | Same for `createConnectionPromise`, `PromisePoolCluster` methods |
| `test/…/test-async-stack.test.mts` | Assert error propagation (code, message, stack presence) instead of caller-frame line offsets |

Continues #3169 — rebased onto the current modular codebase.

Co-authored-by: gunman <gunman@neowiz.com>

## Test plan

- [x] `npm run lint` — passes
- [x] `act` CI Build (typecheck + circular imports) — passes
- [x] `act` CI Linux (Node 22, mysql:8.3, SSL=0, Compression=0) — 205/205 tests pass

### DIFF
--- a/lib/promise/connection.js
+++ b/lib/promise/connection.js
@@ -26,14 +26,13 @@ class PromiseConnection extends EventEmitter {
 
   query(query, params) {
     const c = this.connection;
-    const localErr = new Error();
     if (typeof params === 'function') {
       throw new Error(
         'Callback function is not available with promise clients.'
       );
     }
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       if (params !== undefined) {
         c.query(query, params, done);
       } else {
@@ -44,14 +43,13 @@ class PromiseConnection extends EventEmitter {
 
   execute(query, params) {
     const c = this.connection;
-    const localErr = new Error();
     if (typeof params === 'function') {
       throw new Error(
         'Callback function is not available with promise clients.'
       );
     }
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       if (params !== undefined) {
         c.execute(query, params, done);
       } else {
@@ -74,37 +72,34 @@ class PromiseConnection extends EventEmitter {
 
   beginTransaction() {
     const c = this.connection;
-    const localErr = new Error();
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       c.beginTransaction(done);
     });
   }
 
   commit() {
     const c = this.connection;
-    const localErr = new Error();
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       c.commit(done);
     });
   }
 
   rollback() {
     const c = this.connection;
-    const localErr = new Error();
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject, localErr);
+      const done = makeDoneCb(resolve, reject);
       c.rollback(done);
     });
   }
 
   ping() {
     const c = this.connection;
-    const localErr = new Error();
     return new this.Promise((resolve, reject) => {
       c.ping((err) => {
         if (err) {
+          const localErr = new Error();
           localErr.message = err.message;
           localErr.code = err.code;
           localErr.errno = err.errno;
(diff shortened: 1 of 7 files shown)


chrisveness:
Would it be possible to communicate the caller (application) filename / line-number where the error was triggered?

It has become much harder to debug MySQL errors without knowing where in the application code they originated.

sidorares:
> Would it be possible to communicate the caller (application) filename / line-number where the error was triggered?
> 
> It has become much harder to debug MySQL errors without knowing where in the application code they originated.

This was the original intent of manual error construction, and this PR should not change this mostly

Could you help testing this @chrisveness ? Create a simple repro case, and compare stack trace before / after this version

Also maybe we need to reevaluate if we still have to do this manually - v8 captures async stack traces manually, maybe we can rely on that

chrisveness:
Create file `bad-query.js`:

~~~~js

const connection = await mysql.createConnection({
  host:     'localhost',
  user:     'user',
  database: 'db',
  password: 'pw',
});

const sql = 'SELECT nosuchcolumn FROM information_schema.tables';
const [ results, fields ] = await connection.query(sql);
console.log(results);
await connection.end();
~~~~

Run:

    $ npm install mysql2@3.21.1
    $ node bad-query.js

Gives:

~~~~
.../mysql2-4257/node_modules/mysql2/lib/promise/connection.js:29
    const localErr = new Error();
                     ^

Error: Unknown column 'nosuchcolumn' in 'field list'
    at PromiseConnection.query (.../mysql2-4257/node_modules/mysql2/lib/promise/connection.js:29:22)
    at file://.../mysql2-4257/bad-query.js:11:46
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5) {
  code: 'ER_BAD_FIELD_ERROR',
  errno: 1054,
  sql: 'SELECT nosuchcolumn FROM information_schema.tables',
  sqlState: '42S22',
  sqlMessage: "Unknown column 'nosuchcolumn' in 'field list'"
}
~~~~

Then run:

    $ npm install mysql2@3.22.0
    $ node bad-query.js

Gives:

~~~~
.../mysql2-4257/node_modules/mysql2/lib/promise/make_done_cb.js:6
      const localErr = new Error();
                       ^

Error: Unknown column 'nosuchcolumn' in 'field list'
    at Query.onResult (.../mysql2-4257/node_modules/mysql2/lib/promise/make_done_cb.js:6:24)
    at Query.execute (.../mysql2-4257/node_modules/mysql2/lib/commands/command.js:36:14)
    at Connection.handlePacket (.../mysql2-4257/node_modules/mysql2/lib/base/connection.js:552:34)
    at PacketParser.onPacket (.../mysql2-4257/node_modules/mysql2/lib/base/connection.js:102:12)
    at PacketParser.executeStart (.../mysql2-4257/node_modules/mysql2/lib/packet_parser.js:75:16)
    at Socket.<anonymous> (.../mysql2-4257/node_modules/mysql2/lib/base/connection.js:109:25)
    at Socket.emit (node:events:508:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5) {
  code: 'ER_BAD_FIELD_ERROR',
  errno: 1054,
  sql: 'SELECT nosuchcolumn FROM information_schema.tables',
  sqlState: '42S22',
  sqlMessage: "Unknown column 'nosuchcolumn' in 'field list'"
}
~~~~

... which has no reference to bad-query.js line number.

(Sometimes it might be straightforward to grep for the SQL string, but not always).

sidorares:
yeah, I should have reviewed this a bit more carefully
when we moved stack capture inside callback it broke original funcionality of capturing command stack

I'll see current state of native async stack traces but might actually revert this change if there's no better option

We could have this optional with the flag, but I'm not sure how big of a perf impact are we talking realistically compared to a full round trip of an average query. If it's really small flag not worth it and I'll just revert

sidorares:
I think it's best to revert.

Alternatives considered:

'lightweight stacks' -

```js
  const obj = {};
  Error.captureStackTrace(obj, captureCallSite);
```

Same performance as `new Error()` in node, in Bun `new Error()` is actually faster compared to ` Error.captureStackTrace`

native async stack traces - only captures `await` boundary.

sidorares:
@chrisveness reverted, releasing as v3.22.1

PR #4265, opened 2026-04-16
Revert "perf: defer Error object creation to error handlers in promise wrappers"

sidorares:
Reverts sidorares/node-mysql2#4257

### DIFF
--- a/lib/promise/connection.js
+++ b/lib/promise/connection.js
@@ -26,13 +26,14 @@ class PromiseConnection extends EventEmitter {
 
   query(query, params) {
     const c = this.connection;
+    const localErr = new Error();
     if (typeof params === 'function') {
       throw new Error(
         'Callback function is not available with promise clients.'
       );
     }
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject);
+      const done = makeDoneCb(resolve, reject, localErr);
       if (params !== undefined) {
         c.query(query, params, done);
       } else {
@@ -43,13 +44,14 @@ class PromiseConnection extends EventEmitter {
 
   execute(query, params) {
     const c = this.connection;
+    const localErr = new Error();
     if (typeof params === 'function') {
       throw new Error(
         'Callback function is not available with promise clients.'
       );
     }
     return new this.Promise((resolve, reject) => {
-      const done = makeDoneCb(resolve, reject);
+      const done = makeDoneCb(resolve, reject, localErr);
       if (params !== undefined) {
         c.execute(query, params, done);
       } else {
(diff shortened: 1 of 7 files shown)


sidorares:
Maybe a good idea to add test covering actual correct stack behavior ( or did we miss regression because the change in tests? )

Also, I think we want to pop one frame from the stack here:

```
.../mysql2-4257/node_modules/mysql2/lib/promise/connection.js:29
    const localErr = new Error();
                     ^

Error: Unknown column 'nosuchcolumn' in 'field list'
    at PromiseConnection.query (.../mysql2-4257/node_modules/mysql2/lib/promise/connection.js:29:22)
    at file://.../mysql2-4257/bad-query.js:11:46
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5) {
  code: 'ER_BAD_FIELD_ERROR',
  errno: 1054,
  sql: 'SELECT nosuchcolumn FROM information_schema.tables',
  sqlState: '42S22',
  sqlMessage: "Unknown column 'nosuchcolumn' in 'field list'"
}
```

Should show `bad-query.js:11:46` as a source of error

wellwelwel (approved):
To test the expected behavior, I tried something like:

```ts
import type { ConnectionOptions } from '../../../index.js';
import path from 'node:path';
import process from 'node:process';
import { fileURLToPath } from 'node:url';
import ErrorStackParser from 'error-stack-parser';
import { describe, it, skip, strict } from 'poku';
import { createConnection as promiseCreateConnection } from '../../../promise.js';
import { config } from '../../common.test.mjs';

if (typeof Deno !== 'undefined') skip('Deno: connect ECONNREFUSED');

type Err = Error & { code?: string };

const __filename = fileURLToPath(import.meta.url);
const __basename = path.basename(__filename);

const assertStack = (err: Error) => {
  const frames = ErrorStackParser.parse(err);
  const callerFrame = frames.find(
    (f) => f.fileName === __filename || f.fileName?.endsWith(__basename)
  );
  strict.equal(
    callerFrame?.fileName,
    __filename,
    `Expected error stack to reference ${__basename}; got:\n${err.stack}`
  );
};

await describe('Async stack traces', async () => {
  const createConnection = async (args?: ConnectionOptions) => {
    if (!args && process.env.MYSQL_CONNECTION_URL) {
      return promiseCreateConnection({ uri: process.env.MYSQL_CONNECTION_URL });
    }
    return promiseCreateConnection({ ...config, ...args });
  };

  const conn = await createConnection();

  await it('connection error stack should reference caller', async () => {
    await strict.rejects(
      createConnection({ host: '127.0.0.1', port: 33066 }),
      (err: Err) => {
        strict.equal(err.code, 'ECONNREFUSED');
        assertStack(err);
        return true;
      }
    );
  });

  await it('query error stack should reference caller', async () => {
    await strict.rejects(conn.query('syntax error'), (err: Err) => {
      strict.equal(err.code, 'ER_PARSE_ERROR');
      assertStack(err);
      return true;
    });
  });

  await it('execute error stack should reference caller', async () => {
    await strict.rejects(conn.execute('syntax error'), (err: Err) => {
      strict.equal(err.code, 'ER_PARSE_ERROR');
      assertStack(err);
      return true;
    });
  });

  await conn.end();
});
```

But I couldn't think of any way to keep both the expected behavior and the improvement from #4257 😕

sidorares:
Let's keep this PR revert only and have improvements separately ( test, unnecessary top stack frame removed )

Issue #3982, opened 2025-12-18
Type of `value` is incorrect (Type is `any`, should be more specific)


uPaymeiFixit:
The type of [execute's `values` parameter](https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/protocol/sequences/promise/ExecutableBase.d.ts#L11) is `any` (so basically no type definition). Admittedly I'm not 100% sure I've found all the possibilities for what values can be, but I do know if you pass in an `undefined` value (for example `connection.execute("SELECT 1 + ? + ?", [1, 2, undefined, 4]);` it will throw an error. 

I've noticed LLMs are pretty good at adhering to type definitions, but love to take advantage of the `any` here. Almost got us into trouble at work. How do we feel about using something like the following definition, which is as close a definition I can think of for "any, except undefined or an array with undefined".

```typescript
type Value =
  | string                    // <-- primitive
  | number                    // <-- primitive
  | boolean                   // <-- primitive
  | symbol                    // <-- primitive
  | bigint                    // <-- primitive
  | null                      // <-- primitive
  | Record<string, unknown>   // <-- Object
  | ({} | null)[];            // <-- (any - undefined) array
```

I'll create a PR if this (or something else) looks good.

wellwelwel:
I find it interesting, but it's important to consider the probability to which this could be a breaking change for **TypeScript** users. I suppose this could be resolved by mapping the types that aren't actually supported.

Taking it out of a [_personal_ **ORM** project](https://github.com/wellwelwel/mysql2-orm/blob/01755e1b3e8caccc5645813b436c3854fe111b9f/src/types/param.ts):

```ts
export type Param =
  | string
  | number
  | bigint
  | boolean
  | Date
  | null
  | Blob
  | Buffer;
```

uPaymeiFixit:
@wellwelwel is https://github.com/sidorares/node-mysql2/pull/3985 what you had in mind?

PR #4276, opened 2026-04-22
fix(types): accept Record types in query and execute values

### LINKED ISSUE #4275: Regression: ExecuteValues no longer allows Record type
Initially the type was `any`, then #3985 disallowed a few primitive types, then #4133 tightened the array branch of `ExecuteValues` from `({} | null)[]` to `ExecuteValues[]`. This forces nested objects to resolve through the recursive object branch `{ [key: string]: ExecuteValues }`, which rejects `Record<string, unknown>` at the type level even though it works at runtime. It looks like `QueryValues` probably has a similar issue, but when passing a Record in at the top level instead of inside an array. 

Passing in Records is a pretty common thing to do, either from untyped input, ORMs, or just structuring the data as JSON and typing it as a Record (what our company is doing), so I'd have to imagine this is having an impact on at least a few others.

```typescript
const value: Record<string, unknown> = { hello: 'world' };
conn.execute('INSERT INTO data (json_col) VALUES (?)', [value]);
// Type 'Record<string | number | symbol, unknown>' is not assignable to type 'ExecuteValues'.
```

This might be a good compromise. There are certainly edge cases where mistakes can be made, but for the most part I think it sits nicely between `any` and being too strict / overly prescriptive. 

```diff
 export type ExecuteValues =
   | string
   | number
   | bigint
   | boolean
   | Date
   | null
   | Blob
   | Buffer
   | Uint8Array
   | ExecuteValues[]
-  | { [key: string]: ExecuteValues };
+  | { [key: string]: unknown };

 export type QueryValues =
   | string
   | number
   | bigint
   | boolean
   | Date
   | null
   | undefined
   | Blob
   | Buffer
   | Uint8Array
   | Raw
   | ({} | null | undefined)[]
-  | { [key: string]: QueryValues };
+  | { [key: string]: unknown };
```
https://github.com/sidorares/node-mysql2/blob/77afd80102673e30c121196ac614748bb748c49d/typings/mysql/lib/protocol/sequences/Query.d.ts#L7-L33

In case it helps, I edited my original [tsplayground](https://www.typescriptlang.org/play/?target=7#code/PTAEF5K6dv-gKBKAtgTwM4EcA2AmUARQFcBTAJ3QDUBDXczUAFzM2aYDJQBLAOwBubZjwDmtEQHs+yMAnkLFkRIh6oADpIrNQAb1AAhEgDNjlUAF9QxipNSgA5H0kATMgC4ARibMUHKlAAFACUAUWoASQB5AFUAZVAAWlAAFQALHiZM0AB3NIkWNLJicio6BjZQAAp6TElQUIAPMgBjElZyxgBKUBJMMhcWes8yRDJGzW0WdHVi0kp0Jtb2sk62KNxB8ERQUAAfUHYKflEd-dA+ElQRijODzzF+ZjvQT0lJXDJaGV2DgBEJKNfhcSLhcC8DLhJJ4IT5KC8qrorAdLmCugBtAC6L306IA1mR0O5DsxjnxRJjifMqEs2h16IwNoMLABuUAoAA8yT4ZByATAAGEYsEwgA5FJJBrNOmrBmVKpiZxk0QlBZrTBdMYTLQ6ZgzYq0lbqgUkCgUMh8HTbYFHE4vS7XeHAh6iJ4vN4fL4-c4A1j20Hg4GQ6Gw0xO84xJ4ADgAgmbaOgXob6RVMCazRbmFicaB8YTibbyZSpcsU4x0+bLZYWfzQCEooEonFQn9JYayMx1dVFVoTqqynKNVrJrr9SWZerArZNP0ti9C6dgQ6bu7Hpb3e9Pt8Xr6gedUYHzsGYUG4bdgZHLbH44ngcnZamp5IZwNs8DcQSiSTlcWSHw8c4OR8NWtZ6rM-Y0IOEAQfe6pMjWYFzKUkGptBsGDhWmY1igiEQV24Djkag5Pi+Lg1rIoAALKSC0eLWH+LRSMBzD1Kw7CFMUiGIG4LS4LQ5r0XwjE8NIoDYMhVQ4LgBakicAA0oACIOAD8VLIeqXTEgIkg8GRtZKAZhnKCggCg5KA1AxgAMhErbxDGADioRxBRRmuQotbBK0WguIgLTSOxSkVMSnl+RQLgcgu+5XDc5yYOg1wfApf4AZIQEAHzQfoRRgpIxIODkWibA4IHiQsVQOBEorNsEEqVSkUSgC4Ei0NUABWdR8AA+n5uA9BZlkxE51TKV0DgKeigXkJiXTkSgcQnJ8oDqMcqA8CIQiKYOiClVQ5XNpZoQChKABUoAAGLBFElG9P0FBMAA6gAEqEYS8FsoDKWNoAAIwzdtEkOPth0nedl3XX0lAPc9r18LQqDFARn0KQ4rWSGkfAOH9O3oHtoQHUdoCnRdV03ZDoBPS9oSgLQwkbYjX2kuQWMA0DBNE2DpN3eT0NU24nysC4nUFPTCkHn9FEGGIERVsYWigHxFCiMUy6Q-9ZWA3jwOE6DJPeWTFOvbp0FI6AACcAAM5sAOzfabpv4AArAALFbTvm3bADMfDiygu6gNCrWtBwau7Rr+Mg8T11kEIlpQ5ToAtOagKCwUGUmzyOSgLuVRdD7YBGGGFDWHLDyw1QjXNSHOMVVVL21eKDXGDwnxMFUTXMLQfVWYNCRVCNX0F74AB0Nh2OV2VQpjeeGFCnjF0Xpf8egFcd1X5WVdV9f1bwqC0Errft535nd0NfejaLvIz9CVTog4h8ONN09xhQCZMJIxibahVSoJI7F+agP9gJ9D3mQTU2Ncbh21pHTmcdDbvWUqAGMopWyw3hsbL66JvrI1RujB+zN1aswjhzCGXMDZUyNpVYaCllLUPPrmLBoB8AKQ9tNGs4Ca6b1AHVBqJDW6oLIApEBgjaagOPgNU+NCPq0IwSjNGGNmHmwUozMgrDazP1frkNaaQQRgk-owNeDgYiBD+DGFIVNeGgGbBKPmHYBhCytB9bm8cjYi1zAeBS31VEUXUVgTRzBtGrWaIMRCmADEbzrlwhu8tJCiFbiQzqukFLw0wJgIRLA1DCDhuoYR61RH9R7lQqRRT+7jTOFgs4hjbrRNEErQY-AxpnAzlnQEOc5JnGUW0rxKBRTCAGNTG8oT2GEMcQg0ZGDMEKXwJicaHsFJO0xF0sAPimA5C0f7TwgdGKt1lkXfhgx1B8RaGQNIHw3Bczlv-A54w9FsDAQDcJNVInb0Pm1Dq3UPhd3Eb3EpuZ9CfjypNMgxULCLIaBoPU-SX6JiGZrNmOtwa3UwBg0FPi-HaMlqIaWzwYWQPZiTBWSsEkuFgeQ+BGCLbW1tvbZ2rt3amy9qC0UcM+kHJpsc055gqgByDhqUABjhl4oRfrHmb1oLuF0l9fQuliTfUsNPNyCq4AUUADLkkT+o2VAHZRyvd-FsGKKs3RmATmZ0oLYO6moUCKqtVAWsAABDgiRxizEYo6s0ctkjoj-G4JuPIXDYhxVrQVMCnFwPQeNL1ZAfUDFBfazAjqJhB1dWayUxgGJMSYPxZWkgdD0ChDkAY-LYVEJJhYshorXE5wgBlX6s0wCxvjc65gSb3WHHim8XAGaBLOBzTlfNPkA1wugaWkVLiPpfTiG2j45U2LMCnrWiA1qrUURSAATUCFTSqZ1KaigFFTUIAANGMlFAgHWcpaxdirawpH1HEROPB1A6FCuaRiuBl78F8EwXV-RfL+R0ICzA301JqiggRJ2+A2S7Eg1B9kYAVa3D8nwAKg58BAYHKhAiDgJ6SAcBBqDKAFw-sQ3+wcHtUMoUYNBCZjDmGYlwzByFCZCNIdTE7Mj+EdG4Do9BlAB4r1jjLrYVZ5JchaDxKE1NQkmJLVsEclJ6oqj-rY4OHougzg8A-lURC78blMCgI4BcmM9BnF2Ahuonwh5QlEApwcQ8WIxHULMCgApaD9BzjNejkV4b+NcBmpSzdaCeE+GcKwZAO3FHU9ULTH9-0Low3BwzqmoOmc9BZmJ1nUy2ckGdHgQSqj4FzmyHj0VzBeZOcS6mfm+KBb3CFsLvANM+KHpkHx6XugqeMwnfyKXLOtbYBZi0oh-HuZQPxKFaAOxld87Qfz1XguIAsIgIAA) to add Records.

uPaymeiFixit:
## Closes #4275

- Loosen the object branch on both types to `{ [key: string]: unknown }`. Top-level and array-branch strictness stay intact, so `undefined` inside an `execute` values array is still rejected at compile time.
- Adds a type-level regression test at `test/tsc-build/strict-checks/values.test.ts` covering top-level and nested `Record<string, unknown>` for both callback and promise APIs.
- Updates the `QueryValues` docs section to cover `ExecuteValues`, fix stale type signatures from before #4133, and document the `undefined` behavior difference between `.query()` and `.execute()`.

### DIFF
--- a/test/tsc-build/strict-checks/values.test.ts
+++ b/test/tsc-build/strict-checks/values.test.ts
@@ -1,0 +1,35 @@
+/**
+ * This test strictly validates the accepted shapes of the `values` parameter.
+ * See https://github.com/sidorares/node-mysql2/issues/4275.
+ */
+
+import { mysql, mysqlp } from '../index.test.js';
+import { access, sqlPS } from '../promise/baseConnection.test.js';
+
+// Callbacks
+{
+  const conn = mysql.createConnection(access);
+  const params: Record<string, unknown> = { id: 1, name: 'alice' };
+  const jsonValue: Record<string, unknown> = { hello: 'world' };
+
+  conn.execute(sqlPS, params, () => {});
+  conn.query(sqlPS, params, () => {});
+
+  conn.execute(sqlPS, [1], () => {});
+  conn.query(sqlPS, [1, 'x', null], () => {});
+
+  conn.execute('INSERT INTO data VALUES (?)', [jsonValue], () => {});
+  conn.query('INSERT INTO data VALUES (?)', [jsonValue], () => {});
+
+  // @ts-expect-error: undefined is not a valid execute bind parameter
+  conn.execute(sqlPS, [undefined], () => {});
+}
+
+// Promise
+(async () => {
+  const conn = await mysqlp.createConnection(access);
+  const params: Record<string, unknown> = { id: 1, name: 'alice' };
+
+  await conn.execute(sqlPS, params);
+  await conn.query(sqlPS, params);
+})();
--- a/typings/mysql/lib/protocol/sequences/Query.d.ts
+++ b/typings/mysql/lib/protocol/sequences/Query.d.ts
@@ -15,7 +15,7 @@ export type ExecuteValues =
   | Buffer
   | Uint8Array
   | ExecuteValues[]
-  | { [key: string]: ExecuteValues };
+  | { [key: string]: unknown };
 
 export type QueryValues =
   | string
@@ -30,7 +30,7 @@ export type QueryValues =
   | Uint8Array
   | Raw
   | ({} | null | undefined)[]
-  | { [key: string]: QueryValues };
+  | { [key: string]: unknown };
 
 export interface QueryOptions {
   /**
--- a/website/docs/documentation/typescript-examples.mdx
+++ b/website/docs/documentation/typescript-examples.mdx
@@ -381,11 +381,11 @@ For `ProcedureCallPacket<RowDataPacket[]>`, please see the following examples.
 
 <hr />
 
-### QueryValues
+### QueryValues and ExecuteValues
 
 <Stability level={2} />
 
-The `QueryValues` type represents the allowed types for query parameter values:
+The types for the `values` argument to `.query()` and `.execute()` respectively:
 
 ```ts
 import mysql, { RowDataPacket, QueryValues } from 'mysql2';
@@ -406,7 +406,7 @@ conn.query<RowDataPacket[]>(
 );
 ```
 
-The `QueryValues` type includes the following types:
+`QueryValues` accepts:
 
 - `string`
 - `number`
@@ -414,10 +414,23 @@ The `QueryValues` type includes the following types:
 - `boolean`
 - `Date`
 - `null`
+- `undefined`
 - `Blob`
 - `Buffer`
-- `({} | null)[]` (array of values)
-- `{ [key: string]: QueryValues }` (named placeholders)
+- `Uint8Array`
+- `Raw` (raw SQL fragments, from `sql-escaper`)
+- `({} | null | undefined)[]` (array of values)
+- `{ [key: string]: unknown }` (named placeholders)
+
+`ExecuteValues` accepts the same types with these exceptions:
+
+- No `undefined`. Passing `undefined` as a bind parameter throws `TypeError: Bind parameters must not contain undefined. To pass SQL NULL specify JS null`.
+- No `Raw`. Prepared statements send values separately from the SQL text, so raw fragments can't be spliced in.
+- Nested arrays use `ExecuteValues[]` rather than `({} | null | undefined)[]`.
+
+:::note
+`.query()` substitutes `undefined` as `NULL` in the generated SQL, while `.execute()` throws. Prefer explicit `null` to keep behavior consistent between the two.
+:::
 
 <hr />
 


wellwelwel:
Thanks, @uPaymeiFixit! I saw the #4275 and I can't see this as a regression.

In my view, it shouldn't be **MySQL2**'s role to identify and perform narrowing of `unknown` into defined types (`QueryValues`, `ExecuteValues`). In this case, it would be more viable to go back to `any` and accept any value from the users and let them discover invalid types at runtime (which is what this _PR_ is doing when a user passes objects).

Do you think we should revert both to `any`? The way I see it, a user who wants to deal with `unknown` without handling it and send it directly to the database should do something like:

```ts
const value: Record<string, unknown> = { hello: 'world' };
query('INSERT INTO data (json_col) VALUES (?)', [value as QueryValues]); // <-
```

Full example:

```ts
export type QueryValues = /* or ExecuteValues */
   | string
   | number
   | bigint
   | boolean
   | Date
   | null
   | Blob
   | Buffer
   | Uint8Array
   | QueryValues[]
   | { [key: string]: QueryValues };

declare function query(sql: string, values?: QueryValues): void;

const value: Record<string, unknown> = { hello: 'world' };
query('INSERT INTO data (json_col) VALUES (?)', [value as QueryValues]);
```

uPaymeiFixit:
Sure, let me share my perspective on both directions: `any` _vs._ disallowing `Record` types. 

But let me start with this. I think TypeScript has two huge benefits / goals: preventing runtime errors, and providing documentation. I think both of those directions breaks one or both of those goals. 

### Reverting to `any`

My big motivation for creating the types in the first place, #3982, was because our organization was seeing runtime errors in production; not during testing. Primarily due to passing `undefined` values sometimes even though `undefined` is not a supported type. Both Claude Code and our own developers were passing in complicated types which contained possibly undefined values that really should have been filtered out or replaced with `null`. 

`any` is just saying "don't use TypeScript here", so you lose any benefit of TypeScript for that variable. As MySQL2's types get better developers will start blindly trusting them more and more and so I predict an incorrect type such as `any` will actually start causing more issues.

### Keeping `{ [key: string]: QueryValues }`

My main issue with this is that it feels _too_ restrictive. Both my PR and the existing type are incorrect. In the existing case Record types _are_ allowed, but the type says they aren't. This forces the user, who may not be as familiar with MySQL2, to know that the type is incorrect and to use a type assertion (like your example). My personal experience with this was noticing that upgrading MySQL2 broke a lot of our types, and when Claude Code was fixing them it tried to stringify all of our objects. It took me far too long to remember that MySQL2 actually _did_ support sending objects and to look into what I later discovered to be the new more restrictive `ExecuteValues` type. 
**tl;dr** - it breaks documentation, new developers will assume MySQL doesn't support sending objects

### What this PR's type does and doesn't do

I want to be clear, all three options (`any`, the existing type, and this PR) are all incorrect types. They all lie to the user about which types will result in a smooth runtime. I'm just trying to find the one that fixes the most common use cases. 

Both the existing type and this PR prevent passing in direct `undefined` values; that's the biggest win IMO as it prevents runtime errors that we were _actually_ experiencing in practice.

My PR allows through some types that likely aren't intended, but also don't throw and likely don't even have any significant data impact. For example, passing in `[{ user: { name: 'alice', middleName: undefined } }]` would now be allowed per the type, but at runtime `middleName` wouldn't show up in the database; that's probably fine. I can't find a case where this PR introduces something that hurts the user's experience either by throwing at runtime or significantly changing the data. 

If it helps, here's a table of some use cases / edge cases and how they're handled by the current type (OLD) vs. this PR's type (NEW) vs. the runtime.

I won't die on this hill, but in my opinion every case I could find where the current and new PR's type don't match, this PR's type is what I would prefer given the value. 

---

- **OLD** / **NEW** — ✅ tsc accepts · ❌ tsc rejects
- **Runtime** — what the mysql2 client sends to MySQL for this input. ✅ matches caller intent · ⚠️ silently drops or coerces · ❌ throws before the socket

| Call | OLD | NEW | Runtime |
| --- | --- | --- | --- |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: { name: 'alice', onClick: () => 42 } }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"user":{"name":"alice"}}` — `onClick` omitted |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: { name: 'alice', middleName: undefined } }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"user":{"name":"alice"}}` — `middleName` omitted |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: { id: Symbol('s') } }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"user":{}}` — `id` omitted |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ handlers: [() => 42, () => 42] }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"handlers":[null,null]}` — functions replaced with `null` |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ values: [undefined, 1, 2] }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"values":[null,1,2]}` — `undefined` replaced with `null` |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: { get onClick() { return () => 42 } } }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"user":{}}` — getter result omitted |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ name: 'alice', middleName: undefined }]`<br>`)` | ❌ | ✅ | ⚠️ JSON `{"name":"alice"}` — `middleName` omitted |
| `const r: Record<string, unknown> = { id: 1, name: 'alice' }`<br>`connection.execute(`<br>`  'SELECT :id, :name',`<br>`  r`<br>`)` | ❌ | ✅ | ✅ `?=1`, `?='alice'` (positional after named-placeholder rewrite) |
| `const r: Record<string, unknown> = { hello: 'world' }`<br>`connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [r]`<br>`)` | ❌ | ✅ | ✅ JSON `{"hello":"world"}` |
| `class User { constructor(public name: string) {} }`<br>`connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: new User('alice') }]`<br>`)` | ❌ | ✅ | ✅ JSON `{"user":{"name":"alice"}}` |
| `class User {`<br>`  constructor(public name: string) {}`<br>`  toJSON() { return { serialized_name: this.name } }`<br>`}`<br>`connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: new User('alice') }]`<br>`)` | ❌ | ✅ | ✅ JSON `{"user":{"serialized_name":"alice"}}` |
| `class User {`<br>`  constructor(public name: string) {}`<br>`  toJSON() { return { serialized_name: this.name } }`<br>`}`<br>`connection.execute(`<br>`  'SELECT :id',`<br>`  { id: new User('alice') }`<br>`)` | ❌ | ✅ | ✅ JSON `{"serialized_name":"alice"}` |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [Symbol('x')]`<br>`)` | ❌ | ❌ | ⚠️ `?='Symbol(x)'` — coerced via `toString()` |
| `const r: Record<string, unknown> = { hello: 'world' }`<br>`connection.query(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [r]`<br>`)` | ✅ | ✅ | ⚠️ `INSERT INTO data (json_col) VALUES ('[object Object]')` |
| `connection.query(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: { name: 'alice', onClick: () => 42 } }]`<br>`)` | ✅ | ✅ | ⚠️ `INSERT INTO data (json_col) VALUES ('[object Object]')` |
| `connection.query(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ user: { name: 'alice', middleName: undefined } }]`<br>`)` | ✅ | ✅ | ⚠️ `INSERT INTO data (json_col) VALUES ('[object Object]')` |
| `connection.query(`<br>`  'SELECT ?',`<br>`  [() => 42]`<br>`)` | ✅ | ✅ | ⚠️ `SELECT '() => 42'` — function source inlined as SQL literal |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [new Uint8Array([1, 2, 3])]`<br>`)` | ✅ | ✅ | ⚠️ `?='1,2,3'` — coerced via `toString()`, not sent as binary |
| `const o: { self?: typeof o } = {}`<br>`o.self = o`<br>`connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [o]`<br>`)` | ✅ | ✅ | ❌ throws `TypeError: Converting circular structure to JSON` |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ count: 123n }]`<br>`)` | ✅ | ✅ | ❌ throws `TypeError: Do not know how to serialize a BigInt` |
| `const p = new Proxy({}, { get() { throw new Error('trap') } })`<br>`connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [p]`<br>`)` | ✅ | ✅ | ❌ throws `Error: trap` (user code throws inside serialization) |
| `connection.execute(`<br>`  'SELECT ?, ?, ?',`<br>`  [1, 'x', null]`<br>`)` | ✅ | ✅ | ✅ `?=1`, `?='x'`, `?=NULL` |
| `connection.execute(`<br>`  'SELECT :id, :name',`<br>`  { id: 1, name: 'alice' }`<br>`)` | ✅ | ✅ | ✅ `?=1`, `?='alice'` |
| `connection.execute(`<br>`  'SELECT ?, ?',`<br>`  [new Date('2026-01-01T00:00:00Z'), Buffer.from('hello')]`<br>`)` | ✅ | ✅ | ✅ `?=<DATETIME>`, `?=<BLOB>` |
| `connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [{ hello: 'world', count: 3 }]`<br>`)` | ✅ | ✅ | ✅ JSON `{"hello":"world","count":3}` |
| `const u: { name: string; middleName?: string } = { name: 'alice' }`<br>`connection.execute(`<br>`  'INSERT INTO data (json_col) VALUES (?)',`<br>`  [u]`<br>`)` | ✅ | ✅ | ✅ JSON `{"name":"alice"}` |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [123n]`<br>`)` | ✅ | ✅ | ✅ `?='123'` — coerced via `toString()` |
| `connection.query(`<br>`  'SELECT ?',`<br>`  [undefined]`<br>`)` | ✅ | ✅ | ✅ `SELECT NULL` |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [undefined]`<br>`)` | ❌ | ❌ | ❌ throws `TypeError: Bind parameters must not contain undefined. To pass SQL NULL specify JS null` |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [() => 42]`<br>`)` | ❌ | ❌ | ❌ throws `TypeError: Bind parameters must not contain function(s).` |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [new Map([['a', 1]])]`<br>`)` | ❌ | ❌ | ❌ throws `TypeError` from `StringParser.encode` (unsupported argument type) |
| `connection.execute(`<br>`  'SELECT ?',`<br>`  [new Set([1, 2, 3])]`<br>`)` | ❌ | ❌ | ❌ throws `TypeError` from `StringParser.encode` (unsupported argument type) |

uPaymeiFixit:
I think a useful way to discuss this could be to talk about what challenges we're afraid users will have if we change the types to allow `{ [key: string]: unknown }`. 

For example, a valid concern from the second item in my table above could be:

> I'm concerned a user will write 
> ```typescript
> const values = [{ user: { name: 'alice', middleName: undefined }}];
> const sql = 'INSERT INTO data (json_col) VALUES (?)';
> connection.execute(sql, values)
> ```
> 
> And then be surprised `middleName` was omitted instead of `NULL`, or allowed by the type at all.

Or on the opposite side of the argument, my legitimate concern if we _don't_ make this change is that someone will write code like: 
```typescript
const values: Record<string, unknown> = { hello: 'world' }
const sql = 'INSERT INTO data (json_col) VALUES (?)';
connection.execute(sql, [values]);
```

And the current type won't allow them, so they'll have to loosen their types via type assertions (what my organization is currently having to do) which could lead to bugs in the future if the shape ever changes (and now the assertion asserts the incorrect type).

wellwelwel:
> Or on the opposite side of the argument, my legitimate concern if we _don't_ make this change is that someone will write code like:
> 
> ```ts
> const values: Record<string, unknown> = { hello: 'world' }
> const sql = 'INSERT INTO data (json_col) VALUES (?)';
> connection.execute(sql, [values]);
> ```

My _personal_ point is precisely that this shouldn't be **MySQL2**'s role. For example, as I said in the previous comment:

> In my view, it shouldn't be **MySQL2**'s role to identify and perform narrowing of `unknown` into defined types (`QueryValues`, `ExecuteValues`). In this case, it would be more viable to go back to `any` and accept any value from the users and let them discover invalid types at runtime (which is what this _PR_ is doing when a user passes objects).

I see this exactly the same way as using:

```ts
const values: unknown = 'Hey U 🌞';
const sql = 'INSERT INTO data (msg) VALUES (?)';

connection.execute(sql, [values]);
```

Thinking this way, we would need to define `unknown` as a whole (not just for `Record<string, unknown>`), but **MySQL2** doesn't wait for any input and then validates according to each `typeof`, for example. Instead, it can indeed break (intentionally) with incorrect parameters. At the user level, using `unknown` for this case would have the same effect as `any`, because the end user will be able to pass any parameter and none will throw an error, besides implying that there is handling for all types internally, which doesn't actually happen.

Regardless of `unknown`, is there a specific type that should be accepted but isn't covered by the current types, or the opposite, accepting some type that shouldn't be accepted?

wellwelwel:
Justifying why I say this shouldn't be **MySQL2**'s responsibility:

This could set a precedent like _"if it's a parameter, it's `unknown`"_ throughout the entire codebase, since any user could _"mask"_ a type behind `unknown` (like in the example I showed with `const values: unknown = 'Hey U 🌞'`). This wouldn't only affect **MySQL2**, but practically all **TypeScript** projects where a user can define a parameter and _force_ `unknown`.

On the side of the user who actually wants to use `unknown` but doesn't want to handle it before sending it as a parameter (_speaking_ at the **API** usage level), they should take an approach like `as ExecuteValues` or `as QueryValues`, not because of **MySQL2**, but because of **TypeScript** itself. If the user really wants to send **any** value and "override" the valid types, then they can use `as any` (it's not a recommendation 😅). For example, backing to `Record`:

```ts
type A = Record<string, ExecuteValues>;
type B = Record<string, QueryValues>;
type C = Record<string, string | number | "etc.">;
type D = Record<string, any>;
```

---

@uPaymeiFixit, if you want, I can show some real examples of `unknown` as a parameter that is actually handled internally in another project I develop _vs._ when the type is `unknown` but the responsibility is kept with the user 🙋🏻‍♂️

PR #4278, opened 2026-04-23
fix: allow resetOnRelease in connection config validation

### LINKED ISSUE #4277: add resetOnRelease to valid connection options
## Summary

This PR fixes an inconsistency around the `resetOnRelease` pool option.

`resetOnRelease` is already:
- documented in the TypeScript pool typings
- parsed and stored in `PoolConfig`
- used by the pool release logic

However, it is not included in `validOptions` inside `lib/connection_config.js`.

Because of that, passing `resetOnRelease: true` to `createPool()` works functionally, but still emits the warning below:

> Ignoring invalid configuration option passed to Connection: resetOnRelease

This is misleading for users, and the warning message also says this may become an error in future versions.

## Problem

Current behavior is inconsistent across the codebase:

- `typings/mysql/lib/Pool.d.ts` defines `resetOnRelease?: boolean`
- `lib/pool_config.js` reads and stores `resetOnRelease`
- `lib/base/pool.js` actively uses `this.config.resetOnRelease`
- but `lib/connection_config.js` does not recognize it as a valid option

As a result, a supported pool option is treated as invalid during option validation.

## Proposed fix

Add `resetOnRelease` to `validOptions` in `lib/connection_config.js`.

## Why this change

This keeps runtime validation aligned with:
- the public typings
- the pool configuration parser
- the actual pool implementation

It also avoids unnecessary warnings for a supported feature.

## Minimal change

```diff
diff --git a/lib/connection_config.js b/lib/connection_config.js
index XXXXXXX..YYYYYYY 100644
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@
   // These options are used for Pool
   connectionLimit: 1,
   maxIdle: 1,
   idleTimeout: 1,
   Promise: 1,
   queueLimit: 1,
+  resetOnRelease: 1,
   waitForConnections: 1,
   jsonStrings: 1,
   gracefulEnd: 1,
 };

saivn91:
Fixes #4277

## Summary

`resetOnRelease` is already supported by the pool config and pool release logic,
but it is missing from `validOptions` in `lib/connection_config.js`.

Because of that, using `resetOnRelease: true` with `createPool()` works, but still prints:

> Ignoring invalid configuration option passed to Connection: resetOnRelease

This PR adds `resetOnRelease` to `validOptions` so the runtime validation matches the actual supported pool options.

### DIFF
--- a/lib/connection_config.js
+++ b/lib/connection_config.js
@@ -67,6 +67,7 @@ const validOptions = {
   idleTimeout: 1,
   Promise: 1,
   queueLimit: 1,
+  resetOnRelease: 1,
   waitForConnections: 1,
   jsonStrings: 1,
   gracefulEnd: 1,


wellwelwel (approved):

sidorares:
Thanks @saivn91 ! I edited description to point to the issue you created, the change fixes just that

PR #4291, opened 2026-05-07
fix(pool): reject queued requests on end

gilboom:
## What changed

Reject connection requests that are already queued in `BasePool._connectionQueue` when `pool.end()` closes the pool.

## Why

When a promise pool is saturated and additional `pool.query()` calls are waiting for a connection, calling `pool.end()` marks the pool as closed and closes the active connections, but the queued callbacks are never invoked. The promise wrappers for those queued queries can stay pending forever.

This makes queued requests behave consistently with new `getConnection()` calls after pool closure, which already receive `Pool is closed.`.

## Validation

- Added an integration regression test that saturates a promise pool, waits until two queries are queued, calls `pool.end()`, and asserts the queued queries reject with `Pool is closed.` instead of timing out.
- Verified the new test fails before the fix with `timed out waiting for queued queries to settle; queue length=2`.
- `CI=1 MYSQL_PASSWORD=root MYSQL_DATABASE=mysql node --import tsx test/integration/test-pool-end.test.mts`
- `npm run lint`

### DIFF
--- a/lib/base/pool.js
+++ b/lib/base/pool.js
@@ -185,6 +185,10 @@ class BasePool extends EventEmitter {
         }
       };
     }
+    while (this._connectionQueue.length > 0) {
+      const queuedCallback = this._connectionQueue.shift();
+      process.nextTick(() => queuedCallback(new Error('Pool is closed.')));
+    }
     let calledBack = false;
     let closedConnections = 0;
     let connection;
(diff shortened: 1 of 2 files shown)


wellwelwel (approved):
Thanks, @gilboom! I just confirmed the bug, and the test does indeed fail without the fix 🙋🏻‍♂️

PR #4327, opened 2026-06-05
fix: keep 00:00:00 time for TIMESTAMP in binary protocol with dateStrings

### LINKED ISSUE #4321: BUG : Timestamp 00:00:00 converted to Date format when using dateStrings: true
How to reproduce:

Timestamp column with value : 2025-05-17 00:00:00
Retrieve with dateStrings: true => 2025-05-17 => broken
Expected result with dateStrings: true => 2025-05-17 00:00:00

Test with 2025-05-17 00:00:01
Retrieve with dateStrings: true => 2025-05-17 00:00:01 => working properly

Timestamp should not be converted to Date, even if 00:00:00

gurusharankumarram1:
With `dateStrings: true` and the binary protocol (execute / prepared statements), a TIMESTAMP value whose time is exactly 00:00:00 was returned as a date only — e.g. "2026-05-31" instead of "2026-05-31 00:00:00". The text protocol (query) was unaffected.

MySQL's binary protocol omits an all-zero time component and sends only the date (length === 4), so the driver rebuilds the string in `Packet.readDateTimeString()`. This is the same bug class as #1040 ("datetime fields returned as date when time is 00:00:00"), fixed in #3204 — but that fix restored " 00:00:00" for Types.DATETIME only, leaving Types.TIMESTAMP still
affected. This extends the same handling to TIMESTAMP, matching DATETIME and the text protocol.

Fixes #4321

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -216,12 +216,12 @@ MySQL2 is a **library**, not a long-running app. Development means installing No
 
 ### Services
 
-| Service | Required? | Notes |
-| --- | --- | --- |
-| **Node.js** (≥14; CI uses 22) | Yes | `npm ci` at repo root |
-| **MySQL** | Yes for full tests | Integration/global tests need a `test` database |
-| **Docker** | Recommended | Used to run MySQL the same way as CI |
-| **Docusaurus** (`website/`) | Optional | `cd website && npm ci && npm start` → http://localhost:3000 |
+| Service                       | Required?          | Notes                                                       |
+| ----------------------------- | ------------------ | ----------------------------------------------------------- |
+| **Node.js** (≥14; CI uses 22) | Yes                | `npm ci` at repo root                                       |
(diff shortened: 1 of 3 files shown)


gurusharankumarram1:
To keep this change focused and easy to review, the fractional-seconds handling was intentionally left untouched in this PR.

In the binary protocol, fixed-scale temporal values are currently returned without their trailing zero fraction — for example, a `TIMESTAMP(3)` value of `...42.000` is returned as `...42` — which is the behavior covered by the existing test for #617. The text protocol, by contrast, preserves the declared scale and returns `...42.000`.

Making the binary protocol pad these fractional zeros to stay consistent with the text protocol would change that long-standing behavior, so it seems best treated as a deliberate, separate decision rather than bundled into this fix. It could be handled either as a dedicated issue or by extending this PR.

sidorares:
lgtm, thanks @gurusharankumarram1 !

PR #4373, opened 2026-07-05
feat: support MariaDB data types (UUID, INET4, INET6, VECTOR, JSON) via extended type metadata; run CI against MariaDB

### LINKED ISSUE #1287: JSON doesn't auto parse on MariaDB
Given a simple `users` table

| field  | type                 |
| ------ | -------------------- |
| userId | int(10)              |
| saved  | longtext	utf8mb4_bin |

With a single row

userId: 1, saved: {"value": 2}

```javascript
    
    async function start() {
    		host: 'localhost',
    		user: 'root',
    		database: 'database'
    	});
    	connection.query('SELECT saved FROM users WHERE userId = 1', function(err, results, fields) {
    		console.log(results);
    	})
    }
    
    start();
```

Will return

[ TextRow { saved: '{"value": 1}' } ] in MariaDB
[ TextRow { saved: { value: 1 } } ] in MySQL

vlasky:
# Support MariaDB data types: UUID, INET4, INET6, VECTOR, and the JSON alias

MariaDB sends its own data types (`UUID`, `INET4`, `INET6`, `VECTOR`, and `JSON`, which is an alias for `LONGTEXT`) as standard string/blob types on the wire, so MySQL2 could not distinguish them from `CHAR`/`VARBINARY`/`LONGTEXT` columns. This PR negotiates MariaDB's `MARIADB_CLIENT_EXTENDED_METADATA` capability (MariaDB 10.5+) so the real type is visible, and fixes the issues that surfaced along the way.

The wire format was verified against the MariaDB protocol documentation, the MariaDB server source (through 11.8 and `main`), and a live MariaDB 11.8 server.

Fixes #1287 (with the extended metadata negotiated, MariaDB JSON columns and JSON expressions such as `JSON_EXTRACT(...)` parse to objects by default; `jsonStrings: true` keeps strings).

This PR also makes the existing test suite pass against MariaDB and adds `mariadb:11.8` to the Linux CI matrix, resolving the `TODO - add mariadb to the matrix` in `ci-linux.yml` (originally split out as #4374, consolidated here per review feedback).

### Extended type metadata (`MARIADB_CLIENT_EXTENDED_METADATA`)

- The MariaDB extended server capabilities are parsed from the 4 reserved bytes of the initial handshake packet (present when the server leaves `CLIENT_MYSQL`/`LONG_PASSWORD` unset, i.e. genuine MariaDB 10.2+ servers only; MySQL and MySQL-compatible proxies are untouched).
- When the server offers `MARIADB_CLIENT_EXTENDED_METADATA`, the client requests it in the last 4 reserved bytes of the handshake response and SSL request, and clears `CLIENT_MYSQL` (the server only reads the extended bytes when that bit is unset, mirroring its own side of the negotiation).
- Column definition packets then carry a length-encoded metadata block, exposed on every field object (including the `typeCast` field wrapper) as:
  - `extendedTypeName`: e.g. `'uuid'`, `'inet4'`, `'inet6'`, `'point'`
  - `extendedFormat`: e.g. `'json'`

Both properties are `undefined` on MySQL and on MariaDB < 10.5. Parsing is bounds-checked, so a malformed block from a buggy server/proxy cannot desynchronise the packet.

### Per type

| Type | Behavior |
| --- | --- |
| `UUID`, `INET4`, `INET6` | Returned as strings (unchanged); now identifiable via `extendedTypeName`. |
| `JSON` | Identified via `extendedFormat: 'json'` and parsed into objects by default, matching MySQL's native JSON type. `jsonStrings: true` keeps strings. |
| `VECTOR` | Returned as a `Buffer` of little-endian float32 values. MariaDB attaches no extended metadata to VECTOR columns (verified in the server source through 11.8 and `main`), so they are genuinely indistinguishable from `VARBINARY` on the wire; the docs show how to convert and insert. |
| `GEOMETRY` subtypes | Parsed as before; the subtype (`point`, …) is now visible via `extendedTypeName`. |

### Bugs fixed along the way

- **JSON object parameters failed on MariaDB**: `connection.execute(sql, [{...}])` always failed with `Incorrect arguments to mysqld_stmt_execute` because MariaDB rejects the `MYSQL_TYPE_JSON` parameter type. Objects are now sent as strings to MariaDB (MySQL keeps `MYSQL_TYPE_JSON`).
- **Buffer parameters corrupted MariaDB VECTOR inserts**: Buffers were sent as `MYSQL_TYPE_VAR_STRING`, letting the server treat the bytes as text in the connection charset: inserting a vector via `execute()` silently stored `[0,0,...]` (or errors under strict `sql_mode`). Buffers are now sent as `MYSQL_TYPE_BLOB`, the conventional type for binary parameters. `Execute.fromPacket` (server side, `stmt_execute` event) decodes them as Buffers.
- **`jsonStrings` was missing from the row parser cache key**: two connections differing only in `jsonStrings` could reuse each other's compiled row parsers. This also affected MySQL.
- **MariaDB NO PAD / UCA-14.0.0 collation ids crashed connections**: e.g. after `SET NAMES utf8mb4 COLLATE utf8mb4_nopad_bin`, reading any string column failed with `Encoding not recognized: 'undefined'` and killed the connection. The charset→encoding table now covers the NO PAD range (PAD SPACE id + 1024) and the uca1400 blocks (256-id blocks from 2048).
- **Timezone-dependent test**: the datetime test's "local TZ" assertions assumed the server session time zone matches the client's local zone (true on CI, false on non-UTC dev machines); the session time zone is now pinned, as the `Z`/`+09:30` cases in the same test already do.
- **`test-vector.test.mts` ran MySQL's `TO_VECTOR()` on MariaDB** (it only checked `major >= 9`); it now skips MariaDB, whose vectors are covered by the new integration test.

## ⚠️ Behavior changes to review

1. **MariaDB JSON columns now return parsed objects instead of strings** (consistent with MySQL; `jsonStrings: true` restores the previous behavior). This is the headline semver consideration.
2. **Buffer parameters are typed as `BLOB` instead of `VAR_STRING`** in `COM_STMT_EXECUTE`. No behavioral difference was observed on MySQL 8.4 across the full test suite, but it is technically observable server-side.

## Tests & docs

- New unit tests (run in CI without a MariaDB server): handshake extended-capability parsing, handshake response/SSL request byte layout, extended-metadata column definition parsing (including malformed/truncated blocks), and the collation-id encoding ranges.
- New integration test (`test-mariadb-extended-metadata.test.mts`) covering both protocols, metadata exposure, JSON object round-trips, vector Buffer round-trips, `typeCast`, and `jsonStrings`; skips (via poku `skip`) on non-MariaDB servers, so the current CI matrix is unaffected.
- New documentation page: **MariaDB Data Types** (`website/docs/documentation/mariadb-data-types.mdx`), linked from the docs index.
- Typings: `extendedTypeName`/`extendedFormat` added to `FieldPacket` and the `typeCast` field type.

## Test suite fixes (running the existing suite against MariaDB)

`getMysqlVersion()` in `test/common.test.mts` now also returns `version` (the raw server string) and `isMariaDB`, and the affected tests branch on it:

**MySQL-only SQL, replaced with portable or per-server equivalents:**

- `json-parse` / `json-string`: `CAST(... AS JSON)` is not valid MariaDB syntax; MariaDB uses `JSON_EXTRACT(..., '$')`, which also produces a JSON-typed result there.
- `test-invalid-date-result` / `test-#1019`: `performance_schema.session_variables` does not exist on MariaDB; both only fetched `sql_mode`, so they now use `SELECT @@sql_mode`.
- `test-backpressure-result-streaming`: MariaDB names the CTE depth limit `max_recursive_iterations` instead of `cte_max_recursion_depth`.

**Server behavior differences, expectations branched:**

- `test-date-parameter` / `test-custom-date-parameter`: MariaDB returns an integer for `UNIX_TIMESTAMP(<datetime param>)`; MySQL 8+ returns a DECIMAL string with microseconds.
- `test-multiple-results` / `test-binary-multiple-results`: MariaDB reports integer literals (`SELECT 1`) as `LONG`, MySQL as `LONGLONG`.
- `test-execute-nocolumndef`: MariaDB's `EXPLAIN` has no `partitions`/`filtered` columns and different field metadata; an explicit MariaDB expectation table is used (the regression itself, execute after a no-column-definition prepare, is exercised on both servers).
- `test-insert-json`: MariaDB enforces JSON validity via a CHECK constraint and reports errno 4025 (its `ER_CONSTRAINT_FAILED`) instead of MySQL's `ER_INVALID_JSON_TEXT` (3140). Only the errno is asserted on MariaDB because mysql2's error-name table follows MySQL, where 4025 has a different meaning. That name collision deserves its own discussion: mapping MariaDB error numbers (4000+) through the MySQL table produces misleading `code` values.
- `test-text-parser` (unit): skipped on MariaDB via poku `skip`; the MySQL-specific footgun it tests (JSON columns reported with the BINARY charset, issue #1661) does not exist on MariaDB, where JSON columns are utf8mb4 `LONGTEXT`.

## CI

`mariadb:11.8` (current LTS) is added to the `tests-linux` matrix alongside `mysql:8.3`. The existing docker-run step works unchanged: the mariadb image accepts `MYSQL_ALLOW_EMPTY_PASSWORD`/`MYSQL_DATABASE`, uses the same datadir, and the custom conf only sets `ssl-ca`/`ssl-cert`/`ssl-key`, which are the same variable names on MariaDB. Bun/Deno jobs stay MySQL-only for now to limit matrix growth; more MariaDB versions (e.g. `mariadb:10.11`) can be added later if wanted.

Two additional pre-existing test issues found along the way: the reset-connection tests built raw connections from environment variables but omitted `MYSQL_PORT`, silently connecting to the wrong server on non-default-port setups.

## Verification

Full local suite runs (for reference, current master fails 19 test files against MariaDB 11.8; this PR brings it to 0):

| Server | Options | Result |
| --- | --- | --- |
| MariaDB 11.8 | plain | 212 pass / 0 fail / 1 skip |
| MariaDB 11.8 | `MYSQL_USE_COMPRESSION=1` | pass |
| MariaDB 11.8 | `MYSQL_USE_TLS=1` | pass |
| MySQL 8.3 / 8.4 | plain | pass (no regressions) |

The 16 new `mariadb:11.8` CI jobs (Node 18/20/22/24 x SSL x compression) all passed on GitHub runners in the earlier #4374 run. Also exercised by hand against both servers: TLS connections (SSL request path), `disableEval` static parsers, prepared-statement parameter definitions, and `typeCast`.

One item left for a separate discussion: the MariaDB error-code namespace (MariaDB errno 4025, its `ER_CONSTRAINT_FAILED`, currently maps to `ER_INNODB_AUTOEXTEND_SIZE_OUT_OF_RANGE` via the MySQL error table).

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -29,12 +29,11 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [18, 20, 22, 24]
-        mysql-version: ['mysql:8.3']
+        mysql-version: ['mysql:8.3', 'mariadb:11.8']
         use-compression: [0, 1]
         use-tls: [0, 1]
         mysql_connection_url_key: ['']
         # static-parser: [0, 1]  # Already tested in "ci-coverage"
-        # TODO - add mariadb to the matrix. currently few tests are broken due to mariadb incompatibilities
 
     env:
       MYSQL_CONNECTION_URL: ${{ secrets[matrix.mysql_connection_url_key] }}
--- a/lib/base/connection.js
+++ b/lib/base/connection.js
@@ -87,6 +87,8 @@ class BaseConnection extends EventEmitter {
       },
     });
     this.serverCapabilityFlags = 0;
+    this._isMariaDB = false;
+    this._mariadbExtendedMetadata = false;
     this.authorized = false;
     this.sequenceId = 0;
     this.compressedSequenceId = 0;
--- a/lib/commands/client_handshake.js
+++ b/lib/commands/client_handshake.js
@@ -13,6 +13,7 @@
 const Command = require('./command.js');
 const Packets = require('../packets/index.js');
 const ClientConstants = require('../constants/client.js');
+const MariaDBClientConstants = require('../constants/mariadb_client.js');
 const CharsetToEncoding = require('../constants/charset_encodings.js');
 const auth41 = require('../auth_41.js');
 const { getAuthPlugin } = require('./auth_switch.js');
@@ -35,6 +36,7 @@ class ClientHandshake extends Command {
     super();
     this.handshake = null;
     this.clientFlags = clientFlags;
+    this.mariadbExtendedClientFlags = 0;
     this.authenticationFactor = 0;
   }
 
@@ -45,7 +47,8 @@ class ClientHandshake extends Command {
   sendSSLRequest(connection) {
     const sslRequest = new Packets.SSLRequest(
       this.clientFlags,
-      connection.config.charsetNumber
+      connection.config.charsetNumber,
+      this.mariadbExtendedClientFlags
     );
     connection.writePacket(sslRequest.toPacket());
   }
@@ -130,6 +133,7 @@ class ClientHandshake extends Command {
 
     const handshakeResponse = new Packets.HandshakeResponse({
       flags: this.clientFlags,
+      mariadbExtendedClientFlags: this.mariadbExtendedClientFlags,
       user: this.user,
       database: this.database,
       password: this.password,
@@ -265,6 +269,26 @@ class ClientHandshake extends Command {
     connection.serverCapabilityFlags = this.handshake.capabilityFlags;
     connection.serverEncoding = CharsetToEncoding[this.handshake.characterSet];
     connection.connectionId = this.handshake.connectionId;
+    // MariaDB needs a few protocol deviations, e.g. it rejects the JSON
+    // parameter type in COM_STMT_EXECUTE
+    connection._isMariaDB = /mariadb/i.test(this.handshake.serverVersion);
+    // If a MariaDB server offers extended column metadata, request it so
+    // MariaDB-specific column types (UUID, INET4, INET6, VECTOR and the JSON
+    // alias) can be identified. The server only reads the extended client
+    // capability bytes when the client leaves CLIENT_MYSQL (LONG_PASSWORD)
+    // unset, mirroring the server side of the negotiation.
+    if (
+      this.handshake.mariadbExtendedCapabilityFlags &
+      MariaDBClientConstants.MARIADB_CLIENT_EXTENDED_METADATA
+    ) {
+      this.clientFlags &= ~ClientConstants.LONG_PASSWORD;
+      this.mariadbExtendedClientFlags |=
+        MariaDBClientConstants.MARIADB_CLIENT_EXTENDED_METADATA;
+    }
+    connection._mariadbExtendedMetadata = Boolean(
+      this.mariadbExtendedClientFlags &
+      MariaDBClientConstants.MARIADB_CLIENT_EXTENDED_METADATA
+    );
     const serverSSLSupport =
       this.handshake.capabilityFlags & ClientConstants.SSL;
     // multi factor authentication is enabled with the
--- a/lib/commands/execute.js
+++ b/lib/commands/execute.js
@@ -51,7 +51,8 @@ class Execute extends Command {
       connection.config.charsetNumber,
       connection.config.timezone,
       this._executeOptions.attributes,
-      clientFlags
+      clientFlags,
+      connection._isMariaDB
     );
     //For reasons why this try-catch is here, please see
     // https://github.com/sidorares/node-mysql2/pull/689
@@ -77,7 +78,8 @@ class Execute extends Command {
     //  this.statement.columns[this._receivedFieldsCount] : new Packets.ColumnDefinition(packet);
     const field = new Packets.ColumnDefinition(
       packet,
-      connection.clientEncoding
+      connection.clientEncoding,
+      connection._mariadbExtendedMetadata
     );
     this._receivedFieldsCount++;
     this._fields[this._resultIndex].push(field);
--- a/lib/commands/prepare.js
+++ b/lib/commands/prepare.js
@@ -87,7 +87,11 @@ class Prepare extends Command {
       }
       return this.prepareDone(connection);
     }
-    const def = new Packets.ColumnDefinition(packet, connection.clientEncoding);
+    const def = new Packets.ColumnDefinition(
+      packet,
+      connection.clientEncoding,
+      connection._mariadbExtendedMetadata
+    );
     this.parameterDefinitions.push(def);
     if (this.parameterDefinitions.length === this.parameterCount) {
       return Prepare.prototype.parametersEOF;
@@ -99,7 +103,11 @@ class Prepare extends Command {
     if (packet.isEOF()) {
       return this.prepareDone(connection);
     }
-    const def = new Packets.ColumnDefinition(packet, connection.clientEncoding);
+    const def = new Packets.ColumnDefinition(
+      packet,
+      connection.clientEncoding,
+      connection._mariadbExtendedMetadata
+    );
     this.fields.push(def);
     if (this.fields.length === this.fieldCount) {
       return Prepare.prototype.fieldsEOF;
--- a/lib/commands/query.js
+++ b/lib/commands/query.js
@@ -199,7 +199,8 @@ class Query extends Command {
     if (this._fields[this._resultIndex].length !== this._fieldCount) {
       const field = new Packets.ColumnDefinition(
         packet,
-        connection.clientEncoding
+        connection.clientEncoding,
+        connection._mariadbExtendedMetadata
       );
       this._fields[this._resultIndex].push(field);
       if (connection.config.debug) {
--- a/lib/constants/charset_encodings.js
+++ b/lib/constants/charset_encodings.js
@@ -3,7 +3,7 @@
 // see tools/generate-charset-mapping.js
 // basicalliy result of "SHOW COLLATION" query
 
-module.exports = [
+const encodings = [
   'utf8',
   'big5',
   'latin2',
@@ -315,3 +315,27 @@ module.exports = [
   'utf8',
   'utf8',
 ];
+
+// MariaDB NO PAD collations mirror their PAD SPACE counterparts:
+// collation id = PAD SPACE collation id + 1024
+const padSpaceLength = encodings.length;
+for (let id = 1025; id < 1024 + padSpaceLength; id++) {
+  encodings[id] = encodings[id - 1024];
+}
+
+// MariaDB UCA-14.0.0 (uca1400) collations use fixed 256-id blocks
+// per character set, each covering the PAD SPACE and NO PAD variants
+const uca1400Blocks = [
+  [2048, 'cesu8'], // utf8mb3
+  [2304, 'utf8'], // utf8mb4
+  [2560, 'ucs2'], // ucs2
+  [2816, 'utf16'], // utf16
+  [3072, 'utf32'], // utf32
+];
+for (const [start, encoding] of uca1400Blocks) {
+  for (let id = start; id < start + 256; id++) {
+    encodings[id] = encoding;
+  }
+}
+
+module.exports = encodings;
--- a/lib/constants/mariadb_client.js
+++ b/lib/constants/mariadb_client.js
@@ -1,0 +1,19 @@
+'use strict';
+
+// MariaDB-specific extended capability flags.
+//
+// MariaDB servers (10.2+) do not set the CLIENT_MYSQL (LONG_PASSWORD)
+// capability and instead use 4 reserved bytes of the initial handshake packet
+// to advertise these flags. Clients answer with their own extended
+// capabilities in the last 4 reserved bytes of the handshake response
+// (the server only reads them when the client leaves CLIENT_MYSQL unset).
+//
+// The flags below are the lower 32 bits shifted down, i.e.
+// MARIADB_CLIENT_PROGRESS is documented as 1 << 32.
+// https://mariadb.com/docs/server/reference/clientserver-protocol/1-connecting/connection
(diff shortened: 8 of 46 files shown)


vlasky:
A note on the Benchmark workflow failures (attempts 1-4 on the previous head, `b3ac349`):

- The alert on `read 43 column definitions` (95,580 -> 76,907 ops/sec) looks like shared-runner variance: attempt 5 of the same commit measured back under the threshold, and a local A/B against master shows parity (108.8k vs 108.6k ops/sec on the same machine).
- The workflow run itself failed for an unrelated reason: `comment-on-alert: true` cannot post from fork PRs (read-only `GITHUB_TOKEN`, "Resource not accessible by integration"), so any alert raised on a fork PR errors the whole workflow even with `fail-on-alert: false`. That might be worth a tweak in `benchmark.yml`.

To remove the doubt anyway, the latest push hardens the hot path: the two new `ColumnDefinition` properties (`extendedTypeName`, `extendedFormat`) now live as defaults on the prototype, and instances only carry own values when a MariaDB server actually tags the column. Column definition objects from MySQL servers keep exactly the same shape and allocation as on master.

vlasky:
Follow-up after the second Benchmark alert (run 28756033523): careful alternating A/B runs on a quiet machine did show a small real cost (~4-6%) remaining in the `ColumnDefinition` constructor, separate from the runner noise. The cause was function size: the extended-metadata parse loop pushed the hot constructor past the inlining budget. The latest push moves that loop into a `_parseMariadbExtendedMetadata()` method (called only when the capability was negotiated), and three alternating rounds now measure at parity with master (branch 134.7k/134.7k/135.4k vs master 133.2k/134.7k/135.3k ops/sec on `read 43 column definitions`).

For context on the alert values themselves: the same benchmark has measured 76.9k, 95.6k, 114.9k and 93.4k ops/sec across recent runs of near-identical code, well beyond the 115% threshold on runner variance alone. There is also a ratchet effect in the comparison baseline: the cache's `restore-keys` prefix match returns the most recently created cache, and only successful runs save one, so only fast runs get to set the baseline (76.9k alerted against master's 95.6k and saved nothing; 114.9k passed and became the new baseline; 93.4k then alerted against 114.9k). That systematically biases the comparison against every subsequent average-speed runner, so fork PRs will keep flapping independently of the code, and each alert currently fails the workflow via the fork-token comment error mentioned above.

wellwelwel:
Thanks, @vlasky! Is there any related issue with a previous discussion for me to get familiar with? If not, that's fine 🙋🏻‍♂️

Can you explain why you separated the implementation into two _PRs_? Since the second one seems to test the implementations that are already here, I believe it would be easier to review in one place, given that the code seems to repeat in the core and can be confusing, besides having to keep them in sync for each change.

> [!NOTE]
> The current benchmark depends a lot on **GitHub Actions** having the same availability or equivalent to each past execution (`master`). In most cases, the failures are false positives 🙌🏻
> - The benchmark also can't create the comment in forks, which forces an error even when it would be just a warning.

vlasky:
Thanks for taking a look, @wellwelwel!

**Prior discussion:** there's no pre-existing issue for the feature as a whole; it came out of wanting first-class MariaDB 11.8 support (UUID, INET4/INET6, VECTOR and the JSON alias). It does resolve at least one long-standing issue directly: #1287 (JSON doesn't auto parse on MariaDB). With the extended metadata negotiated, MariaDB JSON columns and expressions such as `JSON_EXTRACT(...)` now parse to objects by default, and `jsonStrings: true` keeps the old behaviour. Happy to write up a summarising issue if that would help the review.

**The two PRs:** the split was feature (this PR) versus "make the existing test suite pass against MariaDB and add `mariadb:11.8` to the CI matrix" (#4374, resolving the TODO in `ci-linux.yml`). The second changes a dozen existing test files for pre-existing MySQL-only assumptions, and I didn't want that churn to obscure the protocol change. But you're right that a stacked pair across forks just shows duplicated commits and has to be kept in sync, so I've consolidated everything into this PR, updated the description to cover both parts, and closed #4374. For what it's worth, the 16 new `mariadb:11.8` jobs (Node 18/20/22/24 x SSL x compression) all passed on the earlier #4374 run.

And thanks for the note on the benchmark; that matches what we found (details in the comments above: runner variance plus the cache baseline ratchet, and the one real cost we did find was inlining-related and is fixed).

sidorares:
Thanks @vlasky ! The change looks good. I'll do a more deeper reading later tonight, but overall - happy to have this merged

sidorares (approved):

macno:
Hi there, thanks for your work, but this is a breaking change!
Why has it been released as minor? Who is using JSON column uses `JSON.parse()` on the returned value and now it throws an error!
I would suggest to release a quick patch where we switch the default value of `jsonStrings ` to `true`!

wellwelwel:
> Why has it been released as minor?

@macno, the automated tests didn't cover **MariaDB** due to several incompatibilities that this PR resolved, so the tests were enabled already in the "happy path". It wasn't intentional, and when users report cases like this in issues, we and the community look for a way to fix the regression whenever possible.

---

> I would suggest to release a quick patch where we switch the default value of `jsonStrings ` to `true`

In this case, forcing this option to `true` would be an even bigger breaking change. I believe the ideal, now that this has been reported, is to analyze the possibility of this only being applied to **MariaDB** until a major version 🤝

@vlasky, can you help us to check this approach viability? 🙋🏻‍♂️

vlasky:
Thanks @wellwelwel. I looked into the viability of making `jsonStrings` default to `true` for MariaDB connections only. Short answer: it is implementable, but I think it is undesirable, and I'd like to make the case before we commit to it.

**How it would work.** The only way to do this without adding new API is a tri-state on `jsonStrings`: `undefined` (option omitted) would mean strings on MariaDB but parsed objects on MySQL, while an explicit `false` would opt into parsing on MariaDB too. The config layer currently collapses the option to a boolean (`options.jsonStrings || false`), but preserving the tri-state is a small change, connection strings already deliver real booleans, and the parser cache key can carry the third state. So viability is not the issue.

**Why I think we shouldn't:**

1. **The pre-3.23.0 behaviour was a bug, not a contract.** The `jsonStrings` documentation has always said "Force JSON to be returned as string. (Default: false)", with no MySQL-only qualifier. The documented contract was "JSON values are parsed unless you opt out". MariaDB returning strings was not a designed default: the driver simply couldn't recognise the column as JSON because it didn't negotiate the extended metadata. 3.23.0 made the option honour its own documentation.

2. **Parse-by-default is what MariaDB itself ships.** The official `mariadb` connector has `autoJsonMap`, default `true`: JSON columns come back as parsed objects on MariaDB 10.5.2+. mysql2 now matches both its own documented contract and the official connector. Reverting would re-introduce a divergence that this feature existed to remove.

3. **Affected users already have a one-line, version-safe fix.** `jsonStrings: true` forces strings on both old and new mysql2 (on older versions it is a harmless no-op for MariaDB), so it can be deployed without pinning versions. This is the option doing exactly its documented job, which is different from a regression with no escape hatch.

4. **The tri-state is a footgun.** With it, passing the documented default value (`jsonStrings: false`) would behave differently from omitting the option. Any ORM or config wrapper that normalises options to explicit defaults would silently opt users back into parsing, defeating the revert in exactly the cases it is meant to protect.

5. **A revert now means breaking the same users twice.** Everyone who adapts to strings-by-default on MariaDB gets a second behaviour change at the next major when the default flips back. One clearly communicated change seems kinder than two.

**What I'd suggest instead:** keep the current default, and remediate through communication: a prominent callout in the release notes and changelog, a migration note at the top of the MariaDB data types docs page ("if you relied on JSON columns being strings, set `jsonStrings: true`"), and that same one-liner as the direct answer to reports like @macno's.

That said, this is ultimately a project stability call and I'll respect it either way. If you and the other maintainers still prefer the conservative path, I'm happy to implement the tri-state (or a clearly named temporary option) and send a PR.

sidorares:
I agree with point one, it was a bug that 3.23 fixed. It's a classical https://xkcd.com/1172/ . I think we should do what @vlasky suggest - accept that some users might be affected and help them migrate

Cameron-Asdf:
Well, this was a fun bit of mystery breakage to debug at work just now. Even just figuring out that it's because mysql2 (set at `^3.14.3`) had updated took a decent amount of time. Personally not sure how much I agree with calling this ok for a minor release.

nymkappa:
One might think it would at least be a worthy mention in the release note, but instead was buried inside a vague "_support MariaDB data types (UUID, INET4, INET6, VECTOR, JSON) via extended type metadata; run CI against MariaDB_"

> The pre-3.23.0 behaviour was a bug, not a contract

The code **is** the contract. A famous guy once said `DO NOT BREAK USERSPACE`.

Easy to patch on our end but come on...

wellwelwel:
I completely understand that, even though it's a bug, it's a bug that users have adapted to working around manually for years, however, saying something like "quick fix this!" or "come on" doesn't help.

I would like to emphasize that all work done here is voluntary, made by the community for the community, and that everyone is welcome to contribute.

I would also like to invite everyone who is liking the comment that tells us to _quick work_, to leave your likes on our contributions that have positively impacted you over the years.

As for what would help, think of something that actually leads somewhere, such as, just for example, "create a `supportMariaDB` flag until **MySQL2** version 4.x", because this gives direction and opens the possibility for the community to decide and vote on the option that makes the most sense.

Finally, I suggest opening a regression Issue, showing/discussing the problem vs. fix possibilities that don't remove this entire _PR_ until a major **MySQL2** version, but that also don't break the functionality for those who have adapted to this problem manually for years 🤝

PR #4374, opened 2026-07-05
test: make the test suite pass against MariaDB and add MariaDB 11.8 to the CI matrix

vlasky:
# Make the test suite pass against MariaDB and add MariaDB to the CI matrix

Stacked on #4373 (the first five commits are that PR; only the last two commits are new here). Together they resolve the `TODO - add mariadb to the matrix` in `ci-linux.yml`.

## Test suite fixes

`getMysqlVersion()` in `test/common.test.mts` now also returns `version` (the raw server string) and `isMariaDB`, and the affected tests branch on it:

**MySQL-only SQL, replaced with portable or per-server equivalents:**

- `json-parse` / `json-string`: `CAST(... AS JSON)` is not valid MariaDB syntax; MariaDB uses `JSON_EXTRACT(..., '$')`, which also produces a JSON-typed result there.
- `test-invalid-date-result` / `test-#1019`: `performance_schema.session_variables` does not exist on MariaDB; both only fetched `sql_mode`, so they now use `SELECT @@sql_mode`.
- `test-backpressure-result-streaming`: MariaDB names the CTE depth limit `max_recursive_iterations` instead of `cte_max_recursion_depth`.

**Server behavior differences, expectations branched:**

- `test-date-parameter` / `test-custom-date-parameter`: MariaDB returns an integer for `UNIX_TIMESTAMP(<datetime param>)`; MySQL 8+ returns a DECIMAL string with microseconds.
- `test-multiple-results` / `test-binary-multiple-results`: MariaDB reports integer literals (`SELECT 1`) as `LONG`, MySQL as `LONGLONG`.
- `test-execute-nocolumndef`: MariaDB's `EXPLAIN` has no `partitions`/`filtered` columns and different field metadata; an explicit MariaDB expectation table is used (the regression itself, execute after a no-column-definition prepare, is exercised on both servers).
- `test-insert-json`: MariaDB enforces JSON validity via a CHECK constraint and reports errno 4025 (its `ER_CONSTRAINT_FAILED`) instead of MySQL's `ER_INVALID_JSON_TEXT` (3140). Only the errno is asserted on MariaDB because mysql2's error-name table follows MySQL, where 4025 has a different meaning. That name collision deserves its own discussion: mapping MariaDB error numbers (4000+) through the MySQL table produces misleading `code` values.
- `test-text-parser` (unit): skipped on MariaDB via poku `skip`; the MySQL-specific footgun it tests (JSON columns reported with the BINARY charset, issue #1661) does not exist on MariaDB, where JSON columns are utf8mb4 `LONGTEXT`.

## CI

`mariadb:11.8` (current LTS) is added to the `tests-linux` matrix alongside `mysql:8.3`. The existing docker-run step works unchanged: the mariadb image accepts `MYSQL_ALLOW_EMPTY_PASSWORD`/`MYSQL_DATABASE`, uses the same datadir, and the custom conf only sets `ssl-ca`/`ssl-cert`/`ssl-key`, which are the same variable names on MariaDB. Bun/Deno jobs stay MySQL-only for now to limit matrix growth; more MariaDB versions (e.g. `mariadb:10.11`) can be added later if wanted.

## Verification (local, full suite)

| Server | Options | Result |
| --- | --- | --- |
| MariaDB 11.8 | plain | 210 pass / 0 fail / 1 skip |
| MariaDB 11.8 | `MYSQL_USE_COMPRESSION=1` | pass |
| MariaDB 11.8 | `MYSQL_USE_TLS=1` | pass |
| MySQL 8.4 | plain | pass (no regressions) |

For reference, current master fails 19 test files against MariaDB 11.8; #4373 brings that to 12, and this PR brings it to 0.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -29,12 +29,11 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [18, 20, 22, 24]
-        mysql-version: ['mysql:8.3']
+        mysql-version: ['mysql:8.3', 'mariadb:11.8']
         use-compression: [0, 1]
         use-tls: [0, 1]
         mysql_connection_url_key: ['']
         # static-parser: [0, 1]  # Already tested in "ci-coverage"
-        # TODO - add mariadb to the matrix. currently few tests are broken due to mariadb incompatibilities
 
     env:
       MYSQL_CONNECTION_URL: ${{ secrets[matrix.mysql_connection_url_key] }}
--- a/lib/base/connection.js
+++ b/lib/base/connection.js
@@ -87,6 +87,8 @@ class BaseConnection extends EventEmitter {
       },
     });
     this.serverCapabilityFlags = 0;
+    this._isMariaDB = false;
+    this._mariadbExtendedMetadata = false;
     this.authorized = false;
     this.sequenceId = 0;
     this.compressedSequenceId = 0;
--- a/lib/commands/client_handshake.js
+++ b/lib/commands/client_handshake.js
@@ -13,6 +13,7 @@
 const Command = require('./command.js');
 const Packets = require('../packets/index.js');
 const ClientConstants = require('../constants/client.js');
+const MariaDBClientConstants = require('../constants/mariadb_client.js');
 const CharsetToEncoding = require('../constants/charset_encodings.js');
 const auth41 = require('../auth_41.js');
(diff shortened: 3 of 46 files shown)


vlasky:
Consolidated into #4373 per review feedback: the feature branch now contains these commits directly, and the PR description there covers the test-suite and CI-matrix changes.

PR #4388, opened 2026-07-10
feat: return unsafe integers inside JSON columns as exact strings with supportBigNumbers

vlasky:
## Problem

JSON column values are parsed with plain `JSON.parse`, which silently rounds integers beyond `Number.MAX_SAFE_INTEGER` to the nearest double, with no error and no way to detect it after the fact:

```js
// column value in the database: {"id": 9007199254740993}
const [rows] = await conn.query('SELECT doc FROM t');
rows[0].doc.id; // 9007199254740992  (wrong, silently)
```

This affects MySQL `JSON` columns and, since #4373, MariaDB `JSON` columns identified via extended metadata. `BIGINT`/`DECIMAL` columns already have an opt-out from this class of corruption (`supportBigNumbers`), but values *inside* JSON documents had none short of `jsonStrings: true` (which gives up parsing entirely).

## Change

When the existing `supportBigNumbers` option is enabled, integers inside parsed JSON documents that cannot be accurately represented as JavaScript Numbers are now returned as exact String objects - mirroring the option's long-documented behaviour for `BIGINT`/`DECIMAL` columns:

```js
const conn = mysql.createConnection({ ..., supportBigNumbers: true });
rows[0].doc.id; // '9007199254740993'  (exact)
```

- Safe integers, floats (including unsafe-magnitude decimals and exponent notation), numeric strings and nulls are unaffected.
- **Default behaviour without the option is unchanged** (existing lossy `JSON.parse`).
- `jsonStrings: true` continues to return the raw (always exact) document text and takes precedence.

## Implementation

- A single `packet.parseJson(encoding, supportBigNumbers)` method now backs all eight JSON parse sites across the four row parsers (text/binary codegen + static variants), alongside the existing `parseVector()`/`parseGeometryValue()` helpers.
- Exactness uses `JSON.parse` source access ([proposal-json-parse-with-source](https://github.com/tc39/proposal-json-parse-with-source), Node.js 22+ / V8 12.2), feature-detected once at module load. Older runtimes keep plain `JSON.parse` behaviour, so nothing changes for them and `engines: node >= 8.0` is unaffected.
- Hot-path cost is one regex test: documents without a run of 16+ digits (the shortest numeral that can exceed `Number.MAX_SAFE_INTEGER`) take the plain `JSON.parse` path with no reviver.
- No parser-cache changes needed: `supportBigNumbers` is already part of the cache key.
- Docs updated in `ConnectionOptions` and `QueryOptions` typings.

## Tests

`test/unit/parsers/support-big-numbers-json.test.mts` covers all four parsers x both JSON flavours (MySQL `Types.JSON` and MariaDB `extendedFormat === 'json'`): unsafe positive/negative integers, safe integers, floats, unsafe-magnitude decimals, exponent notation, numerals inside strings, JSON `null`, bare root-level scalars, the option-off default, and `jsonStrings` precedence. The expectations self-adjust via the same feature detection, so CI lanes on Node 18/20 assert the (unchanged) legacy behaviour while 22/24 assert exactness.

Full suite passes locally (214 files) on Node 22 against MySQL 8.4 and MariaDB 11.8; also verified end-to-end over both text and binary protocols on both servers.

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -39,10 +39,8 @@ function leftPad(num, value) {
 const minus = '-'.charCodeAt(0);
 const plus = '+'.charCodeAt(0);
 
-// JSON.parse reviver source access (proposal-json-parse-with-source),
-// available from Node.js 22 / V8 12.2. Detected once; without it
-// parseJson falls back to plain JSON.parse, which rounds integers beyond
-// Number.MAX_SAFE_INTEGER to the nearest double.
+// JSON.parse source access: proposal-json-parse-with-source,
+// Node.js 22+ / V8 12.2
 const jsonSourceAccessSupported = (() => {
   let supported = false;
   JSON.parse('0', (key, value, context) => {
@@ -52,9 +50,7 @@ const jsonSourceAccessSupported = (() => {
   return supported;
 })();
 
-// The shortest numeral able to exceed Number.MAX_SAFE_INTEGER has 16
-// digits; documents without such a digit run parse exactly with plain
-// JSON.parse, keeping the reviver off the hot path
+// 16 digits is the shortest numeral that can exceed Number.MAX_SAFE_INTEGER
 const jsonBigNumeral = /\d{16}/;
 const jsonIntegerSource = /^-?\d+$/;
 
@@ -685,10 +681,8 @@ class Packet {
     return result;
   }
 
-  // Parses a JSON column value (MySQL JSON, or MariaDB JSON identified via
-  // extended metadata). With supportBigNumbers, integers that cannot be
-  // accurately represented as JavaScript Numbers are returned as exact
-  // String objects, mirroring the option's behaviour for BIGINT columns.
+  // With supportBigNumbers, unsafe integers become exact strings,
+  // mirroring the option's behaviour for BIGINT columns
   parseJson(encoding, supportBigNumbers) {
     const str = this.readLengthCodedString(encoding);
     if (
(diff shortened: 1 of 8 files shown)


lib/packets/packet.js
```diff
@@ -39,6 +39,37 @@ function leftPad(num, value) {
...
+  JSON.parse('0', (key, value, context) => {
+    supported = context !== undefined && typeof context.source === 'string';
+    return value;
+  });
+  return supported;
+})();
+
+// The shortest numeral able to exceed Number.MAX_SAFE_INTEGER has 16
+// digits; documents without such a digit run parse exactly with plain
+// JSON.parse, keeping the reviver off the hot path
```
wellwelwel: When comments are exactly what the logic already does next, I think they become unnecessary. The same goes for other comments that explain the implementation. What do you think about using comments only for things that the logic alone doesn't explain, or for the reasoning behind the decisions? 🙋🏻‍♂️

vlasky: Happy to prune the comments.

wellwelwel: Thanks! It looks like GitHub Actions is down (unrelated to the PR). I'm going to restart the workflow **(Done)**.

vlasky: Pruned in f3294cd5 - dropped everything that narrated the control flow, kept three short rationale notes (the TC39 proposal name/availability behind the feature detection, why the regex is 16 digits, and that parseJson reuses supportBigNumbers' existing BIGINT out-of-range contract rather than adding a new option).

PR #4394, opened 2026-07-13
fix(parser): call typeCast for NULL values in the binary protocol (#3368)

### LINKED ISSUE #3368: typeCast does not work for the `execute` method
I am unable to get the `typeCast` option to work at all for the `execute` method. I have tried configuring it both on the connection level and per query, but with no success.

I have set up a minimal reproduction example here: https://github.com/joakimbeng/mysql2-bug

The reproduction repo uses mysql2 v3.12.0 and node v22.

Am I missing something or shouldn't the `typeCast` option work with both the `query` and `execute` methods since v3.9.0?

Develop-KIM:
Closes #3368.

`execute()` never ran a user `typeCast` on NULL columns: the binary row parser set them to `null` before typeCast could see them, while `query()` (text parser) passes NULL columns through typeCast. So the same `typeCast` behaved differently between the two protocols.

This routes NULL columns through `typeCast` when it's a function, matching the text parser. The wrapper handed to `typeCast` for a NULL is null-safe — its `string()`/`buffer()`/`geometry()` return `null` without reading the packet, since a NULL value carries no bytes in the binary row. When `typeCast` isn't a function (or is `false`), NULL stays `null` as before.

Applied to both the eval (`binary_parser.js`) and `disableEval` (`static_binary_parser.js`) paths.

@wellwelwel pointed at the same spot in the issue — thanks for the root-cause note.

Added `test/integration/connection/test-typecast-null-execute.test.mts`: it fails on `master` (typeCast skips the NULL column) and passes with the fix, and it also checks that a column following a NULL still decodes correctly (no packet desync). Full suite is green in both eval and `STATIC_PARSER=1` modes.

### DIFF
--- a/lib/parsers/binary_parser.js
+++ b/lib/parsers/binary_parser.js
@@ -154,9 +154,7 @@ function compile(fields, options, config) {
     };
   }
 
-  // A NULL value carries no bytes in the binary row packet, so the accessors
-  // return null without reading from it. This mirrors the text parser, where
-  // NULL columns are still passed through a user-supplied typeCast (#3368).
+  /** A NULL value carries no bytes in the binary row packet, so the accessors return null without reading from it. */
   function wrapNull(field) {
     return {
       ...fieldMetadata(field),
--- a/lib/parsers/binary_parser.js
+++ b/lib/parsers/binary_parser.js
@@ -96,7 +96,7 @@ function compile(fields, options, config) {
   const parserFn = genFunc();
   const nullBitmapLength = Math.floor((fields.length + 7 + 2) / 8);
 
-  function wrap(field, packet) {
+  function fieldMetadata(field) {
     return {
       type: typeNames[field.columnType],
       extendedTypeName: field.extendedTypeName,
@@ -105,6 +105,12 @@ function compile(fields, options, config) {
       db: field.schema,
       table: field.table,
       name: field.name,
+    };
+  }
+
+  function wrap(field, packet) {
+    return {
+      ...fieldMetadata(field),
       string: function (encoding = field.encoding) {
         if (field.columnType === Types.JSON && encoding === field.encoding) {
           // Since for JSON columns mysql always returns charset 63 (BINARY),
@@ -144,6 +150,24 @@ function compile(fields, options, config) {
       },
       geometry: function () {
         return packet.parseGeometryValue();
+      },
+    };
+  }
+
+  // A NULL value carries no bytes in the binary row packet, so the accessors
+  // return null without reading from it. This mirrors the text parser, where
+  // NULL columns are still passed through a user-supplied typeCast (#3368).
+  function wrapNull(field) {
+    return {
+      ...fieldMetadata(field),
+      string: function () {
+        return null;
+      },
+      buffer: function () {
+        return null;
+      },
+      geometry: function () {
+        return null;
       },
     };
   }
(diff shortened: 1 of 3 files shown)


test/integration/connection/test-typecast-null-execute.test.mts
```diff
@@ -0,0 +1,62 @@
...
+
+// Regression for #3368: in the binary protocol (execute), NULL columns were
+// short-circuited to `null` before the user typeCast ran, so typeCast was
+// never called for them -- unlike the text protocol (query). These tests
+// assert typeCast now sees NULL columns and can override them, without
+// desyncing the packet reader for the surrounding columns.
+await describe('Typecast NULL Execute (#3368)', async () => {
+  const connection = createConnection();
+
+  try {
```
wellwelwel: There's no need to use `try-catch/finally` in a `describe` or `it` block in tests. The connection is already being declared in a scope above `it`, as is its ending, which ensures that the connection is closed even if the tests fail 🙋🏻‍♂️

test/integration/connection/test-typecast-null-execute.test.mts
```diff
@@ -0,0 +1,62 @@
...
+              }
+              return next();
+            },
+          },
+          (err, rows) => (err ? reject(err) : resolve(rows))
+        );
+      });
+
+      strict.deepEqual(seen, [null]); // typeCast ran for the NULL column
+      strict.equal(res[0].foo, '<was-null>'); // and its return value was used
```
wellwelwel: In testing, we can use messages to identify and visually highlight errors when debugging is necessary, rather than relying on comments, for example:

```suggestion
      strict.deepEqual(seen, [null], 'typeCast ran for the NULL column');
      strict.equal(res[0].foo, '<was-null>', 'and its return value was used');
```

- The same applies to the other comments accompanying the assertions 🙋🏻‍♂️

lib/parsers/static_binary_parser.js
```diff
@@ -94,6 +94,30 @@ function getBinaryParser(fields, _options, config) {
     }
   }
 
+  // A NULL value carries no bytes in the binary row packet, so the accessors
+  // return null without reading from it. This mirrors the text parser, where
+  // NULL columns are still passed through a user-supplied typeCast (#3368).
```
wellwelwel: The implementation and the function name are already self-explanatory, so we can remove the explanation of the implementation from the comment and keep only the reasoning behind the decision, for example:

```suggestion
  /** A NULL value carries no bytes in the binary row packet, so the accessors return null without reading from it. */
```

lib/parsers/binary_parser.js
```diff
@@ -148,6 +154,24 @@ function compile(fields, options, config) {
     };
   }
 
+  // A NULL value carries no bytes in the binary row packet, so the accessors
+  // return null without reading from it. This mirrors the text parser, where
+  // NULL columns are still passed through a user-supplied typeCast (#3368).
```
wellwelwel: Same ideia here:

```suggestion
  /** A NULL value carries no bytes in the binary row packet, so the accessors return null without reading from it. */
```

wellwelwel (changes requested):

wellwelwel:
Thanks @Develop-KIM! I left a few revisions. If you need any help, feel free to reach out 🙋🏻‍♂️

Develop-KIM:
Thanks! Took care of the rest: dropped the try/finally and moved the assertion notes into messages.

PR #4425, opened 2026-07-23
fix: don't leave a failed `execute` as the active command

### LINKED ISSUE #3293: undefined parameter passed to PreparedStatementInfo.execute cause rollback to get stuck
The following code gets stuck at `conn.rollback()`. If I change the undefined parameter to null(name is not nullable), it rolls back as expected. I know the parameters should not be undefined, but there is always a chance of accidents.
```
await conn.beginTransaction();
    try {
      const stmt = await conn.prepare("INSERT INTO foo(name)VALUES(?)");
      try {
        await stmt.execute([undefined]);
      } finally {
        await stmt.close();
      }
      console.log('commiting...');
      await conn.commit();
      console.log('committed.');
    } catch (e) {
      console.error(e);
      console.log('rolling back...');
      await conn.rollback();
      console.log('rolled back.');
    } finally {
      conn.release();
    }
```

Develop-KIM:
Fixes #3293.

### The bug
Inside a transaction, calling `statement.execute([undefined])` on a prepared statement wedges the connection — the following `rollback()`, and every later command, hangs forever.

### Cause
`Execute.start()` wraps `writePacket()` in a try/catch (added in #689 so a bad bind parameter doesn't kill the connection). When parameter encoding throws, it rejects the query via `onResult(error)` but still returns `Execute.prototype.resultsetHeader`. Nothing reached the socket, so no resultset will ever arrive — yet the command stays active waiting for one. It never ends, so `handlePacket` never shifts the next command off the queue and whatever is queued behind it (the rollback) never runs.

`Connection#execute` isn't affected because it checks for `undefined` up front and throws before a command is created. The raw `PreparedStatementInfo.execute` path has no such guard, so it reaches this code.

### Fix
Return `null` from the catch path so the command ends (emits `end`, and `handlePacket` moves on to the queue). This matches how a server error packet is already handled in `Command#execute` — `onResult(err)` followed by ending the command.

### Verification
- Added a regression test in `test-bind-undefined` that reproduces the hang: it deadlocks on the current code and passes with the fix. Also checked the connection is fully usable afterwards — a later query and re-executing the same statement with a valid parameter both succeed.
- Green in both default and `STATIC_PARSER=1` modes; prepare / transaction / reset / promise-wrapper and the existing regression tests still pass.

AI-assisted, disclosed in the commit trailer; I've reviewed the change and can walk through it.

### DIFF
--- a/lib/commands/execute.js
+++ b/lib/commands/execute.js
@@ -63,10 +63,6 @@ class Execute extends Command {
     try {
       connection.writePacket(executePacket.toPacket(1));
     } catch (error) {
-      // Nothing reached the wire, so no resultset will ever arrive. End the
-      // command here instead of advancing to resultsetHeader, otherwise it stays
-      // the active command forever and blocks the queue (e.g. a following
-      // rollback never runs). See #3293.
       this.onResult(error);
       return null;
     }
--- a/lib/commands/execute.js
+++ b/lib/commands/execute.js
@@ -63,7 +63,12 @@ class Execute extends Command {
     try {
       connection.writePacket(executePacket.toPacket(1));
     } catch (error) {
+      // Nothing reached the wire, so no resultset will ever arrive. End the
+      // command here instead of advancing to resultsetHeader, otherwise it stays
+      // the active command forever and blocks the queue (e.g. a following
+      // rollback never runs). See #3293.
       this.onResult(error);
+      return null;
     }
     return Execute.prototype.resultsetHeader;
   }
--- a/lib/commands/execute.js
+++ b/lib/commands/execute.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const Timers = require('timers');
+
 const Command = require('./command.js');
 const Query = require('./query.js');
 const Packets = require('../packets/index.js');
@@ -63,7 +65,15 @@ class Execute extends Command {
     try {
       connection.writePacket(executePacket.toPacket(1));
     } catch (error) {
-      this.onResult(error);
+      if (this.queryTimeout) {
+        Timers.clearTimeout(this.queryTimeout);
+        this.queryTimeout = null;
+      }
+      if (this.onResult) {
+        this.onResult(error);
+      } else {
+        this.emit('error', error);
+      }
       return null;
     }
     return Execute.prototype.resultsetHeader;
--- a/test/integration/connection/test-bind-undefined.test.mts
+++ b/test/integration/connection/test-bind-undefined.test.mts
@@ -36,14 +36,10 @@ await describe('Bind Undefined', async () => {
       'INSERT INTO test_bind_undefined (name) VALUES (?)'
     );
 
-    await strict.rejects(
-      // @ts-expect-error: testing that undefined bind parameter throws TypeError
-      statement.execute([undefined]),
-      {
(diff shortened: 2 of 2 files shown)


test/integration/connection/test-bind-undefined.test.mts
```diff
@@ -26,6 +26,34 @@ await describe('Bind Undefined', async () => {
...
+    );
+
+    await strict.rejects(
+      // @ts-expect-error: testing that undefined bind parameter throws TypeError
+      statement.execute([undefined]),
+      {
+        name: 'TypeError',
+        message: 'Bind parameters must not contain undefined',
+      }
+    );
```
wellwelwel:
```suggestion
    await strict.rejects(statement.execute([undefined]), {
      name: 'TypeError',
      message: 'Bind parameters must not contain undefined',
    });
```

lib/commands/execute.js
```diff
@@ -63,7 +63,12 @@ class Execute extends Command {
     try {
       connection.writePacket(executePacket.toPacket(1));
     } catch (error) {
+      // Nothing reached the wire, so no resultset will ever arrive. End the
+      // command here instead of advancing to resultsetHeader, otherwise it stays
+      // the active command forever and blocks the queue (e.g. a following
+      // rollback never runs). See #3293.
       this.onResult(error);
+      return null;
     }
```
wellwelwel, line 78: There's no need to explain the implementation in the comments. As for the regression, it's already in the tests 🙋🏻‍♂️


```suggestion
    } catch (error) {
      this.onResult(error);
      return null;
    }
```

test/integration/connection/test-bind-undefined.test.mts
```diff
@@ -26,6 +26,30 @@ await describe('Bind Undefined', async () => {
...
+      'INSERT INTO test_bind_undefined (name) VALUES (?)'
+    );
+
+    await strict.rejects(statement.execute([undefined]), {
+      name: 'TypeError',
+      message: 'Bind parameters must not contain undefined',
+    });
+
+    // Regression #3293: the failed execute used to stay the active command and
+    // block the queue, so this rollback — and every command after it — hung forever.
```
wellwelwel:
```suggestion
    // Regression #3293: the failed `execute` used to stay the active command blocks the queue indefinitely.
```

wellwelwel:
Thanks, @Develop-KIM! I'll check if **Copilot**'s recommendations make sense instead of just returning `null`. Feel free to take a look, too 🙋🏻‍♂️

Develop-KIM:
Took a look at Copilot's two notes and both hold up against how `Command#execute` already handles the server-error path, so I pushed a small follow-up:

- Clear `queryTimeout` in the catch. If someone passes a `timeout` option, `_setTimeout()` has already scheduled it by the time `writePacket` throws; ending the command with `return null` left that timer to fire later on a command that's already done.
- Guard `onResult` and fall back to `emit('error')` when there's no callback, same as the error-packet path — the raw prepared-statement/event API can reach here without one.

I also hardened the test: it now always closes the statement, and asserts the statement is still usable with a successful `execute` after the failed one. Kept your minimal `return null` underneath. Green in both eval and static-parser modes.

PR #4458, opened 2026-08-06
fix: keep connectTimeout active until the handshake completes

raprav:
## Summary

Keep `connectTimeout` active until the client handshake completes, instead of clearing it when the first server bytes arrive.

## Problem

The timer is currently cleared when the server sends its greeting. If the server then stops responding without closing the socket, the connection can remain pending indefinitely.

```text
Current behavior:

connectTimeout armed
        │
        ├── server greeting ──> connectTimeout cleared
        ├── client sends HandshakeResponse
        └── server stalls
              ├── connection timeout: no longer active
              ├── query timeout: not armed until the handshake completes
              └── connect() / pool.getConnection(): remain pending


With this change:

connectTimeout armed
        │
        ├── server greeting ──> still armed
        ├── client sends HandshakeResponse
        ├── handshake completes ──> timer cleared
        └── server stalls ────────> ETIMEDOUT
```

We observed this against RDS, where a connection attempt remained pending until the process was restarted.

## Change

For client connections, clear `connectTimeout` when `ClientHandshake` emits `end`, so it covers the complete connection establishment process.

Existing behavior remains unchanged for:

- error paths, which already clear the timer in `_notifyError()`;
- server-side connections, which do not run `ClientHandshake` and continue clearing the timer on first activity.

## Tests

The regression test uses a plain TCP server and MySQL2's `Handshake` packet. The server sends a valid greeting, receives the client's `HandshakeResponse`, and then remains silent without closing the socket.

It verifies that:

- the timer remains armed while the handshake is pending;
- the timeout path reports a fatal `ETIMEDOUT`;
- the timer is cleared after a successful handshake.

The timeout handler is invoked directly, avoiding wall-clock-dependent assertions.

Validated with the full test suite, lint, type checking, build checks, and the targeted test under Node.js, Bun, and Deno.

## Behavioral note

With this change, a handshake that legitimately takes longer than `connectTimeout` will fail with `ETIMEDOUT`. This is intentional and matches the documented behavior: `connectTimeout` limits the time allowed to establish the initial MySQL connection, which is not complete until the handshake finishes. Applications that require a longer handshake window can increase the configured value.


wellwelwel (approved):
Thanks, @raprav! 🙋🏻‍♂️

PR #4460, opened 2026-08-08
fix: honor query-level namedPlaceholders:false

### LINKED ISSUE #2474: namedPlaceholders cannot be disabled at query level
- As reported in #2472

When setting `namedPlaceholders` globally (connection level) to `true`, it's not possible to disable it at the query level.

---

To reproduce, just uncomment it from this current test:

https://github.com/sidorares/node-mysql2/blob/41d21ebc9bcb47df9fc84f1f37da357fd77ba5a7/test/esm/integration/named-placeholders.test.mjs#L18-L29

---

The opposite will work fine:

https://github.com/sidorares/node-mysql2/blob/41d21ebc9bcb47df9fc84f1f37da357fd77ba5a7/test/esm/integration/named-placeholders.test.mjs#L31-L44

Hashim1999164:
## Summary
- Prefer query-level `namedPlaceholders` over connection config in `_resolveNamedPlaceholders` (same pattern as pool), so `false` can disable conversion when the connection has it enabled.
- Pass the resolved `namedPlaceholders` flag into `format()` so a second resolve cannot re-apply connection config after `query()` already decided.
- Re-enable the previously disabled integration coverage and add a unit test for resolve/format override behavior. `execute` with object values and `namedPlaceholders: false` correctly throws TypeError about bind parameters needing an array.

Fixes #2474

## Test plan
- [x] Unit: `test/unit/connection/test-named-placeholders-resolve.test.mts` (config true + options false, config false + options true, undefined falls back to config, format third arg)
- [x] Manual assert covering the query → format double-resolve path
- [ ] CI integration: `test/integration/named-placeholders.test.mts` (local Docker MySQL not pulled due to disk; CI should run these)


Hashim1999164:
Fixed a regression: `Query` was defaulting `namedPlaceholders` to `false`, which blocked connection-level config from enabling conversion. Unset query options now stay undefined so config can apply, while explicit `namedPlaceholders: false` still wins.

lib/base/connection.js
```diff
@@ -596,6 +596,11 @@ class BaseConnection extends EventEmitter {
       sql: sql,
       values: values,
     };
+    // Keep query-level namedPlaceholders through formatting so connection
+    // config cannot re-enable conversion after query() already resolved.
```
wellwelwel:
```suggestion
```

lib/base/connection.js
```diff
@@ -619,7 +624,11 @@ class BaseConnection extends EventEmitter {
 
   _resolveNamedPlaceholders(options) {
     let unnamed;
-    if (this.config.namedPlaceholders || options.namedPlaceholders) {
+    // Prefer query-level namedPlaceholders when set (same as pool.query).
```
wellwelwel:
```suggestion
```

lib/commands/query.js
```diff
@@ -20,7 +20,13 @@ class Query extends Command {
     this.sql = options.sql;
     this.values = options.values;
     this._queryOptions = options;
-    this.namedPlaceholders = options.namedPlaceholders || false;
+    // Keep undefined when unset so connection config can apply.
```
wellwelwel:
```suggestion
```

test/integration/named-placeholders.test.mts
```diff
@@ -43,21 +46,27 @@ await describe('Test namedPlaceholder as command parameter in connection', async
...
-  //   }
-  // });
+  await it(async () => {
+    const c = createConnection({ namedPlaceholders: true }).promise();
+
+    try {
+      await c.execute({ sql: query, namedPlaceholders: false }, values);
+      strict.fail('Enabled in connection config, disabled in execute command');
+    } catch (err: unknown) {
+      // With namedPlaceholders disabled, object bind params are rejected.
```
wellwelwel:
```suggestion
```

wellwelwel (changes requested):
Thanks, @Hashim1999164! There's no need to "re"explain the implementation in the comments 🙋🏻‍♂️

test/integration/named-placeholders.test.mts
```diff
@@ -11,21 +10,25 @@ await describe('Test namedPlaceholder as command parameter in connection', async
...
+          ? String((err as { sqlMessage?: unknown }).sqlMessage)
+          : '';
+      strict(
+        sqlMessage.match(/right syntax to use near ':named'/),
+        'Enabled in connection config, disabled in query command'
+      );
+    } finally {
+      await c.end();
+    }
+  });
```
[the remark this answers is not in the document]
wellwelwel: @Hashim1999164, can you apply this fix? 🙋🏻‍♂️

PR #4473, opened 2026-08-09
fix(pool): give each pooled connection its own config copy

### LINKED ISSUE #477: changeUser affects pool configuration!!!
Hi there,

this is annoying me in  mysql2@1.1.2:

    var pool=require('mysql2').createPool({
      connectionLimit : 5, 
      host            : 'localhost',
      user            : 'root',
      //password      : ... 
      database       : 'mysql',
      multipleStatements: true
    });
    
    // get pool connection:
    pool.getConnection((err,con) => {
      if(err) return console.log(err);
      con.query('select user();', (err, res) => {
        if(err) return console.log(err);
        console.log(res); // --> root@localhost
    
        // change connection user:
        con.changeUser({
          user: 'test',
          password: 'test'
        }, err => {
          if(err) return console.log(err);
    
          con.query('select user();', (err, res) => {
            if(err) return console.log(err);
            console.log(res); // --> test@localhost
    
            // destroy connection
            con.destroy();
    
            // get a fresh connection
            pool.getConnection((err,con) => {
              if(err) return console.log(err);
              con.query('select user();', (err, res) => {
                if(err) return console.log(err);
                console.log(res); // --> STILL test@localhos !!!!
                pool.getConnection((err,con) => {
                  if(err) return console.log(err);
    
                  // get a second connection, maybe destroy didn't work...?
                  con.query('select user();', (err, res) => {
                    if(err) return console.log(err);
                    console.log(res); // --> test@localhost !!!!!!!
                  });
                });
              });
            });
    
          });
        });
      });
    });

Take care

M

bilashcse:
Fixes #477

## Problem

`BasePool#getConnection()` handed every `PoolConnection` the **same** config object:

```js
connection = new PoolConnection(this, {
  config: this.config.connectionConfig,
});
```

and `ChangeUser` rewrites `connection.config` in place (`lib/commands/change_user.js`):

```js
this.currentConfig.user = this.user;
this.currentConfig.password = this.password;
this.currentConfig.database = this.database;
this.currentConfig.charsetNumber = this.charsetNumber;
```

`currentConfig` is `this.config` from `Connection#changeUser()`, which for a pooled connection *is* `pool.config.connectionConfig`. So a single `changeUser()` call permanently rewrote the pool credentials and every connection the pool opened afterwards logged in as the new user - exactly what was reported in #477. `mysqljs/mysql` does not behave this way.

## Fix

Each connection now gets its own copy of the pool connection config. The prototype and property descriptors are preserved, so the copy is still a `ConnectionConfig`, and the pool keeps its own config as the single source of truth, meaning connections created later still pick up changes made to the pool config itself.

## Changes

- `lib/base/pool.js`: new `_createConnectionConfig()` helper, used when constructing a `PoolConnection`.
- `test/unit/pool/test-pool-connection-config-isolation.test.mts`: unit test (no server needed) covering that each connection gets a distinct config, that values and the prototype survive the copy, and that a `changeUser`-style mutation does not leak to the pool config or to later connections.
- `test/global/integration/pool/test-pool-change-user-isolation.test.mts`: end-to-end regression test - `changeUser()` on a pooled connection, destroy it, then assert a freshly created connection is back to the pool user.
- `website/docs/index.mdx`: short note in the "Using Connection Pools" section documenting that `changeUser()` only affects the connection it is called on.

## Notes

- No public API change; `_createConnectionConfig()` is internal.
- Nothing in the codebase relied on pooled connections sharing the config object identity.
- The global test is auto-discovered by `poku.config.js` and is skipped when the test user lacks privileges.

### DIFF
--- a/test/global/integration/pool/test-pool-change-user-isolation.test.mts
+++ b/test/global/integration/pool/test-pool-change-user-isolation.test.mts
@@ -61,4 +61,3 @@ await describe('Pool: changeUser() isolation', async () => {
   await admin.query("DROP USER IF EXISTS 'pooluser1'@'%'");
   await admin.end();
 });
-
--- a/test/global/integration/pool/test-pool-change-user-isolation.test.mts
+++ b/test/global/integration/pool/test-pool-change-user-isolation.test.mts
@@ -1,0 +1,64 @@
+import type { PoolConnection, RowDataPacket } from '../../../../index.js';
+import process from 'node:process';
+import { describe, it, skip, strict } from 'poku';
+import { createPool } from '../../../common.test.mjs';
+
+if (`${process.env.MYSQL_CONNECTION_URL}`.includes('pscale_pw_')) {
+  skip('Skipping test for PlanetScale');
+}
+
+await describe('Pool: changeUser() isolation', async () => {
+  const pool = createPool({ connectionLimit: 1 });
+
+  const onlyUsername = (name: string) => name.substring(0, name.indexOf('@'));
+
+  const getConnection = () =>
+    new Promise<PoolConnection>((resolve, reject) => {
+      pool.getConnection((err, connection) =>
+        err ? reject(err) : resolve(connection)
+      );
+    });
+
+  const currentUser = (connection: PoolConnection) =>
+    new Promise<string>((resolve, reject) => {
+      connection.query<RowDataPacket[]>(
+        'SELECT CURRENT_USER() AS `user`',
+        (err, rows) => (err ? reject(err) : resolve(onlyUsername(rows[0].user)))
+      );
+    });
+
+  const admin = pool.promise();
+
+  await admin.query(
+    "CREATE USER IF NOT EXISTS 'pooluser1'@'%' IDENTIFIED BY 'pooluser1pass'"
+  );
+  await admin.query("GRANT SELECT ON *.* TO 'pooluser1'@'%'");
+  await admin.query('FLUSH PRIVILEGES');
+
+  await it('keeps the pool user for connections created later', async () => {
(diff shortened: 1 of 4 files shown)


test/global/integration/pool/test-pool-change-user-isolation.test.mts
```diff
@@ -0,0 +1,63 @@
+import type { PoolConnection, RowDataPacket } from '../../../../index.js';
+import process from 'node:process';
+import { describe, it, skip, strict } from 'poku';
+import { createPool } from '../../../common.test.mjs';
+
+if (`${process.env.MYSQL_CONNECTION_URL}`.includes('pscale_pw_')) {
+  skip('Skipping test for PlanetScale');
+}
+
+await describe('Pool: changeUser() isolation', async () => {
```
wellwelwel, line 5:
```suggestion
import type { PoolConnection, RowDataPacket } from '../../../../index.js';
import { describe, it, skip, strict } from 'poku';
import { createPool } from '../../../common.test.mjs';

await describe('Pool: changeUser() isolation', async () => {
```

test/global/integration/pool/test-pool-change-user-isolation.test.mts
```diff
@@ -0,0 +1,58 @@
+import type { PoolConnection, RowDataPacket } from '../../../../index.js';
+import { describe, it, skip, strict } from 'poku';
```
wellwelwel:
```suggestion
import { describe, it, strict } from 'poku';
```

wellwelwel:
Thanks, @bilashcse 🙋🏻‍♂️✨

PR #4476, opened 2026-08-10
ci: add tests to handshake error greeting

### LINKED ISSUE #606: Concurrent call to mysql breaks after some interaction
I have created an singleton connection to use the same connection_id() so the rest app can make use of TEMPORARY TABLES. After some concurrent calls, the connection is broken with this error:

```
node_modules/mysql2/lib/connection.js:127
connection.threadId = handshakeCommand.handshake.connectionId;
TypeError: Cannot read property 'connectionId' of null
at ClientHandshake.<anonymous>
```

He is the code

```node

import { Me } from '../shared/Me';
import * as knex from 'knex';

export class BucketDb {

    private static _connection: { [sessionId: string]: knex } = {};

    public static connect(connectionData) {

        if (!this._connection || (this._connection && !this._connection[Me.profile().sessionId])) {
            this._connection[Me.profile().sessionId] = knex({
                client: 'mysql2',
                connection: connectionData,
                pool: {
                    min: 0, max: 1
                }
            });
        }

        // this._connection[Me.profile().sessionId].raw('select connection_id()').then((r) => {
        //     console.log(Me.profile().sessionId, r[0][0]);
        // });

        return this._connection[Me.profile().sessionId];

    }

}

```

wellwelwel:
Closes #606 (already fixed in #607).

### DIFF
--- a/test/unit/connection/test-error-packet-instead-of-handshake.test.mts
+++ b/test/unit/connection/test-error-packet-instead-of-handshake.test.mts
@@ -1,0 +1,61 @@
+import { Buffer } from 'node:buffer';
+import { createServer } from 'node:net';
+import { describe, it, skip, strict } from 'poku';
+import mysql from '../../../index.js';
+
+if (typeof Deno !== 'undefined') skip('Deno: process is not terminated');
+
+const ER_CON_COUNT_ERROR = 1040;
+
+const errorGreeting = (): Buffer => {
+  const message = Buffer.from('Too many connections', 'latin1');
+  const payload = Buffer.concat([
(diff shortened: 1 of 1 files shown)


sidorares:
lol that was 9 years ago :) time flies

wellwelwel:
> lol that was 9 years ago :) time flies

Yes! 😂

I'm going through several old issues that I suspect have already been fixed, but were never closed due to lack of tests to prove it.

- There are still a few more. I'll go through them one by one to be sure 🙋🏻‍♂️

PR #4478, opened 2026-08-10
fix(typings): align callback Pool and PoolConnection types with runtime

### LINKED ISSUE #3964: Some types are incorrect on mysql2 (callback based) but not mysql2/promise
I was writing some wrappers to abstract away the difference for some conversion code and I noticed that the runtime mysql2 and mysql2/promise don't match the types that are shipped.

Namely:

1. https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/Pool.d.ts is not set up to extend the connection like the PromisePool does, which means it does not inherit escape/escapeId/format, but those functions do exist at runtime, seen here: https://github.com/sidorares/node-mysql2/blob/master/lib/base/pool.js#L224
When I set a debugger and instantiated a callback based pool and called escape on the object, it got to that code, but I had to write it as `pool['escape']('1')` and turn off TS warnings because TS didn't believe the function existed.  
2. All that said, the `PromisePool` extends the Connection class, so it has all the members the connection has.  This fixes the above, though it introduces a slight bug, as Connections have a `threadId` and neither a `mysql2/Pool` or a `mysql2/Promise/Pool` have a `threadId` on them.
3. The callback based `PoolConnection.promise()` function is typed to return a Pool, but it actually returns a promise wrapped `mysql2/promise/PoolConnection`.  You can see the type at https://github.com/sidorares/node-mysql2/blob/master/typings/mysql/lib/PoolConnection.d.ts#L7 but the runtime code at https://github.com/sidorares/node-mysql2/blob/master/lib/pool_connection.js#L8

Why does any of this matter you might ask?  Well for most people it might not, definitely if they're raw JS using the types as hints.  But if they're converting from mysql to mysql2 without going to `/promise`, the missing functions could throw them for a loop.  For me, I'm torn because I'm making some wrapper classes that are providing a consistent interface between the two drivers to share between my projects as a utility so they can be used by projects in different states of conversion.  The issue is that I have to deal with the fact that there are two types to deal with: 1. the runtime truth, which I have to declare to make TS happy when I'm doing operations, and 2. a type that is compatible with the mysql2 shipped types or when people pass in constructed types, they'll be shown as incompatible.

bilashcse:
Closes #3964

### Problem

The callback API typings drifted from the runtime implementation in two places:

1. `PoolConnection.promise()` was declared as returning a promise-based `Pool`, but `lib/pool_connection.js` returns a `PromisePoolConnection`. Consumers lost `release()` and were offered `getConnection()` instead.
2. The callback `Pool` had no `escape`, `escapeId` or `format`, even though `lib/base/pool.js` implements all three. The promise-based `Pool` already inherits them from `Connection`, so the two APIs disagreed and users had to fall back to `pool['escape'](...)` with the checker silenced.

### Changes

- `typings/mysql/lib/PoolConnection.d.ts` — `promise()` now returns `PoolConnection` from `mysql2/promise`.
- `typings/mysql/lib/Pool.d.ts` — adds `escape`, both `escapeId` overloads and `format`, reusing the signatures already declared in `Connection.d.ts`.
- `test/tsc-build/mysql/createPool/callbacks/escape.test.ts` — compile-time coverage for the escaping helpers.
- `test/tsc-build/mysql/createPool/callbacks/promise.test.ts` — compile-time coverage for `.promise()`, with a `@ts-expect-error` on `getConnection()` so the return type cannot regress back to `Pool`.
- `website/docs/examples/connections/create-pool.mdx` — documents both behaviours.

Typings, tests and docs only; there is no runtime change.


wellwelwel (approved):

wellwelwel:
Thanks again, @bilashcse ✨

PR #4481, opened 2026-08-12
fix: keep leading zeros in TIME fractional seconds

NotAFlightRisk:
Noticed this reading durations back out of a `TIME(3)` column. Through `execute()` any non-zero fraction below 0.1s comes back too large, because the microsecond field goes into the string without being padded to six digits first - so 50 ms reads as 500 ms, and 1 ms as 100 ms.

Same rows, same columns, `query()` vs `execute()` on MySQL 9.7.2:

```
stored             query()            execute()
01:02:03.050       01:02:03.050       01:02:03.5
01:02:03.001       01:02:03.001       01:02:03.1
01:02:03.000123    01:02:03.000123    01:02:03.123
```

`readDateTimeString` already does `leftPad(6, ms)` on the same field a few lines up, so this just brings TIME into line with it. The trailing-zero trim is untouched, `-01:23:44.88` still comes out the same.

New test vectors are real packets off the wire, like the ones already in there.


sidorares:
LGTM
one thing might be worth adding is integration test: real server, insert into temp table then read and assert expected value.

NotAFlightRisk:
Makes sense, added! It fails without the fix, `01:02:03.5` instead of `01:02:03.05`

wellwelwel (approved):

NotAFlightRisk:
Cheers @sidorares, @wellwelwel :) the integration test was worth asking for - a lot more convincing than my unit test on its own.

sidorares:
thanks @NotAFlightRisk - published as v3.23.4

PR #4486, opened 2026-08-15
perf: remove per-query and per-row allocation hotspots

sidorares:
## Summary

Ten independent performance fixes (one commit each) found by profiling the client against Docker MySQL 8.3/9 across typical workloads: small command/response loops, 1–100 row queries, 10k–1M row scans, 10/100-column rows, and date-heavy results. The last commit adds the benchmark harness (`benchmarks/perf/`) and the full write-up (`benchmarks/perf/ANALYSIS.md`) with methodology, baseline profiles, and negative results.

All changes are pure JS with feature-detected fallbacks — no native modules. Full suite (226 files), lint, and typecheck pass; each commit parses standalone for bisectability.

## Highlights

- **`Query.toPacket` serialized every query twice** on MySQL 8+ (query-attributes capability): a dry run against `Packet.MockBuffer()`, which itself rebuilt a patched Buffer per call. Now single-pass when no attributes are set.
- **The binary row parser allocated typeCast wrapper objects (metadata + 3 closures) per column per row even with no `typeCast` configured** — this made `execute()` slower than `query()` on wide rows.
- **Text DATE/DATETIME went through `new Date(string)`** (V8 slow path, ~62% of date-heavy scans); now parsed as digits straight from the buffer, value-identical, with string-path fallback for non-standard shapes.
- Smaller fixes: lazy column-def caching without `Object.defineProperty`, parser-cache key without per-field `JSON.stringify`, direct `utf8Slice`/`latin1Slice` decode, cached resultset arrays in the row path, `Long`-free int64 reads, and the promise wrapper now honors `trace: false` for its per-query stack capture.

## Measured (isolated wire replay = client CPU only; e2e = Docker MySQL 8.3)

| Scenario | Baseline | This branch | Δ |
|---|---|---|---|
| `SELECT 1` loop (replay) | 40.7k ops/s | 250k ops/s | **6.1×** |
| 1 row × 10 cols (replay) | 28.3k | 96.9k | **3.4×** |
| 100 rows × 10 cols (replay) | 12.9k | 22.1k | +71% |
| 100k date rows (replay) | 10.2 | 14.3 | +40% |
| 1M rows × 3 cols (replay) | 4.57 | 5.30 (5.3M rows/s) | +16% |
| 100×100 `execute()` (e2e) | 600 ops/s | 1 087 | **+81%** |
| 100×10 query / execute (e2e) | 2 178 / 2 126 | 2 755 / 2 795 | +26% / +31% |
| insert loop (e2e) | 16% client CPU | ~9% client CPU | latency stays fsync-bound |

## ⚠️ One intentional contract change

`perf: reuse a single Packet instance for complete packets` changes the internal `PacketParser.onPacket` contract: packets are delivered through one mutable instance, so a consumer retaining a packet past the synchronous callback must `clone()` it. The only such place in the client (paused-connection queue) now clones, and the packet-parser unit test was updated accordingly. If external code constructs `PacketParser` directly and retains packets, this is the one commit to scrutinize — happy to drop it or put it behind an option; it's worth ~5% on large scans plus GC relief.

## Reproducing

```sh
docker run -d --name mysql83-bench -e MYSQL_ALLOW_EMPTY_PASSWORD=1 -e MYSQL_DATABASE=test -p 3308:3306 mysql:8.3
MYSQL_PORT=3308 node benchmarks/perf/setup.js
MYSQL_PORT=3308 node benchmarks/perf/capture.js
./benchmarks/perf/run-all.sh all 3308
```

See `benchmarks/perf/ANALYSIS.md` for baseline profiles, per-fix attribution, negative results (manual UTF-8 decode no longer beats native on modern V8; `Object.create(config)` options is a 2× regression; WASM parser assessment), and ranked remaining opportunities.


sidorares:
Pushed two commits addressing the patch-coverage report:

- `test: cover datetime, int64 and string decode edge paths` — targeted unit tests for every uncovered path codecov flagged. `lib/parsers/string.js` is now at 100% line coverage and all branch-added lines in `lib/packets/packet.js` are covered (verified locally with c8 + lcov cross-referenced against `git blame`).
- `fix: preserve string parsing for non-offset timezones in parseDateTime` — the new tests caught a real divergence: the fast path returned Invalid Date for timezone values outside 'local'/'Z'/'+HH:MM', while the old string path let V8 parse some timezone abbreviations (only when fractional seconds are present, e.g. `2020-01-01 13:45:56.123456PST`). Unrecognized timezones now fall back to the string path, so behaviour is bit-for-bit unchanged.

sidorares:
Two more fixes pushed, addressing review findings:

**`fix: read numPackets before queueing the compressed inflate task`** — the PR description's claim that the paused-connection queue was "the only place" retaining packets past the synchronous callback was wrong: `handleCompressedPacket` in `lib/compressed_protocol.js` closes over the packet and read `packet.numPackets` after a nextTick + async `zlib.inflate`. Safe today only because the reused instance's `numPackets` is invariantly 1 (multi-part payloads get fresh instances), but a silent compressed-sequence-id desync if that invariant ever shifts. `numPackets` is now read into a local before queueing. The new regression test (`test-compressed-protocol-packet-reuse.test.mts`) mutates the reused instance right after the synchronous hand-off and fails against the previous code — verified for both the deflated and passthrough branches. Also ran the integration suite with `MYSQL_USE_COMPRESSION=1`: green.

**`fix: return a fresh Date for invalid dates instead of a shared instance`** — the fast path returned the module-level `INVALID_DATE` singleton, and `Date` is mutable: `setFullYear()` on an Invalid Date makes it valid per spec, so one caller mutating a zero-date result would corrupt every later invalid-date value process-wide (and hand every caller the same object). All invalid-date returns — including the pre-existing binary `readDateTime` paths — now allocate `new Date(NaN)`; the invalid path is rare, so the allocation is irrelevant. Regression tests cover text + binary paths including the mutate-then-reparse scenario.

sidorares:
Pushed `fix: replicate toString bounds coercion exactly in decode fast path` — instead of just NaN-proofing the two guards, the fast path now runs the same bounds algorithm as `Buffer.prototype.toString` (`start <= 0 → 0`, `start >= len → ''`, `|0` coercion which also maps NaN → 0 and truncates fractionals, `undefined`/oversized end → len). That covers the fuzz shapes (NaN start with oversized end, negative end) plus fractional offsets and `undefined` end, so `decode` stays a true drop-in even for callers that don't exist yet.

The new test asserts decode ≡ toString over the full grid of degenerate offsets (NaN, ±Infinity, negative, fractional, oversized, undefined) × fast-path encodings. Spot-checked the hot path is unaffected: 100k×10cols replay unchanged (~2.05M rows/s).

sidorares:
@wellwelwel a bunch of Claude perf related findings. I'll let it sit for a bit, want to review it myself first, but mostly looks OK. I saw you tagging copilot for reviews - wonder what would it find here

wellwelwel:
> @wellwelwel a bunch of Claude perf related findings. I'll let it sit for a bit, want to review it myself first, but mostly looks OK. I saw you tagging copilot for reviews - wonder what would it find here

Hey, @sidorares! I was off for a while due to tech events 🙋🏻‍♂️

About **Copilot** reviews, I tag it under various conditions, but especially when everything is done by **AI**, even if I respond humanly, it will be an **AI** reading what I said, so I use **Copilot** so that the **AI**s interact with each other before I act humanly, but also when:

- I see common findings already "trained" in the **Copilot** Review skill
- I can't find much context about what the contributor did (usually due to lack of issues or a very vague explanation)
- When I want a quick/basic second perspective/review

About what **Copilot** reviews (and how it reviews), from time to time I update the native skill for this based on the contributions the project receives at [.github/skills/code-review/SKILL.md](https://github.com/sidorares/node-mysql2/blob/master/.github/skills/code-review/SKILL.md) ✨

PR #4487, opened 2026-08-19
fix(promise): forward pool error events

### LINKED ISSUE #3749: error event missing typescript
I am not sure if this is possible:

```ts
import { createPool } from 'mysql2/promise'

const pool = await createPool({ /* ... */ })

pool.on('error', (err) => {
  // handle all the errors here
})
```

From the source code, it does inherit "error" listener from somewhere, but the types do not allow doing this.

Sorry I am a little busy these days so I can not dig into codes, also I try to look though the docs and there are also no descriptions about this.

Is the types wrong or I am not supposed to do this?

GiHoon1123:
Refs #3749

Promise pools inherit lifecycle events from the underlying core pool, but the error event was not forwarded and was not exposed in the promise pool types.

This adds error to the forwarded pool events and adds a matching `pool.on('error', ...)` overload for the promise API. The existing promise wrapper event test now covers error forwarding, and the TypeScript build tests cover the promise pool event signature.

Tests:
- `npm run typecheck`
- `npm run lint`
- `npm run test:build`


wellwelwel (approved):

sidorares:
@wellwelwel don't merge just yet. Wanted to check first that pool tracks idle connection errors at all - that might be not the case

GiHoon1123:
Traced it — you're right to hold off.

`PoolConnection`'s constructor swallows connection errors itself:

```js
this.once('error', () => {
  this._removeFromPool();
});
```

`_removeFromPool()` → `pool._removeConnection(this)`, which only takes the connection, not the error. So an idle connection's error never reaches `Pool` — there's nothing for `Pool` to emit, and nothing for this PR's forwarding to forward.

Making `Pool` actually emit those errors would need `PoolConnection` to forward them upward, guarded by a `listenerCount` check — without that guard, any app that's never called `pool.on('error', ...)` would crash on the next idle connection error instead of it being silently dropped like today. That's a real behavior change across both the callback and promise APIs, not something I want to fold into this PR.

Scoping this PR back to what #3749 actually asked for: the type is missing for an event `inheritEvents` already lists, so I'll keep it to that and drop the `error` addition to `inheritEvents`/`promise.d.ts` unless you'd rather track the idle-connection-error gap separately.

GiHoon1123:
Correction on my last comment: `inheritEvents` didn't already list `error` before this PR — I added it (along with the `promise.d.ts` entry). Doesn't change the conclusion, just wanted the history accurate. Will drop both since `Pool` never actually emits `error` today.

GiHoon1123:
Closing this. Once the error-forwarding piece is dropped there's nothing left to add here - the original request was for a type on an event Pool doesn't actually emit. If the idle-connection-error gap is worth fixing, that's a separate, bigger change to PoolConnection's own error handling, not this PR.

Issue #2793, opened 2024-06-21
namedPlaceholders: "ER_WRONG_ARGUMENTS" in data object with property number type for LIMIT clause.


alditis:
### Description
I have noticed some behavior that I did not expect in the **LIMIT clause.**
When I run the query with data.LIMIT = 3 **(number type)** not works it.
```js
const data = {
    USER: "myuser", // Ok
    COUNT: 10,      // Ok
    LIMIT: 3        // Error
};
```
But with data.LIMIT = "3" **(string type)** works it.
```js
const data = {
    USER: "myuser", // Ok
    COUNT: 10,      // Ok
    LIMIT: "3"      // Ok
};
```

### Reproducing error
```js
import mysql from 'mysql2/promise';

const connection = await mysql.createConnection({
    host: 'myhost',
    user: 'mydbuser',
    password: 'mydbpass',
    database: 'mydb',
    namedPlaceholders: true
});

(async()=> {
    const sql = "SELECT `code` " + 
            "FROM `tag` " +
            "WHERE `author` = :USER AND `count` = :COUNT " + 
            "LIMIT :LIMIT";

    const data = {
        USER: "myuser", // Ok 
        COUNT: 10,      // Ok
        LIMIT: 3        // With number type not works it, but with string type "3" works it.
    };
    
    try {
        const [results] = await connection.execute(sql, data);
        console.log(results);
    } catch (err) {
        console.log(err);
    }
})();
```

### The error
```
Error: Incorrect arguments to mysqld_stmt_execute
    at PromiseConnection.execute (..AbortController../src/node_modules/mysql2/promise.js:112:22)
    at file://.../src/test.js:31:44
    at file://.../src/test.js:36:3
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ER_WRONG_ARGUMENTS',
  errno: 1210,
  sql: 'SELECT `code` FROM `tag` WHERE `author` = ? and `count` = ? LIMIT ?',
  sqlState: 'HY000',
  sqlMessage: 'Incorrect arguments to mysqld_stmt_execute'
```
  
###   Environment
mysql2: 3.10.1
node: v20.13.1

### Question
Is there a way to avoid the error and use the number type **in LIMIT clause** without having to convert it to a string type?

SteveRedding:
https://github.com/sidorares/node-mysql2/issues/1239    in prepare statement mysql2 may convert javascript number type into 
mysql double type.

[...]

Issue #2302, opened 2023-12-01
Incorrect arguments to mysqld_stmt_execute, params matching query


meclondrej:
I am trying to search through rows in a mysql table with mysql2 using `LIKE`, but i am getting an error in the console about incorrect arguments. The params for the prepared statement are matching the needs of the query string.

Code:
```typescript
import mysql from "mysql2";

export const searchMovies = (phrase: string, limit: number): Promise<Movie[]> => {
    return new Promise((resolve, reject) => {
        const pattern = `%${phrase.replace("!", "!!").replace("%", "!%").replace("_", "!_")}%`;
        const sql = "SELECT * FROM movies WHERE `Name` LIKE ? LIMIT ?";
        const params = [pattern, limit];
        db.execute(sql, params, (err, rows: RowDataPacket[]) => {
            if (err) reject(err);
            if (rows == undefined) return [];
            resolve(rows.map(x => ({
                id: x.ID,
                name: x.Name,
                language: x.Language,
                year: x.Year
            })));
        });
    });
};
```

Console error:
```
/home/node/app/node_modules/mysql2/lib/packets/packet.js:728
    const err = new Error(message);       
                ^

Error: Incorrect arguments to mysqld_stmt_execute
    at Packet.asError (/home/node/app/node_modules/mysql2/lib/packets/packet.js:728:17)
    at Execute.execute (/home/node/app/node_modules/mysql2/lib/commands/command.js:29:26)
    at Connection.handlePacket (/home/node/app/node_modules/mysql2/lib/connection.js:478:34)
    at PacketParser.onPacket (/home/node/app/node_modules/mysql2/lib/connection.js:97:12)
    at PacketParser.executeStart (/home/node/app/node_modules/mysql2/lib/packet_parser.js:75:16)
    at Socket.<anonymous> (/home/node/app/node_modules/mysql2/lib/connection.js:104:25)
    at Socket.emit (node:events:514:28)   
    at addChunk (node:internal/streams/readable:376:12)
    at readableAddChunk (node:internal/streams/readable:349:9)
    at Readable.push (node:internal/streams/readable:286:10) {
  code: 'ER_WRONG_ARGUMENTS',
  errno: 1210,
  sqlState: 'HY000',
  sqlMessage: 'Incorrect arguments to mysqld_stmt_execute',
  sql: 'SELECT * FROM movies WHERE `Name` LIKE ? LIMIT ?'
}
```

Issue #1789, opened 2023-01-13
mysql2 2.3 with MySQL 8.0 fails excute query with parameters in LIMIT, eg.: SELECT ... LIMIT ?, ?


cesco69:
mysql2 2.3 with MySQL 8.0 fails when excuting query with parameters in LIMIT, eg.:

```ts
const pool = mysql.createPool({
    host: '...',
    port: '...',
    user: '...',
    password: '...',
    database: '...',
    multipleStatements: true,
    timezone: 'Z'
}).promise();

await pool.execute(`SELECT 1 LIMIT ?, ?`, [0, 1000]);
```

Error:

```
Error: Incorrect arguments to mysqld_stmt_execute
    at PromisePool.execute (D:\backend\node_modules\.pnpm\mysql2@2.3.3\node_modules\mysql2\promise.js:359:22)
```

sidorares:
I believe the cause is the same as discussed in https://github.com/sidorares/node-mysql2/issues/1239

As a workaround, try to send `LIMIT` parameters as a string, e.i

```js
await pool.execute(`SELECT 1 LIMIT ?, ?`, ['0', '1000']);
```

The problem is from the server expecting an int type parameter and client sending it as a float ( because its a JS number value )

cesco69:
ok, works with string!

Issue #1760, opened 2022-12-24
Unable to query again using the geometry type obtained from the database


coder-hxl:
geometry: 
![image](https://user-images.githubusercontent.com/85156687/209440545-3698c0ef-0c8d-42c1-80aa-1f89639d94e1.png)

code:
![image](https://user-images.githubusercontent.com/85156687/209440292-de2ad882-092b-406d-ae5d-c5e5db7082b7.png)

Print point result: 
![image](https://user-images.githubusercontent.com/85156687/209440043-ad8a540b-3e5f-48ae-8ed4-0699133a3533.png)

Print error result
![image](https://user-images.githubusercontent.com/85156687/209440274-3ddafba6-5912-4f4e-89cd-4aa28cd3a9aa.png)

sidorares:
could you tell what following code prints for you?

```js
cont conn = await pool.getConnection();
const ps = conn.prepare(roomStatement);
console.log(ps.parameters);
```

I believe in your case `data.polygon` is serialised as a `JSON.stringify` string, which is probably not what server expects

https://github.com/sidorares/node-mysql2/blob/016900f13fc4160ab63fa5717fa3f9b9bc6195a9/lib/packets/execute.js#L54

What is the syntax you'd use in SQL to insert points in a single query? Does `pool.query` works for your sql / data?

see also #1407 and #1239

we currently have the code that reads geometry https://github.com/sidorares/node-mysql2/blob/016900f13fc4160ab63fa5717fa3f9b9bc6195a9/lib/packets/packet.js#L519 but we need to add the opposite, serialisation

coder-hxl:
> 你能告诉你下面的代码打印了什么吗？
> 
> ```js
> cont conn = await pool.getConnection();
> const ps = conn.prepare(roomStatement);
> console.log(ps.parameters);
> ```
> 
> 我相信你的情况`data.polygon`被序列化为一个`JSON.stringify`字符串，这可能不是服务器所期望的
> 
> https://github.com/sidorares/node-mysql2/blob/016900f13fc4160ab63fa5717fa3f9b9bc6195a9/lib/packets/execute.js#L54
> 
> 在 SQL 中使用什么语法在单个查询中插入点？`pool.query`适用于您的 sql / 数据吗？
> 
> 另见#1407和#1239
> 
> 我们目前有读取几何的代码
> 
[...]

PR #4488, opened 2026-08-19
feat: typed parameters, and adopt integer types the server reports

sidorares:
Adds `TypedParameter` for stating a bind parameter's MySQL type explicitly, and makes the driver adopt the integer types `COM_STMT_PREPARE` already reports.

The first commit is the red baseline: tests reproducing the current failures, passing on MariaDB and failing on MySQL 8.3 and 9.7. The second makes them pass.

## Background

Bind parameter types are inferred from the JavaScript value in [`toParameter`](https://github.com/sidorares/node-mysql2/blob/master/lib/packets/encode_parameter.js). That mapping arrived in #353 and #705, replacing an earlier design that sent everything as a string and let the server coerce.

MySQL 8.0.22 then changed the server: a statement is prepared once at `PREPARE` rather than re-prepared per execution, so the parameter types of the first execution stick. This is #1239, open since October 2020 and the most-reported issue in this area. #1407 was an attempt to fix it by following the prepare response for every parameter; it stalled because the hint is not always meaningful, and the follow-up plan discussed there and in #1483 was a hybrid — explicit containers plus a safer default.

Prior discussion: #1239, #1407, #353, #705, #548, #446, #1483, #1173, #1760, #348, and the closed duplicates #1789, #1623, #2793, #2302, #1521.

## What is wrong today

**`LIMIT ?` with a number is rejected.** Every JS number is sent as `DOUBLE`, and MySQL requires an integer here.

```js
await conn.execute('SELECT * FROM t LIMIT ?', [10]);
// Error: Incorrect arguments to mysqld_stmt_execute
```

This is the single most reported symptom. Measured across every context I could find, `LIMIT`/`OFFSET` is the *only* place a `DOUBLE` is refused outright — everywhere else the server coerces, which is why the remaining problems are silent.

**A cached statement is fixed by its first execution.** Bind a number once and the position stays `DOUBLE`, so an exact `BIGINT` bound later is read back through it:

```js
await conn.execute('INSERT INTO t (big) VALUES (?)', [1]);
await conn.execute('INSERT INTO t (big) VALUES (?)', ['9007199254740993']);
// stored: 9007199254740992   — no error, no warning
```

**Comparing a number to an indexed string column loses the index, permanently.** The server compares the column numerically, and the cached statement keeps that shape. On a 100k-row table:

```
one prepared statement, bound with: str, num, str, bigint, str
before   3887 ops/s    100    100    103    107     <- never recovers
```

Deterministically, via `Handler_read_next` on a 2000-row table: 1 row read per lookup before the number, 2000 after — for every later execution, including the correctly-typed ones. MariaDB re-optimises each execution so it recovers; MySQL does not.

**The unsigned flag is never sent** — [it is hardcoded to `0`](https://github.com/sidorares/node-mysql2/blob/master/lib/packets/execute.js), so the top half of `BIGINT UNSIGNED` is unreachable — and **binary versus text cannot be expressed**, which is #1760 and the MariaDB `VECTOR` case.

## `TypedParameter`

```js
const { TypedParameter: T } = require('mysql2');

conn.execute('SELECT * FROM t LIMIT ?', [T.BIGINT(10)]);
conn.execute('INSERT INTO t (big) VALUES (?)', [T.BIGINT('9007199254740993')]);
conn.execute('INSERT INTO t (big) VALUES (?)', [T.BIGINT.unsigned('18446744073709551615')]);
conn.execute('SELECT id FROM t WHERE code = ?', [T.VARCHAR(userInput)]);
conn.execute('INSERT INTO t (raw) VALUES (?)', [T.BLOB(buffer)]);
```

Named after both the protocol types and their SQL spellings (`BIGINT`/`LONGLONG`, `INT`/`LONG`, `MEDIUMINT`/`INT24`). Integer factories expose `.unsigned`. Values that do not fit are rejected at the call site:

```js
T.TINYINT(300);    // RangeError: TINY parameter out of range: 300 is not within -128..127
T.BIGINT(9007199254740993);
// RangeError: ... exceeds Number.MAX_SAFE_INTEGER and has already lost precision;
// pass a string or BigInt instead
```

`T.BIGINT(null)` sends SQL `NULL` while keeping the declared type, so a position does not change type between null and non-null executions.

Neither server accepts every type as a bind type — MySQL refuses `INT24 YEAR ENUM SET BIT GEOMETRY`, MariaDB refuses `JSON VECTOR` — so declared types travel as the nearest type both accept (`MEDIUMINT`→`LONG`, `YEAR`→`SHORT`, `ENUM`/`SET`→`STRING`, `VECTOR`→`BLOB`, `JSON`→`VAR_STRING` on MariaDB). `BIT` and `GEOMETRY` get no factory, since neither has an unambiguous encoding; use `T.BIGINT(mask)` and `T.BLOB(wkb)`.

## Integer types adopted from the prepare response

For untyped values the driver now consults the reported parameter type, but adopts it only when **both** hold:

1. the reported type is an integer type valid as a bind type on every server (`TINY`, `SHORT`, `LONG`, `LONGLONG`), and
2. the value is already an integer — `number` that is a safe integer, `bigint`, or `boolean` — that fits it.

Anything else keeps today's inference. That is deliberately narrow, and each restriction is load-bearing:

- **A non-integer hint is never adopted.** A parameter definition carries no provenance — `SELECT ?` and `WHERE varchar_col = ?` produce byte-identical hints (`VAR_STRING`, empty schema/table/orgTable, name `?`). Adopting `VAR_STRING` would fix the index case but would also turn `SELECT ?` bound with `42` into the string `"42"`, and the two are indistinguishable. That is the wall #1407 hit.
- **A non-integer value is never upgraded.** Following the hint for `1.5` or `'7abc'` would turn a working call into a client-side error.
- **Nothing is adopted on servers that resolve no types**, and no vendor or version check is needed to achieve it. MariaDB answers `MYSQL_TYPE_NULL` for every parameter (verified: 35/35 across 27 statements) and MySQL 5.7 answers `VAR_STRING` with a zero length for every parameter. Both fail gate 1 structurally, as would any other server or proxy that reports nothing. Neither needs the adoption anyway: both pre-date the 8.0.22 change and still re-prepare when a parameter type changes.

This fixes `LIMIT ?`/`OFFSET ?` with a number, and stops a cached statement being fixed as `DOUBLE` by an integer first execution — so the `BIGINT` rounding above no longer happens. It does **not** fix the index case, which still needs an explicit `T.VARCHAR(...)`; that is a limit of what a provenance-free hint can safely support.

## Compatibility

Only MySQL 8.0 and later resolve parameter types at all:

| Server           | Reports parameter types                                |
| ---------------- | ------------------------------------------------------ |
| MySQL 8.0 and up | Yes, including the unsigned flag                       |
| MySQL 5.7        | No — `VAR_STRING`, length 0, charset 63 for every `?`   |
| MariaDB          | No — `MYSQL_TYPE_NULL` for every `?`                    |

What goes on the wire for each column type and value. `LONGLONG` marks where the hint is adopted; every other cell is unchanged from today.

**MySQL 9.7.2** (8.3.0 identical apart from `DATETIME` charset)

| column type | PREPARE hint | plain number | plain string | Date | Buffer |
| --- | --- | --- | --- | --- | --- |
| `TINYINT` | LONGLONG | **LONGLONG** | VAR_STRING | DATETIME | BLOB |
| `SMALLINT` | LONGLONG | **LONGLONG** | VAR_STRING | DATETIME | BLOB |
| `MEDIUMINT` | LONGLONG | **LONGLONG** | VAR_STRING | DATETIME | BLOB |
| `INT` | LONGLONG | **LONGLONG** | VAR_STRING | DATETIME | BLOB |
| `BIGINT` | LONGLONG | **LONGLONG** | VAR_STRING | DATETIME | BLOB |
| `BIGINT UNSIGNED` | LONGLONG uns | **LONGLONG uns** | VAR_STRING | DATETIME | BLOB |
| `YEAR` | YEAR uns | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `DECIMAL(20,2)` | NEWDECIMAL | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `FLOAT` | DOUBLE | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `DOUBLE` | DOUBLE | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `VARCHAR(32)` | VAR_STRING | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `CHAR(8)` | VAR_STRING | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `TEXT` | LONG_BLOB | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `VARBINARY(16)` | VAR_STRING bin | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `BLOB` | LONG_BLOB bin | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `DATE` | DATE | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `DATETIME(6)` | DATETIME | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `TIME(6)` | TIME | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `TIMESTAMP(6)` | DATETIME | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `JSON` | JSON | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `ENUM` | VAR_STRING | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `SET` | VAR_STRING | DOUBLE | VAR_STRING | DATETIME | BLOB |
| `BIT(8)` | BIT uns | DOUBLE | VAR_STRING | DATETIME | BLOB |

| expression | PREPARE hint | plain number | plain string |
| --- | --- | --- | --- |
| `LIMIT ?` | LONGLONG uns | **LONGLONG uns** | VAR_STRING |
| `OFFSET ?` | LONGLONG uns | **LONGLONG uns** | VAR_STRING |
| `SELECT ?` | VAR_STRING | DOUBLE | VAR_STRING |
| `SELECT ? + ?` | DOUBLE | DOUBLE | VAR_STRING |
| `WHERE i IN (?, ?)` | LONGLONG | **LONGLONG** | VAR_STRING |
| `DATE_ADD(?, ...)` | DATE | DOUBLE | VAR_STRING |

Note `YEAR` and `BIT`: the server reports them but refuses them as bind types, so they are excluded from the adoptable set. Including `YEAR` initially caused `INSERT INTO t (year_col) VALUES (?)` to start failing with `ER_WRONG_ARGUMENTS`; the test suite caught it.

**MariaDB 12.3.2 and MySQL 5.7.44** — no hint resolves, so every cell is inference, identical to today:

| server | column type | PREPARE hint | plain number | plain string | Date | Buffer |
| --- | --- | --- | --- | --- | --- | --- |
| MariaDB 12.3 | *(all 23, and every expression incl. `LIMIT ?`)* | NULL | DOUBLE | VAR_STRING | DATETIME | BLOB |
| MySQL 5.7 | *(all 23, and every expression incl. `LIMIT ?`)* | VAR_STRING len 0 | DOUBLE | VAR_STRING | DATETIME | BLOB |

## Tests

- `test/unit/packets/test-typed-parameter.test.mts` — 71 assertions: wire encoding for every supported type, two's complement, the full unsigned 64-bit range, range and precision rejection, the wire-type mapping, typed null, and every gate of the hint policy including the placeholder types 5.7 and MariaDB report. 99% statement coverage of `lib/packets/typed_parameter.js`.
- `test/integration/connection/test-execute-integer-parameters.test.mts` — the red baseline from the first commit.
- `test/integration/connection/test-typed-parameter.test.mts` — round trips for exact/unsigned/negative 64-bit, binary, typed null, forced string.
- `test/integration/connection/test-typed-parameter-plan.test.mts` — index retention via `Handler_read_next`, no timing.
- `test/integration/connection/test-prepare-hint-fidelity.test.mts` — what each server reports, the missing provenance, and the cases the driver deliberately ignores.

Full suite: **231/231 on MySQL 5.7.44, 8.3.0 and MariaDB 12.3.2**. On MySQL 9.7, 230/231 with one pre-existing failure (`test-execute-nocolumndef`, which asserts 8.x `EXPLAIN` metadata and fails on master too).

## Notes for review

- Default behaviour changes for integer positions on MySQL. I did not put this behind a connection option; happy to add one if you would prefer it opt-in for a release.
- `lib/constants/types.js` gains `0xf2: 'VECTOR'` in the reverse map. The named export and `typings/.../types.d.ts` already had it; only the number-to-name lookup was missing.
- The prepare response is already parsed and cached, so no extra round trip. `TypedParameter` also gives `mysql2` an answer for #548, #446, #1483, #1173 and #1760 at the call site.
- Cached parameter metadata goes stale after DDL (mysql2 keeps the statement, the server silently re-prepares). Harmless here — a stale integer hint still only accepts integers — but it would matter for any broader use of the hint.
- The first CI run caught two things worth noting. `YEAR` is reported by MySQL but refused as a bind type, so including it in the adoptable set broke `INSERT INTO t (year_col) VALUES (?)`; it is now excluded. And MySQL 5.7 turned out to resolve no parameter types at all, which the capability check now handles without naming a version.


sidorares:
Pushed 7e4fc7e to fix the red CI.

**`build`** — the new docs page was formatted with the root prettier config, not the website's own. Reformatted with `website/`'s config.

**MySQL 5.7** — the interesting one. 5.7 resolves no parameter types at all: it answers `VAR_STRING` with length 0 and charset 63 for every `?`, the same non-answer MariaDB gives as `MYSQL_TYPE_NULL`. I had only tested 8.3, 9.7 and MariaDB locally and generalised "MySQL reports real types" into an assertion.

No production code changed for this. The placeholder fails gate 1 structurally, so nothing is adopted on 5.7 exactly as intended, and every behavioural test already passed there — only the test that documents *what the server reports* was wrong. It now detects that capability by preparing a known-integer statement rather than naming versions, which also covers proxies and forks.

5.7 does not need the adoption anyway: it pre-dates the 8.0.22 change and still re-prepares when a parameter type changes, so `LIMIT ?` with a number works there already.

Also addressed the codecov report: added coverage for the remaining encoders (FLOAT, DATE, TIMESTAMP, TIME in all four input forms, length-coded text, and the unsupported-type path). `lib/packets/typed_parameter.js` goes from 92% to 99% statement coverage, 71 unit assertions.

Full suite now **231/231 on MySQL 5.7.44, 8.3.0 and MariaDB 12.3.2**; 230/231 on 9.7 with the pre-existing `test-execute-nocolumndef` failure that also fails on master.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

sidorares:
The implementation is mostly CC, but I reviewed it and it looks good. Would be good to have another review @wellwelwel  when you have a chance, both in terms of surface API and internals. Surface API is probably even more important, once its in there is no way back

PR #4498, opened 2026-08-24
feat: add pool acquireTimeout to bound getConnection queue waits

### LINKED ISSUE #4497: Feature request: getConnection timeout
I ran into a problem where a service caused a deadlock by waiting on `getConnection`. Here is a rough example of the code:

```js
await pool.withTransaction(async tx => {
  await service.prepare()
  await doSomething(tx)
})
```

Many long-running requests caused the pool to saturate, then the service caused a deadlock and kept the transaction open. This is easily fixed by not giving `service` the pool, or by checking that we aren't already in a transaction.

However, a timeout could be useful to avoid these unexpected situations:

```js
// 5s timeout on getConnection
const pool = createPool({ acquireTimeout: 5_000 });

// and/or
await pool.getConnection({ timeout: 1_000 })

// or with signals
await pool.getConnection({ signal: AbortSignal.timeout(1_000) })
```

A timeout can be implemented by users, but they can't remove the request from the queue on timeout. There is also a gotcha that leaks connections if not implemented correctly. So, I think this should be a library feature.

CarlOlson:
- Closes https://github.com/sidorares/node-mysql2/issues/4497
- Implemented via Claude w/ TDD.

I'm unsure if `acquireTimeout` is the best name. I could also image `getConnectionTimeout`, `waitForConnectionTimeout`, or allowing `waitForConnections` to take a number.

<details><summary>Claude summary</summary>
<p>

A queued getConnection() previously waited indefinitely for a free connection. acquireTimeout (ms, default 0 = disabled) rejects the wait with a POOL_ACQUIRE_TIMEOUT error and splices the callback out of the connection queue so a later release() returns the connection to the free pool instead of a dead waiter.

</p>
</details>


wellwelwel:
Thanks, @CarlOlson! Can you check the implementations that aren't covered by tests? I triggered **Copilot** for a quick review, but I haven't reviewed it in more detail yet 🙋🏻‍♂️

CarlOlson:
Copilot changes should be addressed and coverage was increased.

<details><summary>Claude summary</summary>
<p>

Fixes

  - Timer outliving the queue wait (Copilot): the acquire timer kept running after a waiter was dequeued for a replacement
  connection, so a slow replacement handshake could fire POOL_ACQUIRE_TIMEOUT even though the request had already left the queue.
  Added a _dequeueConnectionRequest() helper that cancels the timer on every dequeue (all shift sites). acquireTimeout now bounds the
  queue wait only; the replacement connect is governed by connectTimeout, as documented.
  - Race removed as a bonus: cancelling on dequeue also eliminated the shift()/nextTick race, so the settled guard and its
  connection-release branch are gone — simpler, and it clears out most of the previously-uncovered lines.
  - Test fallback leak (Copilot): the regression fallback now releases any unexpectedly-delivered connection, so a regression
  produces a finite failing test instead of hanging.
  - Error code assertion (Copilot): tests now assert err.code === 'POOL_ACQUIRE_TIMEOUT', not just the message.

  Coverage (@wellwelwel)

  Added tests for the paths that were missing:
  - serving a timeout-armed waiter before expiry (timer disarmed on release)
  - the dead-connection re-entry path (waiter re-served with a fresh connection, timer disarmed on the way out)

  Every added/changed line is now covered except pool.js reset-fail re-entry (resetOnRelease: true + a reset() failure) — that branch
  was uncovered before this PR too, and forcing a reset failure needs brittle fault injection, so I left it. The helper still runs
  there and correctly disarms the timer.

</p>
</details>

sidorares:
@CarlOlson my claude still thinks the first issue is in the PR (haven't got time to validate that manually yet):

> acquireTimeout is not an upper bound on the wait (main finding). _dequeueConnectionRequest (lib/base/pool.js:159) disarms the timer, and every re-serve path (releaseConnection on a dead connection, resetOnRelease failure) then calls getConnection(cb) again — which re-enqueues the waiter with a fresh full timeout. Reproduced: acquireTimeout: 500, waiter settled after 2309 ms while another caller kept stealing the freed slot. Under connection churn (destroys on error, read-only-error destroys in pool.query, reset failures) the deadline can be pushed out indefinitely — the exact deadlock-bounding guarantee issue [#4497](https://github.com/sidorares/node-mysql2/issues/4497) asks for. Minimal fix: store an absolute deadline on the waiter and re-arm with the remaining budget instead of a full one; or arm once at the public getConnection entry (and release a late-arriving connection back to the pool, which is the leak gotcha that presumably motivated the narrow scope).

CarlOlson:
@sidorares Thanks for the merge. I believe it should be fixed now, but I'm not sure I fully understand all the code paths.

PR #4500, opened 2026-08-24
fix: correct length-coded number size for the 3-byte range

### LINKED ISSUE #4499: Off-by-one in Packet.lengthCodedNumberLength() causes "COM_STMT_EXECUTE serialized N bytes, expected N+1" for string/JSON params >= 65535 bytes
### Bug report

**mysql2 version:** 3.24.0 (also present in 3.24.1)
**Node.js version:** 22
**OS:** Linux (Docker `node:22`)
**Connection method:** `sequelize` (v6) using `mysql2` as dialect driver, via prepared statements (`execute`)

### Description

Any query bound with a string/JSON parameter whose UTF-8 byte length falls in the range
`0xffff (65535) <= length < 0xffffff (16777215)` fails with:

```
Error: Internal error: COM_STMT_EXECUTE serialized 164094 bytes, expected 164095
```

(exact byte counts vary per payload; the delta between "serialized" and "expected" is always exactly 1)

This started after upgrading from `3.23.4` to `3.24.0`. Downgrading back to `3.23.4` immediately resolves it, with no other code changes.

### Root cause

`lib/packets/packet.js`:

```js
static lengthCodedNumberLength(n) {
  if (n < 0xfb) {
    return 1;
  }
  if (n < 0xffff) {
    return 3;
  }
  if (n < 0xffffff) {
    return 5;   // <-- should be 4
  }
  return 9;
}
```

vs. the function that actually writes the value:

```js
writeLengthCodedNumber(n) {
  if (n < 0xfb) {
    return this.writeInt8(n);
  }
  if (n < 0xffff) {
    this.writeInt8(0xfc);
    return this.writeInt16(n);
  }
  if (n < 0xffffff) {
    this.writeInt8(0xfd);
    return this.writeInt24(n);   // 1 (marker) + 3 (Int24) = 4 bytes, not 5
  }
  ...
}
```

For `0xffff <= n < 0xffffff`, `lengthCodedNumberLength()` returns **5**, but `writeLengthCodedNumber()` (called from `writeLengthCodedBuffer()`/`writeLengthCodedString()`) only actually writes **4** bytes (`0xfd` marker + 3-byte `Int24`).

Since PR #4494 ("serialize COM_STMT_EXECUTE in a single exact-size pass") switched from measuring packet size by running the real writer against a `MockBuffer` (self-consistent by construction) to pre-computing size arithmetically via `lengthCodedNumberLength()` (a separate, independently-maintained estimate), this 1-byte discrepancy is no longer self-correcting, and the new `offset !== length` guard added in the same PR throws instead of silently producing a corrupted packet.

Any parameter (string, JSON, blob) whose encoded byte length lands in `[65535, 16777215)` will trigger this — e.g. large JSON payloads, long text blobs, etc.

### Minimal reproduction — level 1, isolated function check (no DB connection, no Execute class)

Verified against a real `npm install mysql2@3.24.0` in an isolated project (not just reading source):

```js
const Packet = require('./node_modules/mysql2/lib/packets/packet.js');

for (const n of [65535, 70000, 164094, 1000000, 16777214]) {
  const predicted = Packet.lengthCodedNumberLength(n);

  const buf = Buffer.allocUnsafe(16);
  const p = new Packet(0, buf, 0, buf.length);
  p.offset = 0;
  p.writeLengthCodedNumber(n);
  const actual = p.offset;

  console.log(n, '-> predicted:', predicted, 'actual:', actual, predicted === actual ? 'OK' : 'MISMATCH');
}
```

Actual output on `mysql2@3.24.0`:
```
65535 -> predicted: 5 actual: 4 MISMATCH
70000 -> predicted: 5 actual: 4 MISMATCH
164094 -> predicted: 5 actual: 4 MISMATCH
1000000 -> predicted: 5 actual: 4 MISMATCH
16777214 -> predicted: 5 actual: 4 MISMATCH
```

### Minimal reproduction — level 2, full `Execute.toPacket()` (no DB connection, reproduces the exact production error)

```js
const Execute = require('./node_modules/mysql2/lib/packets/execute.js');

const bigString = '가'.repeat(23334); // Korean text, 3 bytes/char in UTF-8 -> 70002 bytes
console.log('parameter byte length:', Buffer.byteLength(bigString, 'utf8'));

const exec = new Execute(
  1,               // statement id
  [bigString],     // parameters
  45,              // charsetNumber (utf8mb4)
  'local',         // timezone
  null,            // attributes
  0,               // clientFlags
  false,           // jsonAsString
  []               // parameterDefinitions
);

try {
  const packet = exec.toPacket();
  console.log('OK, no error. packet length:', packet.buffer.length);
} catch (err) {
  console.log('THREW:', err.message);
}
```

**Actual results, run against real installs of each version (`npm install mysql2@<version>` in isolated folders):**

| mysql2 version | Result |
|---|---|
| `3.23.4` | `OK, no error. packet length: 70024` |
| `3.24.0` | `THREW: Internal error: COM_STMT_EXECUTE serialized 70020 bytes, expected 70021` |
| `3.24.1` | `THREW: Internal error: COM_STMT_EXECUTE serialized 70020 bytes, expected 70021` (same as 3.24.0 — **not fixed**) |

This confirms both the regression point (introduced in 3.24.0 by PR #4494) and that it is still present in the latest 3.24.1.

### Expected behavior

`lengthCodedNumberLength(n)` should return `4` for `0xffff <= n < 0xffffff`, matching what `writeLengthCodedNumber(n)` actually writes.

### Suggested fix

```diff
  if (n < 0xffffff) {
-   return 5;
+   return 4;
  }
```

### Additional notes

3.24.1 (PR #4495) does not fix this, confirmed by actually running the reproduction above against `mysql2@3.24.1` — it throws the identical error. That PR's own stated validation criterion was "byte-identical output compared to v3.24.0", so this length-calculation path was intentionally left unchanged.

lazerg:
`Packet.lengthCodedNumberLength()` returns 5 for the `0xffff <= n < 0xffffff` range, but `writeLengthCodedNumber()` only ever writes 4 bytes there (a 1-byte `0xfd` marker plus a 3-byte Int24). Before #4494, packet sizes were measured with a real dry-run write, so this drift never surfaced. Since #4494 switched to computing sizes arithmetically from `lengthCodedNumberLength()`, any string/JSON/blob parameter whose encoded length falls in that range now throws `COM_STMT_EXECUTE serialized N bytes, expected N+1`. This fixes the length function to match what the writer actually emits.

Fixes #4499

### DIFF
--- a/lib/packets/packet.js
+++ b/lib/packets/packet.js
@@ -1135,7 +1135,7 @@ class Packet {
       return 3;
     }
     if (n < 0xffffff) {
-      return 5;
+      return 4;
     }
     return 9;
   }
--- a/test/unit/packets/test-length-coded-number.test.mts
+++ b/test/unit/packets/test-length-coded-number.test.mts
@@ -53,14 +53,14 @@ describe('lengthCodedNumberLength matches the bytes writeLengthCodedNumber emits
     strict.strictEqual(Packet.lengthCodedNumberLength(250), writtenBytes(250));
   });
 
-  it('3-byte range (0xFC tag)', () => {
+  it('0xFC tag', () => {
     strict.strictEqual(
       Packet.lengthCodedNumberLength(0xfffe),
       writtenBytes(0xfffe)
     );
   });
 
-  it('4-byte range (0xFD tag)', () => {
+  it('0xFD tag', () => {
     strict.strictEqual(
       Packet.lengthCodedNumberLength(0xffff),
       writtenBytes(0xffff)
@@ -71,7 +71,7 @@ describe('lengthCodedNumberLength matches the bytes writeLengthCodedNumber emits
     );
   });
 
-  it('9-byte range (0xFE tag)', () => {
+  it('0xFE tag', () => {
     strict.strictEqual(
       Packet.lengthCodedNumberLength(0xffffff),
       writtenBytes(0xffffff)
--- a/test/unit/packets/test-length-coded-number.test.mts
+++ b/test/unit/packets/test-length-coded-number.test.mts
@@ -39,3 +39,42 @@ describe('writeLengthCodedNumber / readLengthCodedNumber roundtrip', () => {
     strict.strictEqual(roundtrip(0x1ffffffffff), 2199023255551);
   });
 });
+
+describe('lengthCodedNumberLength matches the bytes writeLengthCodedNumber emits', () => {
+  const writtenBytes = (n: number): number => {
+    const buffer = Buffer.alloc(16);
+    const packet = new Packet(0, buffer, 0, buffer.length);
+    packet.offset = 0;
+    packet.writeLengthCodedNumber(n);
+    return packet.offset;
+  };
+
+  it('1-byte range', () => {
+    strict.strictEqual(Packet.lengthCodedNumberLength(250), writtenBytes(250));
+  });
+
+  it('3-byte range (0xFC tag)', () => {
+    strict.strictEqual(
+      Packet.lengthCodedNumberLength(0xfffe),
+      writtenBytes(0xfffe)
+    );
+  });
+
+  it('4-byte range (0xFD tag)', () => {
+    strict.strictEqual(
+      Packet.lengthCodedNumberLength(0xffff),
+      writtenBytes(0xffff)
+    );
+    strict.strictEqual(
+      Packet.lengthCodedNumberLength(0xfffffe),
+      writtenBytes(0xfffffe)
+    );
+  });
+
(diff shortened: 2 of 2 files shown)


sidorares:
Thanks @lazerg ! Fix looks solid, thanks for catching the bug (which existed for long time, but recent change accidentally uncovered it). I added some integration tests in https://github.com/sidorares/node-mysql2/pull/4503

sidorares:
@lazerg fix published as mysql@3.24.2

PR #4513, opened 2026-08-30
fix(typings): PoolCluster node events emit a string nodeId

NotAFlightRisk:
The pool cluster unit tests already treat the node id as a string - `test-connection-error-remove.test.mts` asserts the removed id equals `'SLAVE1'`, and the restore-events one checks `'MASTER'` for both online and offline. The typings say `number`, on all three events.

Wrong in both directions:

- `nodeId.toUpperCase()` is the call that actually works, and tsc rejects it with TS2339
- `nodeId.toFixed(2)` type-checks fine, then dies with `nodeId.toFixed is not a function`

`add()` sets the id to either the group name you passed or `CLUSTER::${n}`, and it goes out unchanged, so a string is the only thing it can be. A number never gets that far anyway, beacuse the node lookup calls `.match()` on the id first and throws.

Nothing caught it because `test/tsc-build` has no `.on(` calls in it at all. The unit tests do listen, but they annotate the id themselves, which picks the catch-all overload instead. The two files here are that check, and theyre most of the diff.

The promise side only declares `remove`, so that's the only one I touched there.

Its technically source-breaking for anyone who wrote to the old type, though the only code that stops compiling is code that already throws.

### DIFF
--- a/promise.d.ts
+++ b/promise.d.ts
@@ -126,7 +126,7 @@ export interface PoolCluster extends EventEmitter {
   of(pattern: string, selector?: string): PoolNamespace;
 
   on(event: string, listener: (...args: any[]) => void): this;
-  on(event: 'remove', listener: (nodeId: number) => void): this;
+  on(event: 'remove', listener: (nodeId: string) => void): this;
   on(event: 'warn', listener: (err: Error) => void): this;
 }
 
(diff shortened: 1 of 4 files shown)


wellwelwel (approved):
I tested your diff locally and the changed tests and types don't indicate a breaking change in this case. In fact, they are currently wrong since in the core itself (**JS**), the code would break at runtime if the user tried to treat the result as a number.

Thanks for the **bugfix**, @NotAFlightRisk!

NotAFlightRisk:
Cheers @wellwelwel. The breaking-change bit was the one part I wasn't sure of, so thanks for testing it yourself rather than taking my word for it.

